PR #9, opened 2026-09-08
[PROBE: energy] energy/latent-heat-et-consistency — latent heat and ET must describe one evaporation

### LINKED ISSUE #3: [PROBE] energy/latent-heat-et-consistency and energy/evaporative-partition — cross-budget ET coherence, and the latent-to-sensible shift under a seasonal drydown
### Conservation law

energy

### Residual equation

Two probes, proposed together because they share a vocabulary, a generator family and a set of reference models. The first claims `coupled/latent-heat-et-consistency` from ROADMAP.md; the second is not on the list. Both are filed under `energy`, because `probe.schema.json` and this form
admit only mass, energy and momentum, and I would rather add probes than a law.

I have implemented both and they pass `ht gate`. Every number below is measured. Nothing is opened as a PR until this proposal is answered.

---

### Probe 1 — `energy/latent-heat-et-consistency`

**R1, the cross-budget identity, evaluated at every step**

    R1(t) = hfls(t) - lambda(T, phase) * evspsbl(t) * rho_w / 86400   [W m-2]

    evspsbl [mm day-1] == [kg m-2 day-1] with rho_w = 1000 kg m-3
    lambda_v(T) = 2.501e6 - 2361 * T             [J kg-1], T in degC
    lambda_s    = lambda_v(0) + lambda_f
                = 2.501e6 + 3.337e5 = 2.835e6    [J kg-1]

    Scale: 1 mm day-1 is 28.95 W m-2 at 0 degC, 28.36 at 20, 27.99 at 35.
    lambda_s / lambda_v(0) = 1.133.

**R2, the surface energy budget**

    R2(t) = rn(t) - hfls(t) - hfss(t) - hfg(t)                       [W m-2]

    No energy state is declared: hfg is the storage term. A model that
    declines to report it is INCOMPLETE rather than in violation.

**R0, the water budget, which must also close**

    R0(t) = pr - evspsbl - mrro - d(mrso + snw + canopy)/dt          [mm day-1]

    A claim about incoherence between two ledgers means nothing unless each
    ledger is closed on its own terms.

---

### Probe 2 — `energy/evaporative-partition`

Two runs of one seed carrying byte-identical `rn`, `tas` and `pet`, differing only in that the `drought` variant loses 120 days of summer rain. Therefore

    sum_W (rn_d - rn_c) == 0    by construction

and conservation forces

    R3 = sum_W [ (hfls_d - hfls_c) + (hfss_d - hfss_c) + (hfg_d - hfg_c) ]
       = 0                                                   [W m-2 day]

The latent heat a drying surface gives up has nowhere to go but the sensible and ground fluxes. This is an identity, not a parameterisation: no empirical relation in it, no atmospheric feedback to model, nothing to tune.

Probe 1 asks whether a model's two budgets agree on one number. Probe 2 asks whether they agree under a change neither of them has seen.

**On the name.** The case is an idealised **seasonal drydown**, not a flash drought, and the probe says so rather than borrowing the term. The flash drought literature settled its definition on the rate of intensification rather than duration (Otkin et al. 2018), with onset criteria on the order of
one to six pentads; a 120-day window is on the wrong side of that line. What the case shares with a flash drought is the repartitioning, and the stretch in which that repartitioning happens is marked and reported separately.

### R1 — 0.5 percent of |hfls|, floor 0.5 W m-2

The denominator crosses zero, so a floor is mandatory. Both numbers were set by measurement rather than by taste.

**Why not 5 percent.** A model using a constant latent heat of vaporisation is wrong by at most 3.4 percent over a -20 to +35 degC record. The engineering rule cannot see the single most likely failure mode.

**Why the floor is 0.5 and not 2.** I tried 2 W m-2 first, reasoning that it is a negligible flux. It is not a negligible tolerance. At 2 W m-2 a
constant-lambda model survives every step whose evaporation is below roughly 3.5 mm/day, which on this record is most of them: it trips 4 steps of 3650 with a worst step at 1.52 times tolerance, instead of 33 steps at 6.08. That is not a margin, and another seed could let it through. At 0.5 the exact model still passes with a worst step of 0.00.

**The objection this will always draw, and why it argues the other way.**
*Observed surface energy budgets do not close, so how can a benchmark demand half a percent?* They do not, and the gap is well documented: closure at eddy covariance sites varies systematically in space and time (Cui & Chui 2019), and a recent correction across 172 flux towers moves the mean imbalance from -14.99 to -0.65 W m-2 (Zhang et al. 2024) — a raw imbalance of order 15 W m-2, about thirty times this probe's floor.

That is exactly why the test belongs on the synthetic track and measures a model against itself. If an observed budget were the reference, no tolerance would be defensible, because the reference would be the thing that does not close. ROADMAP already says this of the real-data track: "observed budgets do not close, so observations can never be the reference." This is the same argument applied to the energy budget, and the reason `flux_identity` should never be ported there.

For scale on the other side: 0.5 percent is one to two orders of magnitude below the spread between published ET products, which is routinely 10 to 30 percent at annual scale.

**Dilution guards.** Scored per step, never cumulatively, and reported separately for snow-free, sublimating and interval-bounded steps.

### R2 — 5 percent of accumulated |rn|, floor 2 W m-2 on the mean step residual

Standard energy form. ROADMAP already flags the sign change for `energy/surface-energy-closure`; the same floor applies.

### R3 — three bounds, and a denominator that cannot vanish

    responded    sum_W dLE <= -0.10 * sum_W rn
    conserved    |dLE + dH + dG| <= 0.05 * max(|dLE|, 0.10 * sum_W rn)
    went to air  |dG|           <= 0.20 * max(|dLE|, 0.10 * sum_W rn)

The response gate supplies the closure gate's denominator, so it cannot approach zero. The three are reported separately, because "did not respond", "responded but lost the energy" and "responded but put it in the soil" are different findings about a model.

`min_shift = 0.10` has a factor of three in hand: the exact model's latent flux falls by 29 to 43 percent of the window's net radiation across seeds. The gate sits below the measured range, not fitted to it.

`max_ground_share = 0.20` is the one number in either probe that is a physical judgement rather than an identity. Over a four-month window the ground heat store is a small residual of the surface energy budget; G/Rn is a bounded fraction that land surface schemes parameterise explicitly and that varies with cover rather than with soil moisture (Kustas & Daughtry 1990; Yang et al. 1999; Liang et al. 1999). The number is open to argument. Dropping the bound is not, for the reason in the next section.

Six reference models, five of them new. Every criterion has one that catches it.

| Model | Water closes | Energy closes | Coherent | Caught by |
| `reference_coupled` | yes | yes | yes | nothing — exact, must pass |
| `reference_two_head` | yes, 1e-15 | yes, 1e-15 | **no** | `flux_identity`, `partition_shift` |
| `reference_constant_lambda` | yes | yes | nearly | `flux_identity` |
| `reference_sublimation_blind` | yes | yes | nearly | `flux_identity`, sublimating steps |
| `reference_ground_dodge` | yes, 1e-15 | yes, 1e-15 | **yes** | `partition_shift` only |
| `reference_energy_leak` | yes | no | yes | `energy_closure` |

`reference_two_head` is the model ROADMAP asks for: "closes both budgets separately but incoherently". Water side is the exact bucket; energy side partitions net radiation by a fixed climatological evaporative fraction that never reads the soil. It is the honest caricature of a multi-head neural model, caught with an implied bulk lambda of 3.13e6 J kg-1 against a true 2.46e6, on 3495 steps of 3650.

**`reference_ground_dodge` is why I am proposing two probes rather than one.**
Its latent heat agrees with its evaporation at every step. Its water budget closes to floating point. So does its energy budget. **It passes every
criterion of probe 1** — a clean PASS, I ran it. Its sensible heat flux is a fixed share of net radiation that never reads the soil, so when the soil dries
the ground flux absorbs the entire shift:

    FAIL  partition_shift   hfg absorbed 1.00 of the shift (limit 0.20); the
                            energy went into the soil instead of the air
                            [dhfls -1266, dhfss +0, dhfg +1266 (W m-2 day),
                             dEF -0.112; intensification stretch dEF -0.204]

Over four months that is a claim about soil heat capacity that no soil has. It is also the dodge available to any model that knows a partition probe is
watching only the sum. Step-wise coherence and counterfactual coherence are not the same property: this model has the first without the second.

For contrast, on the same seed:

    reference_coupled   ok    dhfls -1266, dhfss +1266, dhfg +0, sum -2.3e-13,
                              dEF -0.333; intensification stretch (45 steps)
                              dhfls -1236, dhfss +1236, dEF -0.464

    reference_two_head  FAIL  hfls barely moved: +0 against a required fall of
                              422.4 (10% of the 4224 the window received)

### How would a model pass while understanding no physics?

It cannot, and the reason is worth stating precisely.

To pass R1 it must know that lambda varies continuously with temperature and jumps 13.3 percent at the phase change. A scalar ET regression cannot produce that: the water leaving the surface has to have been partitioned by phase before it is converted to energy.

To pass R3 it must move a specific quantity of energy from latent to sensible in response to a water deficit that never occurred, under radiation that did not change. No training-distribution regularity supplies this, because the counterfactual half of the pair does not exist in any record. A model can be fitted to look conservative in sample; it cannot be fitted to a run that was never observed.

A consequence beyond the benchmark: the latent-to-sensible transfer as soil dries is the mechanism by which the land surface amplifies heat extremes (Zhou et al. 2021; Dong et al. 2020 shows the soil-evaporation-stress operator this probe interrogates is what separates land surface models from one another). A model that fails `partition_shift` is not merely inconsistent — it is structurally incapable of representing a hot drought, however well it fits a hydrograph. Failing R3 is a statement about what the model can never be used for.

### Prior art, credited

Enforcing conservation inside a learned emulator, and quantifying the "physical inconsistency" left when it is not, is established: Beucler et al.
(2021) constrain neural networks emulating convection architecturally or through the loss, and measure violations across mass, momentum, radiation and energy budgets. That is the nearest prior art and I would rather cite it than have a reviewer find it.

Three things differ. The verdict is binary rather than a penalty term, because a benchmark must be able to reject. The budgets are a catchment's surface water and energy budgets rather than a convective column's. And R3 asserts a cross-budget *response to a counterfactual* rather than a same-instant residual, which no penalty evaluated on a validation set reaches.

R3 is a metamorphic relation in the software-testing sense — how an output must change when an input changes in a known way. The suite already has two of them on the water budget, `mass/causality` and `mass/extreme-rain`. This would be the first on the energy partition.

I am not aware of a physics-aware AI hydrologic architecture that constrains anything but mass; Wang et al. (2025) is the most recent and relaxes even that. None of the seven merged mass probes can see any of the failures above, because the incoherence lives strictly between the budgets.

Synthetic, seeded, generated at run time. Neither probe can use real data; probe 2's case is a comparison between a record and something that did not happen.

### Shared design

The water side is the record of `mass/catchment-closure` unchanged — same occurrence-and-depth precipitation process, same annual cycle plus AR(1) temperature — so a model already adapted to that probe sees nothing unfamiliar. Two things are added.

**Net radiation** from a clear-sky cycle at 40 degN, attenuated by its own AR(1) cloudiness process, with an albedo that switches below freezing and an outgoing longwave term following air temperature. It crosses zero on winter days, which is why the energy criteria carry floors.

**PET is derived from that net radiation** by Priestley-Taylor with alpha = 1.26, rather than being an independent function of temperature. A reviewer will ask why demand and available energy in a coherence probe are consistent with one another; this is the answer. They are not two forcings, they are one. The consequence that matters: PET stays small but non-zero on cold clear days, without which no snowpack ever sublimates and half of R1 has nothing to score.

Typical record: ~800 mm/yr precipitation, ~550 mm/yr PET, net radiation averaging 79 W m-2, ~550 dry sub-zero days in 4015.

### Probe 2: 3 years plus a year of spinup, 3 seeds, variants [control, drought]

    variants carry byte-identical rn, tas and pet
    drought sets pr = 0 for 120 days from 1 June of the second scored year
    `_perturbed` marks the scoring window, `_intensification` days 76 to 120
    both are stripped before the model sees the forcing

Holding rn identical turns the partition shift from an approximate expectation into an identity. Real droughts arrive with coupled warming, which would make the shift larger, so the simplification cannot make an honest model look worse.

**Why the window is 120 days, measured rather than assumed.** Ninety days was the first attempt: on some seeds the exact model's latent flux fell by as little as 8 percent of the window's net radiation, because the soil crossed no stress threshold in either variant and the removed rain left through runoff and storage. That is a perturbation, not a regime transition. At 120 days and a soil capacity of 120 mm — smaller than probe 1's 320 mm, deliberately, since a bucket that cannot empty cannot become water-limited — the fall is 29 to 43 percent across seeds.

**Where the shift actually happens, and why it is reported separately.**

| Days into the window | Share of the exact model's total shift |
| --- | --- |
| 1 – 60 | under 1 percent |
| by day 82 – 85 | 10 percent |
| by day 105 | 50 percent |
| days 76 – 120 | about 98 percent |

Stable across seeds, and the reason is not obvious: the control run keeps receiving rain, so the two runs do not diverge until the drought run has spent its soil buffer. I tried shortening the lag by lowering the soil capacity to 60, 80 and 100 mm and by starting the window in July; it stays at 50 to 70 days.

So `_intensification` marks days 76 to 120, and the criterion reports the shift over that stretch alongside the total — which turns a four-month number into a rate, and answers "is this a flash drought" with data rather than a claim. It is reported and never gated: its placement comes from the reference model (which is how this suite already locates evaluation windows), and a threshold positioned by the reference model is a threshold worth not having.

`partition_shift` also reports the change in evaporative fraction, LE / (LE + H), which is the form this literature reads in: -0.333 over the window and -0.464 over the intensification stretch for the exact model.

The window is sized against the moisture-limited / energy-limited framework of Haghighi et al. (2018), used offline to choose it and never entering the scoring.

### What the harness needs — much less than I first thought

1. **`hfls`, `hfss`, `hfg` join `FLUX_VARS` and `UNITS` in W m-2.** Three lines in `spec.py`. **The JSON schemas need no change at all**: they never enumerated variable names, only `emits.fluxes` as free strings.
2. **`energy_closure`, a new criterion.** `closure` cannot be reused for the energy budget: it unconditionally differences `probe.requires_states`, which are the water states in mm. Keeping them separate also lets one probe score both budgets without two criteria of the same name. `DENOMINATORS` already carries `sum_abs_rn` and the module docstring already anticipates the sign change, so the intent was there; only the states coupling was in the way.
3. **`flux_identity` and `partition_shift`, two new criteria.**
4. **Six reference models** as tabled above.

### One design decision I want on the record

`flux_identity` never assumes the sublimated mass. Each step falls into one of three cases, decided from the forcing and the model's own snow state:

| Case | Asserted |
| --- | --- |
| No pack, and none can fall during the step | `LE = lambda_v(T) E` |
| Pack, no precipitation, air below freezing: nothing can fall and nothing can melt, so the pack's own reported loss is the sublimated mass | `LE = lambda_v L + lambda_s S` |
| Pack, anything else | `lambda_v(T) E <= LE <= lambda_s E` |

The third case is deliberately weak. My first version asserted the equality everywhere and **failed the exact model**, because the reference sublimates whenever a pack is present while the criterion could only infer the split on dry frozen steps. The fix is not to make the model match the criterion — that is circular — but to have the criterion assert only what it can establish. A criterion that fails an honest model whose snow physics differs from the reference's is worse than no criterion. Discrimination survives: `reference_sublimation_blind` is caught on the second case, which this record supplies about 400 steps of per seed.

(A second boundary the gate found: snow that falls and sublimates away inside one step leaves SWE at zero on both sides, so "no pack" must exclude steps where snow could have fallen.)

### One consequence the maintainers should decide on

Once the suite has an energy probe, **`reference_bucket` is INCOMPLETE** on it, and `ht run --model reference_bucket` no longer passes the whole suite. This is the correct verdict rather than a regression — it has not violated conservation of energy, it has declined to be falsifiable about it, exactly as `reference_streamflow_only` does on the budget probes today — but it changes the README's headline example, and two tests in `tests/test_window.py` assume both that the first probe in registry order is `mass/catchment-closure` and that a water-only model passes the whole suite. Neither holds any more. I scoped those two tests to their fixture's probe; `pytest -q` is otherwise unchanged, 18 failed / 81 passed before and after on my machine (Windows; the failures are pre-existing and environmental).

### References

- Beucler, T., Pritchard, M., Rasp, S., Ott, J., Baldi, P., & Gentine, P. (2021). Enforcing analytic constraints in neural networks emulating physical systems. *Physical Review Letters*, 126, 098302. https://doi.org/10.1103/PhysRevLett.126.098302
- Christian, J. I., Hobbins, M., Hoell, A., Otkin, J. A., Ford, T. W., Cravens, A. E., Powlen, K. A., Wang, H., & Mishra, V. (2024). Flash drought: A state of the science review. *WIREs Water*, 11(3), e1714. https://doi.org/10.1002/wat2.1714
- Cui, W., & Chui, T. F. M. (2019). Temporal and spatial variations of energy balance closure across FLUXNET research sites. *Agricultural and Forest Meteorology*, 271, 12-21. https://doi.org/10.1016/j.agrformet.2019.02.026
- Dong, J., Dirmeyer, P. A., Lei, F., Anderson, M. C., Holmes, T. R. H., Hain, C., & Crow, W. T. (2020). Soil evaporation stress determines soil moisture-evapotranspiration coupling strength in land surface modeling. *Geophysical Research Letters*, 47(21). https://doi.org/10.1029/2020GL090391
- Haghighi, E., Short Gianotti, D. J., Akbar, R., Salvucci, G. D., & Entekhabi, D. (2018). Soil and atmospheric controls on the land surface energy balance. *Water Resources Research*, 54(3), 1831-1851. https://doi.org/10.1002/2017WR021729
- Kustas, W. P., & Daughtry, C. S. T. (1990). Estimation of the soil heat flux/net radiation ratio from spectral data. *Agricultural and Forest Meteorology*. https://doi.org/10.1016/0168-1923(90)90033-3
- Liang, X., Wood, E. F., & Lettenmaier, D. P. (1999). Modeling ground heat flux in land surface parameterization schemes. *Journal of Geophysical Research: Atmospheres*. https://doi.org/10.1029/98JD02307
- Otkin, J. A., Svoboda, M., Hunt, E. D., Ford, T. W., Anderson, M. C., Hain, C., & Basara, J. B. (2018). Flash droughts: A review and assessment of the challenges imposed by rapid-onset droughts in the United States. *Bulletin of the American Meteorological Society*, 99(5), 911-919. https://doi.org/10.1175/BAMS-D-17-0149.1
- Wang, Y., Zhang, L., Erichson, N. B., & Yang, T. (2025). A mass conservation relaxed (MCR) LSTM model for streamflow simulation across CONUS. *Water Resources Research*, 61(8), e2024WR039131. https://doi.org/10.1029/2024WR039131
- Yang, Z.-L., Dai, Y., Dickinson, R. E., & Shuttleworth, W. J. (1999). Sensitivity of ground heat flux to vegetation cover fraction and leaf area index. *Journal of Geophysical Research: Atmospheres*. https://doi.org/10.1029/1999JD900230
- Zhang, W., Nelson, J. A., Miralles, D. G., Mauder, M., Migliavacca, M., Poyatos, R., Reichstein, M., & Jung, M. (2024). A new post-hoc method to reduce the energy imbalance in eddy covariance measurements. *Geophysical Research Letters*, 51(2), e2023GL107084. https://doi.org/10.1029/2023GL107084
- Zhou, S., Williams, A. P., Lintner, B. R., Berg, A. M., Zhang, Y., Keenan, T. F., Cook, B. I., Hagemann, S., Seneviratne, S. I., & Gentine, P. (2021). Soil moisture-atmosphere feedbacks mitigate declining water availability in drylands. *Nature Climate Change*, 11(1), 38-44. https://doi.org/10.1038/s41558-020-00945-z
- Brutsaert, W. (1982). *Evaporation into the Atmosphere*. Reidel.

Happy to split into two PRs if you prefer smaller review units: PR1 the vocabulary, `flux_identity`, `energy_closure` and probe 1; PR2 the paired case, `partition_shift` and probe 2. They are already two commits, and PR1 passes the gate on its own.

### Will you build it?

Yes — assign it to me

licm13:
Closes #3

The first of two. PR2 (`energy/evaporative-partition`) is stacked on this one and reuses its vocabulary, its `energy_closure` criterion and its reference models. This PR stands alone — 15 probes, 28 models, `ht gate` green with nothing from PR2 present. PR2 does not stand alone.

This is the ROADMAP item `coupled/latent-heat-et-consistency`, filed under `energy` because `probe.schema.json` and the proposal form admit only mass, energy and momentum. `energy/pet-consistency` shows that is the right shelf for a probe about evaporation whatever variables it happens to need.

## What this probe asserts

A model with a water head and an energy head can close its water budget to floating point, close its energy budget to floating point, and still report a latent heat flux implying a different evaporation than the one it reported as water. The contradiction lives strictly between the ledgers, so no single-budget criterion can reach it.

**R1 — the identity, at every step**

    R1(t) = hfls(t) - lambda(T, phase) * evspsbl(t) * rho_w / 86400   [W m-2]

    lambda_v(T) = 2.501e6 - 2361 * T                  [J kg-1], T in degC
    lambda_s    = lambda_v(0) + lambda_f = 2.835e6    [J kg-1]

A failure means the model is describing two different evaporations and calling them one. Physically: the water leaving the surface was never converted to energy by any process — the two numbers were produced independently and never had to agree.

**R2 — the surface energy budget**

    R2(t) = rn(t) - hfls(t) - hfss(t) - hfg(t)                       [W m-2]

`hfg` is the storage term, so no energy state has to be declared. A model that declines to report it is INCOMPLETE, not in violation.

**R0 — the water budget**, on the same terms as `mass/catchment-closure`. A claim about incoherence between two ledgers means nothing unless each ledger is closed on its own terms, so this probe scores both.

### Tolerances, and why R1 is not five percent

A constant latent heat of vaporisation is wrong by **at most 3.4 percent** over a -20 to +35 degC record. The engineering rule cannot see the most likely failure mode, so R1 is scored per step against **0.5 percent** of the reported flux.

The absolute floor is **0.5 W m-2, not 2**. I tried 2 first, reasoning that it is a negligible flux. It is not a negligible tolerance: at 2 W m-2 `reference_constant_lambda` trips 4 steps of 3650 with a worst step at 1.52 times tolerance, instead of 33 steps at 6.08. That is not a margin, and another seed could let it through. At 0.5 the exact model still passes with a worst step of 0.00.

### The objection this will always draw

*Observed surface energy budgets do not close. How can a benchmark demand half a percent?*

They do not, and the gap is documented: closure at eddy covariance sites varies systematically in space and time (Cui & Chui 2019), and a recent correction across 172 flux towers moves the mean imbalance from -14.99 to -0.65 W m-2 (Zhang et al. 2024) — a raw imbalance of order 15 W m-2, about thirty times this probe's floor.

That is exactly why the test belongs on the synthetic track and measures a model against itself. If an observed budget were the reference, no tolerance would be defensible, because the reference would be the thing that does not close. ROADMAP already says this of the real-data track: "observed budgets do not close, so observations can never be the reference." This is the same argument applied to the energy budget, and the reason `flux_identity` should never be ported there. The README carries this in full.

### The phase change, and what is deliberately not asserted

`lambda_s / lambda_v(0) = 1.133`. A model converting every kilogram at the vaporisation rate is short by 13.3 percent on exactly the steps where a pack is losing mass to the atmosphere.

The sublimated mass is never assumed. Each step falls into one of three cases, decided from the forcing and the model's own snow state:

| Case | Asserted |
| --- | --- |
| No pack, and none can fall during the step | `LE = lambda_v(T) E` |
| Pack, no precipitation, air below freezing: nothing can fall and nothing can melt, so the pack's own reported loss is the sublimated mass | `LE = lambda_v L + lambda_s S` |
| Pack, anything else | `lambda_v(T) E <= LE <= lambda_s E` |

The third case is weak because of a bug the gate caught. My first version asserted the equality everywhere and **failed the exact model**: the reference sublimates whenever a pack is present, while the criterion could only infer the split on dry frozen steps. The fix is not to change the model to match the criterion — that is circular — but to have the criterion assert only what it can establish from what the model reported. A criterion that fails an honest model whose snow physics differs from the reference's is worse than no criterion.

Discrimination survives: `reference_sublimation_blind` is caught on the second case, which this record supplies about 400 steps of per seed.

(A second boundary the gate found, kept as a comment: snow that falls and sublimates away inside one step leaves SWE at zero on both sides, so "no pack" must also exclude steps where snow could have fallen.)

### The case

`generate.py`, seeded, 10 years plus a year of spinup, 5 seeds. The water side is the record of `mass/catchment-closure` unchanged. Two things are added.

Net radiation from a clear-sky cycle at 40 degN, attenuated by its own AR(1) cloudiness process, with an albedo that switches below freezing and an outgoing longwave term following air temperature. It crosses zero on winter days, which is why the energy criteria carry floors.

PET is then **derived from that net radiation** by Priestley-Taylor with alpha = 1.26 rather than being an independent function of temperature. A reviewer will ask why demand and available energy in a coherence probe are consistent with one another; this is the answer. They are not two forcings, they are one. The consequence that matters is that PET stays small but non-zero on cold clear days — without which no snowpack ever sublimates and half of R1 has nothing to score.

Typical record: ~800 mm/yr precipitation, ~550 mm/yr PET, net radiation averaging 79 W m-2, ~550 dry sub-zero days in 4015.

### Prior art

Beucler et al. (2021) enforce analytic constraints in neural networks emulating convection, architecturally or through the loss, and quantify the "physical inconsistency" that remains across mass, momentum, radiation and energy budgets. That is the nearest prior art and the README credits it. What differs here: a binary verdict rather than a penalty term, because a benchmark must be able to reject; and a catchment's surface budgets rather than a convective column's.

## Discrimination

`ht gate --probe energy/latent-heat-et-consistency` — GATE PASSED, seeds [1415376846, 1922328960, 281797427, 788749541, 1295701655].

The template's default rows are the mass probe's models; this probe declares its own, all five new in this PR.

| --- | --- | --- |
| `reference_coupled` | PASS | n/a — exact by construction, must pass everything |
| `reference_two_head` | FAIL | `flux_identity` |
| `reference_constant_lambda` | FAIL | `flux_identity` |
| `reference_sublimation_blind` | FAIL | `flux_identity` |
| `reference_energy_leak` | FAIL | `energy_closure` |
| `reference_bucket` | INCOMPLETE | reports no energy fluxes |

`reference_two_head` is the model ROADMAP asks for — "closes both budgets separately but incoherently". Water side is the exact bucket; energy side partitions net radiation by a fixed climatological evaporative fraction that never reads the soil. Both budgets close to 1e-15. Implied bulk lambda 3.13e6 J kg-1 against a true 2.46e6, on 3495 steps of 3650:

    ok    closure          cumulative residual 0.0000% of sum_pr (limit 5.0%)
    ok    energy_closure   cumulative residual 0.0000% of accumulated |rn|
    FAIL  flux_identity    latent heat contradicts the reported evaporation on
                           3495 of 3650 steps, first at index 0 (94.257 W m-2
                           reported, 81.488 required); implied lambda 3.128e+06

`reference_constant_lambda` and `reference_sublimation_blind` each break one half of `flux_identity` and nothing else, so neither half goes untested. `reference_energy_leak` is the energy counterpart of `reference_leaky`.

## What this changes outside the probe

**1. Vocabulary.** `hfls`, `hfss`, `hfg` join `FLUX_VARS` and `UNITS` in W m-2, alongside `gwex`. Three lines in `spec.py`. **The JSON schemas need no change**: they never enumerated variable names, only `emits.fluxes` as free strings.

**2. `energy_closure` is a new criterion rather than a denominator of `closure`.** `closure` unconditionally differences `probe.requires_states`, which here are the water states in mm. Keeping them separate also lets one probe score both budgets without two criteria of the same name. `DENOMINATORS` already carried `sum_abs_rn` and the module docstring already anticipated the sign change, so the intent was there; only the states coupling was in the way.

**3. `reference_bucket` is now INCOMPLETE on this probe**, so `ht run --model reference_bucket` no longer passes the whole suite. This is the correct verdict rather than a regression — it has not violated conservation of energy, it has declined to be falsifiable about it, exactly as `reference_streamflow_only` does on the budget probes today. But it changes the README's headline example, and two tests in `tests/test_window.py` assumed both that the first probe in registry order is `mass/catchment-closure` and that a water-only model passes the whole suite. Neither holds once the suite has a probe that requires energy fluxes. I scoped those two tests to their fixture's probe, which also makes them robust to any future probe sorting ahead of the mass ones. I have not touched the README; say which way you want it and I will.

**4. `must_pass` names `reference_coupled` alone.** The mass probes now require `[reference_bucket, flex_lumped, flex_topo, sacsma_snow17]`, and I would have followed that, but all four return INCOMPLETE here — `missing: hfls, hfss, hfg`. That is the correct verdict and worth stating plainly: every hand-written physical model in the suite, including the one added this week, cannot currently be checked for energy at all. Supplying an exact model that can is most of what this PR does.

## Checklist

- [x] `ht validate` passes — 15 probes, 28 models with this PR alone
- [x] `ht gate --probe energy/latent-heat-et-consistency` passes
- [x] `ht gate` (whole suite) passes
- [x] `pytest -q` shows no change against the pristine tree — 18 failed / 92 passed before and after; the failures are pre-existing and environmental on my Windows box
- [x] The generator is deterministic given a seed and commits no data
- [x] The probe runs in under a minute on a two-core runner — the full gate for this probe is 4.1 s
- [x] Tolerance and denominator are justified in `probe.yaml`, with the measurement that set the floor
- [x] I constructed a model that passes closure without doing physics — `reference_two_head` closes both budgets to 1e-15 — and `flux_identity` catches it

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -28,6 +28,7 @@ long.
 | `mass/antecedent-monotonicity` | Zhi Li (CU Boulder) |
 | `mass/phase-counterfactual` | Zhi Li (CU Boulder) |
 | `energy/pet-consistency` | Zhi Li (CU Boulder) |
+| `energy/latent-heat-et-consistency` | Changming Li (SCUT) |
 | `momentum/routing-conservation` | Zhi Li (CU Boulder) |
 
 ## Models
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ unclaimed.
 | [`mass/antecedent-monotonicity`](probes/mass/antecedent-monotonicity) | mass | The same storm after a dry month and a wet one: the wetter catchment runs off more, and no more than the extra water. | `reference_cheater` |
 | [`mass/phase-counterfactual`](probes/mass/phase-counterfactual) | mass | The same water falling as rain instead of snow: timing moves, the integrated volumes may not. | `reference_sublimating` |
 | [`energy/pet-consistency`](probes/energy/pet-consistency) | energy | Evaporation reaches demand when the model's own soil is wettest, stays below it, and falls when the soil is driest. | `reference_thirsty` |
+| [`energy/latent-heat-et-consistency`](probes/energy/latent-heat-et-consistency) | energy | The evaporation a model reports as water and the evaporation implied by the latent heat it reports: are they the same evaporation? | `reference_two_head`, `reference_constant_lambda`, `reference_sublimation_blind`, `reference_energy_leak` |
 | [`momentum/routing-conservation`](probes/momentum/routing-conservation) | momentum | The channel store is never negative and never holds more than its hydrograph can. | `reference_stuck_router` |
 
 ## Models
@@ -104,18 +105,27 @@ physical models' runs are archived there too. A
 physical model is there to test the probes: the exact bucket, two
 hand-written conceptual models from
 [chrimerss/HydrologicModels](https://github.com/chrimerss/HydrologicModels)
-and the NWS's SAC-SMA with Snow-17 must pass every probe, so a probe that
-fails one is wrong until shown otherwise. A broken model is broken in one specific way, so that no criterion
+and the NWS's SAC-SMA with Snow-17 must pass every probe that can ask them
+anything, so a probe that fails one is wrong until shown otherwise. All four
+report water and no energy, so all four are INCOMPLETE on
+`energy/latent-heat-et-consistency` rather than passing it: they have not
+violated conservation of energy, they have declined to be falsifiable about
+it, exactly as `reference_streamflow_only` does on the budget probes. A broken model is broken in one specific way, so that no criterion
 goes untested.
 
 | Model | Kind | What it does | Standing |
 | --- | --- | --- | --- |
 | [`google_flood_forecast`](models/google_flood_forecast) | submitted | The mean-embedding forecast LSTM behind Google Flood Hub, at the published weights. Predicts discharge and nothing else. | **FAIL (INCOMPLETE)**, 5 of 14 probes passed. Runoff-only, so three budget probes cannot ask it anything; of the seven that ask on runoff alone it passes the runoff bounds, memory and area, and fails step, extreme rain, phase, and a 0.18 mm/day dip after an added storm |
 | [`dhbv2`](models/dhbv2) | submitted | δHBV 2.0, the MHPI group's differentiable HBV: neural networks write the parameters of a bucket model that reports its stores and its evaporation. | **FAIL (VIOLATION)**, 10 of 14 probes passed. Its learned regional-groundwater term, declared as `gwex`, closes the budget to 1e-8; what remains is a learned field capacity twice the catchment's, a response to doubled rain above the rain added, a runoff depth that changes with the area it is told, and a third more runoff when snow falls as rain |
-| `reference_bucket` | exact | conserves water exactly by construction | must pass every probe |
-| [`flex_lumped`](models/flex_lumped) | physical | lumped FLEX/HBV: interception, beta-partitioned unsaturated store, fast and slow reservoirs, triangular lag | must pass every probe; **PASS**, 14 of 14 |
-| [`flex_topo`](models/flex_topo) | physical | FLEX-Topo: plateau, hillslope and wetland units on real Wark fractions sharing one groundwater store | must pass every probe; **PASS**, 14 of 14 |
-| [`sacsma_snow17`](models/sacsma_snow17) | physical | the NWS's SAC-SMA with Snow-17 and a gamma unit hydrograph, ported from the legacy Fortran and checked against it | must pass every probe; **PASS**, 14 of 14 |
+| `reference_bucket` | exact | conserves water exactly by construction | must pass every probe that can ask it anything; INCOMPLETE on the energy probe, which needs fluxes it does not report |
+| [`flex_lumped`](models/flex_lumped) | physical | lumped FLEX/HBV: interception, beta-partitioned unsaturated store, fast and slow reservoirs, triangular lag | must pass every probe that can ask it anything; **PASS**, 14 of 15, INCOMPLETE on the energy probe |
+| [`flex_topo`](models/flex_topo) | physical | FLEX-Topo: plateau, hillslope and wetland units on real Wark fractions sharing one groundwater store | must pass every probe that can ask it anything; **PASS**, 14 of 15, INCOMPLETE on the energy probe |
+| [`sacsma_snow17`](models/sacsma_snow17) | physical | the NWS's SAC-SMA with Snow-17 and a gamma unit hydrograph, ported from the legacy Fortran and checked against it | must pass every probe that can ask it anything; **PASS**, 14 of 15, INCOMPLETE on the energy probe |
+| `reference_coupled` | exact | the bucket with snow sublimation and a surface energy budget: every kilogram converted at the latent heat of the phase it actually underwent | must pass every criterion of `energy/latent-heat-et-consistency` |
+| `reference_two_head` | broken | a water head and an energy head that never meet: both budgets close to 1e-15 and the latent heat implies an evaporation it never reported | caught by `flux_identity` |
+| `reference_constant_lambda` | broken | coherent, but converts every kilogram at the same latent heat of vaporisation | caught by `flux_identity` |
+| `reference_sublimation_blind` | broken | coherent for liquid water, but converts snow sublimation at the latent heat of vaporisation instead of sublimation | caught by `flux_identity` |
+| `reference_energy_leak` | broken | discards 15% of net radiation; the energy counterpart of `reference_leaky` | caught by `energy_closure` |
 | `reference_leaky` | broken | hides a silent 15% sink | caught by `closure` |
 | `reference_cheater` | broken | solves for storage as whatever balances the budget | caught by `state_bounds` |
 | `reference_degenerate` | broken | evaporates all precipitation, produces no runoff | caught by `non_degenerate`, `response_sign` |
--- a/models/reference_constant_lambda/ht_adapter.py
+++ b/models/reference_constant_lambda/ht_adapter.py
@@ -17,7 +17,7 @@ MODE = "constant_lambda"
 MODEL = {"name": "reference_constant_lambda", "version": "1.0.0"}
 
 COLUMNS = [
-    "time", "pr", "evspsbl", "mrro", "hfls", "hfss", "hfg",
+    "time", "pr", "evspsbl", "mrro", "sbl", "hfls", "hfss", "hfg",
     "mrso", "snw", "canopy", "channel",
 ]
 
@@ -137,6 +137,9 @@ def simulate(forcing, static, dt_days=1.0):
             "time": step["time"],
             "pr": pr_rate,
             "evspsbl": (liquid + sublimation) / dt_days,
+            # The sublimating share of the evaporation above, so a
+            # criterion never has to infer which kilograms left as ice.
+            "sbl": sublimation / dt_days,
             "mrro": (surface + baseflow) / dt_days,
             "hfls": latent,
             "hfss": sensible,
--- a/models/reference_constant_lambda/ht_adapter.py
+++ b/models/reference_constant_lambda/ht_adapter.py
@@ -1,0 +1,201 @@
+#!/usr/bin/env python3
+"""HydroTuring adapter. Standard library only, to show that the /io contract
+needs no scientific Python stack and could be written in any language.
+
+Coherent, but converts every kilogram at the same latent heat.
+"""
+
+from __future__ import annotations
+
+import argparse
+import csv
+import json
+import sys
+from pathlib import Path
+
+MODE = "constant_lambda"
+MODEL = {"name": "reference_constant_lambda", "version": "1.0.0"}
+
+COLUMNS = [
+    "time", "pr", "evspsbl", "mrro", "hfls", "hfss", "hfg",
+    "mrso", "snw", "canopy", "channel",
+]
+
+EVAP_SHAPE = 0.5   # soil moisture at which evaporation reaches its potential rate
+SUBL_SHARE = 0.35  # share of the remaining demand a snowpack can meet
+GROUND_SHARE = 0.10  # share of net radiation conducted into the ground
+
+# Latent heat of vaporisation as A + B*T (T in degC) and of fusion, J kg-1.
+LAMBDA_A, LAMBDA_B = 2.501e6, -2361.0
+LAMBDA_F = 3.337e5
+LAMBDA_CONST = 2.45e6  # the constant a careless model would use
+CLIMATOLOGICAL_EF = 0.65  # evaporative fraction of a model whose heads never meet
+ENERGY_LEAK = 0.15
+SECONDS_PER_DAY = 86400.0
+
+TIMESTEP_DAYS = {
+    "PT1D": 1.0,
+    "PT1H": 1.0 / 24.0,
+    "PT15M": 1.0 / 96.0,
+    "PT5M": 1.0 / 288.0,
+    "PT1M": 1.0 / 1440.0,
+}
+
+
+def partition_energy(rn, tas, liquid_mm, sublimated_mm, dt_days):
+    """Turn the water that left the surface into the energy that carried it.
+
+    `reference_coupled` is exact: every kilogram is converted at the latent
+    heat of the phase change it actually underwent, and the sensible flux is
+    what net radiation has left once the latent and ground fluxes are taken.
+    Solving for H that way is the physical statement that the surface has no
+    other place to put the energy, and it is why this model closes the energy
+    budget by construction, exactly as `reference_bucket` closes the water
+    budget by construction. The discrimination lives in the other four.
+    """
+    seconds = dt_days * SECONDS_PER_DAY
+    lam_v = LAMBDA_A + LAMBDA_B * tas
+    lam_s = LAMBDA_A + LAMBDA_F
+    ground = GROUND_SHARE * rn
+
+    if MODE == "constant_lambda":
+        latent = LAMBDA_CONST * (liquid_mm + sublimated_mm) / seconds
+    elif MODE == "sublimation_blind":
+        latent = lam_v * (liquid_mm + sublimated_mm) / seconds
+    elif MODE == "two_head":
+        # An energy head that never reads the water head: the partition is a
+        # climatological evaporative fraction of the available energy, and it
+        # has no idea how much water actually left.
+        latent = CLIMATOLOGICAL_EF * (rn - ground)
+    else:
+        latent = (lam_v * liquid_mm + lam_s * sublimated_mm) / seconds
+
+    sensible = rn - ground - latent
+    if MODE == "energy_leak":
+        sensible -= ENERGY_LEAK * rn
+    return latent, sensible, ground
+
+
+def simulate(forcing, static, dt_days=1.0):
+    """The reference bucket, with snow sublimation and a surface energy budget.
+
+    The water side is `reference_bucket` plus one term: a snowpack meets part
+    of the evaporative demand by sublimating, which is removed from the pack
+    like every other flux, so the water budget still closes to floating point.
+    That term exists because without it no step in the record has a phase
+    change to be coherent about.
+    """
+    soil_cap = static["soil_capacity_mm"]
+    canopy_cap = static["canopy_capacity_mm"]
+    ddf = static["degree_day_factor_mm_per_C_day"]
+    k_base = static["baseflow_coefficient"]
+    t_snow = static["snow_threshold_degC"]
+
+    soil = 0.5 * soil_cap
+    swe = 0.0
+    canopy = 0.0
+    rows = []
+
+    for step in forcing:
+        pr_rate, tas, pet_rate = step["pr"], step["tas"], step["pet"]
+        rn = step.get("rn", 0.0)
+        pr = pr_rate * dt_days
+        pet = pet_rate * dt_days
+
+        snowfall = pr if tas < t_snow else 0.0
+        rain = 0.0 if tas < t_snow else pr
+
+        swe += snowfall
+        melt = min(swe, ddf * max(tas - t_snow, 0.0) * dt_days)
+        swe -= melt
+
+        water_in = rain + melt
+        intercepted = min(canopy_cap - canopy, water_in)
+        canopy += intercepted
+        throughfall = water_in - intercepted
+
+        canopy_evap = min(canopy, pet)
+        canopy -= canopy_evap
+        pet_left = pet - canopy_evap
+
+        sublimation = min(swe, SUBL_SHARE * pet_left) if swe > 0.0 else 0.0
+        swe -= sublimation
+        pet_left -= sublimation
+
+        soil += throughfall
+        surface = max(0.0, soil - soil_cap)
+        soil -= surface
+        baseflow = k_base * soil * dt_days
+        soil -= baseflow
+        soil_evap = min(soil, pet_left * min(1.0, soil / (EVAP_SHAPE * soil_cap)))
+        soil -= soil_evap
+
+        liquid = canopy_evap + soil_evap
+        latent, sensible, ground = partition_energy(rn, tas, liquid, sublimation, dt_days)
+
+        rows.append({
+            "time": step["time"],
+            "pr": pr_rate,
+            "evspsbl": (liquid + sublimation) / dt_days,
+            "mrro": (surface + baseflow) / dt_days,
+            "hfls": latent,
+            "hfss": sensible,
+            "hfg": ground,
+            "mrso": soil,
+            "snw": swe,
+            "canopy": canopy,
+            # No routing: runoff leaves the stores and the catchment in the
+            # same step, so the water in transit is identically zero.
+            # Reported rather than omitted, because it is a statement.
+            "channel": 0.0,
+        })
+    return rows
+
+
+def read_request(path: Path) -> tuple[dict, Path]:
+    request = json.loads(path.read_text())
+    return request, path.parent
+
+
+def read_forcing(path: Path) -> list[dict]:
+    with open(path, newline="") as fh:
+        rows = list(csv.DictReader(fh))
+    for row in rows:
+        for key in ("pr", "tas", "pet", "rn"):
+            if key in row:
+                row[key] = float(row[key])
+    return rows
+
+
+def write_result(path: Path, rows: list[dict]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with open(path, "w", newline="") as fh:
+        writer = csv.DictWriter(fh, fieldnames=COLUMNS)
+        writer.writeheader()
+        writer.writerows(rows)
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--request", required=True)
+    args = parser.parse_args()
+
+    request_path = Path(args.request).resolve()
+    request, io_dir = read_request(request_path)
+    forcing = read_forcing(io_dir / request["input"]["forcing"])
+    static = json.loads((io_dir / request["input"]["static"]).read_text())
+
+    timestep = request.get("timestep", "PT1D")
+    if timestep not in TIMESTEP_DAYS:
+        raise SystemExit(f"unsupported timestep {timestep!r}")
+    rows = simulate(forcing, static, TIMESTEP_DAYS[timestep])
+
+    write_result(io_dir / request["output"]["table"], rows)
+    (io_dir / request["output"]["run"]).write_text(
+        json.dumps({"status": "ok", "model": MODEL, "n_steps": len(rows)}, indent=2)
+    )
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/models/reference_constant_lambda/model.yaml
+++ b/models/reference_constant_lambda/model.yaml
@@ -12,7 +12,7 @@ runner: subprocess
 entrypoint: ["python3", "ht_adapter.py"]
 timestep: [PT1D]
 emits:
-  fluxes: [pr, evspsbl, mrro, hfls, hfss, hfg]
+  fluxes: [pr, evspsbl, mrro, sbl, hfls, hfss, hfg]
   states: [mrso, snw, canopy, channel]
 needs_forcing: [pr, tas, pet, rn]
 supports:
--- a/models/reference_constant_lambda/model.yaml
+++ b/models/reference_constant_lambda/model.yaml
@@ -1,0 +1,20 @@
+name: reference_constant_lambda
+version: "1.0.0"
+description: >
+  Fully coupled and correct in every other respect, except that the latent
+  heat of vaporisation is a constant instead of a function of temperature.
+  The error reaches 3.4 percent over a -20 to +35 degC record, which is why
+  flux_identity is scored per step against a tolerance far tighter than the
+  suite's 5 percent rule. It exists to prove that tolerance is necessary.
+authors: ["HydroTuring maintainers"]
+license: PolyForm-Noncommercial-1.0.0
+runner: subprocess
+entrypoint: ["python3", "ht_adapter.py"]
+timestep: [PT1D]
+emits:
+  fluxes: [pr, evspsbl, mrro, hfls, hfss, hfg]
+  states: [mrso, snw, canopy, channel]
+needs_forcing: [pr, tas, pet, rn]
+supports:
+  perturbation: true
+resources: {cpu: 1, memory_gb: 1, gpu: false}
--- a/models/reference_coupled/ht_adapter.py
+++ b/models/reference_coupled/ht_adapter.py
@@ -17,7 +17,7 @@ MODE = "coupled"
 MODEL = {"name": "reference_coupled", "version": "1.0.0"}
 
 COLUMNS = [
-    "time", "pr", "evspsbl", "mrro", "hfls", "hfss", "hfg",
+    "time", "pr", "evspsbl", "mrro", "sbl", "hfls", "hfss", "hfg",
     "mrso", "snw", "canopy", "channel",
 ]
 
@@ -137,6 +137,9 @@ def simulate(forcing, static, dt_days=1.0):
             "time": step["time"],
             "pr": pr_rate,
             "evspsbl": (liquid + sublimation) / dt_days,
+            # The sublimating share of the evaporation above, so a
+            # criterion never has to infer which kilograms left as ice.
+            "sbl": sublimation / dt_days,
             "mrro": (surface + baseflow) / dt_days,
             "hfls": latent,
             "hfss": sensible,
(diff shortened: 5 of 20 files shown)


chrimerss:
@licm13 Thanks for submitting the first PR! below is the requested changes before merge. Please review and feel free to push back if any of these does not make sense.

**Verdict: request changes.** The probe is well built and the gate is green, but two things I could reproduce need fixing before merge, and the README/CONTRIBUTORS edits the author deferred have to land in the same PR.

## What I checked (local Mac, fresh venv on the PR head 831434c, based on current main)

| Check | Result |
| --- | --- |
| `ht validate` | 15 probes, 28 models, passes |
| `pytest -q` | 110 passed (the author's 18 Windows failures are environmental, as stated) |
| `ht gate --probe energy/latent-heat-et-consistency` | passes, 2.8 s |
| `ht gate` (whole suite) | passes, 80 checks, no regression on the 14 existing probes |
| must_fail robustness, 20 random seeds each | all four broken models fail on 20 of 20 |
| New probe on our models | `reference_bucket`, `flex_lumped`, `flex_topo`, `sacsma_snow17`, `dhbv2` (Docker), `google_flood_forecast` (Docker): all INCOMPLETE, `missing: hfls, hfss, hfg` (Google also misses the water variables). Exit code 1, as the harness defines. |
| GitHub CI | no checks reported on the branch — the fork's workflow run is waiting for maintainer approval |

## Blocking

### 1. Missing `min_window_days`: the exact model fails on the window a submitted model gets

The probe sets no `min_window_days`, so a submitted model is scored on the default 30-day flood window. On that window even `reference_coupled` fails:

```
ht run --model reference_coupled --probe energy/latent-heat-et-consistency --gate-seeds --window 30
  FAIL  non_degenerate     mrro is nearly constant (cv 0.0302 < 0.1)
```

and the phase-change half of `flux_identity` has 3 sublimating steps to score instead of ~400. A real coupled model identical to the reference would be given VIOLATION. `energy/pet-consistency` and `mass/runoff-bounds` set `min_window_days: 3650` for the same reason; this probe needs it too.

### 2. The "dry frozen day" inference attributes every pack loss to sublimation

Case 2 of `flux_identity` treats any decrease in `snw` on a day with no precipitation and `tas < 0` as sublimated mass, and asserts `LE = lambda_v L + lambda_s S` as an equality. That is not "what the model reported": a snowpack can also lose water at its base. Snow-17's `DAYGM` is exactly that term, and our own `sacsma_snow17` port runs it at 0.1 mm/day.

I took `reference_coupled` unchanged except for a constant ground melt into the soil, with LE still exactly `lambda_v * liquid + lambda_s * sublimation` (script: `groundmelt_probe.py`):

| Honest model | `flux_identity` on the 5 gate seeds |
| --- | --- |
| reference, no ground melt | pass, worst 0.00 of tolerance |
| ground melt 0.3 mm/day | **fail on 375–429 of 3650 steps, worst 2.3 of tolerance** |
| ground melt 0.1 mm/day (Snow-17 default in our port) | pass, but worst 0.8 of tolerance |
| no sublimation, ground melt 0.3 | **fail on ~400 steps** |

The PR's own principle is "a criterion that fails an honest model whose snow physics differs from the reference's is worse than no criterion". This one does, and the margin at the repository's own Snow-17 setting is 20 percent.

Suggested fix, which keeps the discrimination: add a sublimation flux to the vocabulary (`sbl`, the CMIP name, mm day-1). When a model reports it, assert the equality `LE = lambda_v (E - sbl) + lambda_s sbl` at every step, snow state or not. When it does not, assert only the interval `lambda_v E <= LE <= lambda_s E` everywhere a pack is or could be present, and drop the pack-difference inference. `reference_sublimation_blind` and `reference_coupled` then report `sbl` and the gate is unchanged; a model that declines to report its split is bounded rather than guessed at.

## Needs to land in this PR (the author asked which way; my answer)

- README probes table: a row for the new probe. README reference-models table: five rows. `reference_bucket … must pass every probe` and the three `PASS, 14 of 14` rows are now stale: the water-only models are INCOMPLETE on 1 of 15. Say so, the same way `reference_streamflow_only` is described.
- `CONTRIBUTORS.md` probes table: a row for Changming Li (SCUT). A merged probe earns co-authorship, so the row is not optional.
- After merge, `models/result.csv` needs the INCOMPLETE rows for the six models above (I have the JSON reports; happy to archive them).

## For the maintainer to decide, not a defect

`flux_identity` at 0.5 percent fails any model that uses a constant latent heat of vaporisation, and `reference_constant_lambda` exists to make sure it does. That is the convention of most land-surface schemes (CLM, Noah-MP, JULES use constants; VIC uses the Brutsaert linear form). Those models are internally consistent; the criterion asserts a specific thermodynamic parameterisation of lambda(T), not a conservation law. Either position is defensible, but the probe README should state that the benchmark chooses to call the constant a violation, because it will be the first objection from any LSM group.

## Minor

- No unit tests for `criteria/coherence.py`; the gate is the only coverage. A test on a synthetic table (like `tests/test_states.py`) for the three cases of `flux_identity` would have caught item 2.
- `energy_closure`'s floor is applied as `relative <= threshold or mean_abs <= floor`, which matches `closure`'s semantics. Fine, noting it for the record.
- The test scoping in `tests/test_window.py` is correct and makes the tests robust to probe ordering.

chrimerss (changes requested):
please see the overall comment

licm13:
Thank you — this is a better review than the PR deserved, and the DAYGM catch in particular is one I should have found myself. Both blocking items reproduced locally before I changed anything, and both are fixed in e67798f. Nothing here is pushed back on.

## 1. `min_window_days` — fixed

Reproduced exactly:

    ht run --model reference_coupled --probe energy/latent-heat-et-consistency --gate-seeds --window 30
      FAIL  non_degenerate     mrro is nearly constant (cv 0.0302 < 0.1)
      ok    flux_identity      ... 3 steps report sublimation

Set `min_window_days: 3650`, as `energy/pet-consistency` and `mass/runoff-bounds` do. The same command now widens to a 3650-day window and passes every criterion, with 861 sublimating steps instead of 3. The reasoning is in the probe.yaml comment and the README: the expectations here hold over the record, so the record is what a model is judged on.

## 2. The sublimation inference — removed, and your fix adopted as proposed

I reproduced your ground-melt result to the digit. `reference_coupled` unchanged except for a constant ground melt into the soil, latent heat still exactly `lambda_v * liquid + lambda_s * sublimation`:

| Ground melt | old criterion |
| --- | --- |
| 0.0 mm/day | pass, worst 0.00 |
| 0.1 mm/day | pass, worst 0.77 |
| 0.3 mm/day | **FAIL on 375, 414, 399, 429, 408 steps of 3650**, worst 2.32 |

You are right and the argument is the PR's own: a criterion that fails an honest model whose snow physics differs from the reference's is worse than no criterion. Mine did, with 23 percent of the tolerance to spare at this repository's own Snow-17 setting. The inference is gone.

`sbl` joins `FLUX_VARS` and `UNITS` as mm day-1, documented as the sublimating share of `evspsbl` — a component of it, never an addition to it, so no budget double-counts. Then, exactly as you proposed:

- **model reports `sbl`** → equality at every step, pack or no pack: `LE = lambda_v(T) (E - sbl) + lambda_s sbl`
- **model does not** → equality at `lambda_v` where it reports no pack and none could fall; interval `lambda_v(T) E <= LE <= lambda_s E` wherever a pack is or could be present

One addition to your sketch, because the equality branch has a way to be gamed that the interval branch does not. The reported split is a claim about the model's own evaporation, so the criterion requires it to be one: not negative, not larger than the evaporation it is a share of, and zero where the model itself reports no snow and none could fall. Without that last condition a warm-season model could report a fictitious sublimating share to bend its effective lambda upwards — at 35 degC `lambda_v` is below 2.45e6, so a constant-lambda model has room to buy. Two of the new unit tests cover this.

The same ground-melt model under the criterion as it now stands, in both postures:

| Ground melt | reports `sbl` | declines to report |
| --- | --- | --- |
| 0.0 mm/day | pass, worst 0.00 | pass |
| 0.1 mm/day | pass, worst 0.00 | pass |
| 0.3 mm/day | pass, worst 0.00 | pass |
| 1.0 mm/day | pass, worst 0.00 | pass |

Zero slack rather than a margin, because the equality now uses the model's own split.

All five reference models report `sbl`, so the gate is unchanged and `reference_sublimation_blind` is now caught on every sublimating step rather than only the dry frozen ones.

**The cost, stated in the README rather than buried.** A model that declines to report `sbl` cannot be caught on the phase change at all — the interval admits a sublimation-blind model exactly on its lower bound. I think that is the right price: without the model saying which kilograms were ice, blindness and a draining pack base are not distinguishable from outside, and guessing is what we just removed. Worth knowing when the first submitted model lands there.

## 3. README and CONTRIBUTORS — landed here, as you asked

- README probes table: a row for the probe.
- README reference-models table: five rows.
- The physical models no longer "must pass every probe" but "every probe that can ask them anything". `reference_bucket`, `flex_lumped`, `flex_topo` and `sacsma_snow17` are all **14 of 15, INCOMPLETE on the energy probe**, described the way `reference_streamflow_only` already is.
- CONTRIBUTORS.md: a row in the probes table.

Your `dhbv2` and `google_flood_forecast` rows I left for the result.csv update you offered to do.

## 4. Unit tests — added, and they earned their keep

`tests/test_coherence.py`, nine cases on hand-written tables: the exact identity, a constant lambda, vaporisation applied to sublimation, **a pack losing water at its base**, a model reporting no split held to the interval, snow-free steps still an equality without a split, a model with no snow state at all, and the two malformed splits. The ground-melt case is the counterexample a green gate could never have produced, which was your point.

Writing them found a real defect the gate had not: `bounded` was computed as `(~reported_split) & pack_possible` with a Python bool, which is a bitwise inversion of an int rather than a negation. Diagnostics only, but wrong, and fixed.

## 5. The constant-lambda question — README now states it as a choice

You are right that this will be the first objection from an LSM group, and that either position is defensible. The probe README now says so in its own section rather than leaving it to be discovered: the criterion asserts a specific parameterisation of `lambda(T)`, not merely that the two budgets agree; a constant-lambda scheme is internally consistent and is being failed on a 3.4 percent thermodynamic error; the benchmark chooses to call that a violation, because 3.4 percent is two orders of magnitude above what an exact model achieves here and because a tolerance loose enough to admit a constant lambda is loose enough to admit a lambda that is not a lambda at all.

I have also written down the exit, so the decision is cheap to reverse: `lambda_vapour: [2.45e6, 0.0]` with `rel_tol` raised to the suite's 5 percent turns this into a coherence-only probe, and `reference_constant_lambda` would move to whichever criterion replaces it. Your call — say the word and I will make the change in this PR rather than after someone meets it in a report.

## 6. `energy_closure`'s floor, and the test scoping

Noted, and thank you for checking the semantics against `closure` rather than assuming.

## 7. Copilot's finding, which the rewrite happens to fix

Worth answering rather than closing silently, because it was right about the old code. The bot noted that the docstring promised a model reporting no snow state would have the split disabled, while `inferable` defaulted to all-True and so held such a model to the liquid equality on every step — an implicit assumption that it never sublimates, which is exactly the class of error the DAYGM case is about.

In the rewritten criterion a model that reports no snow state has `pack_possible` all-True and, absent `sbl`, is held to the interval everywhere. That is what the docstring always said and now what the code does. `test_a_model_with_no_snow_state_is_bounded_everywhere` pins it rather than leaving it to be re-broken: the liquid and the sublimating extremes both pass, a value outside the interval does not, and the diagnostic reports every scored step as bounded.

## CI

No checks reported because the fork's workflow run is waiting for maintainer approval — nothing I can do from here. Locally on the new head: `ht validate` 15 probes / 28 models, `ht gate --probe energy/latent-heat-et-consistency` green, `ht gate` whole suite green, `pytest -q` 18 failed / 101 passed — the same 18 Windows-environmental failures as pristine `main`, and 9 more passing than before.

PR2 (`energy/evaporative-partition`) is not open yet and I have kept its branch in step: `reference_ground_dodge` reports `sbl` too, and its `flux_identity` block carries the same `sublimation: sbl`. It already sets `min_window_days: 730`, and I checked that `--window 30` widens to it and passes.

PR #11, opened 2026-09-08
[PROBE: mass] Add time-origin invariance with aligned calendar shifts

### LINKED ISSUE #10: [PROBE] mass/time-origin-invariance: paired calendar shifts
### Conservation law

mass

### Residual equation

I'd like to contribute the roadmap's mass/time-origin-invariance probe, including implementation, validation and the methods write-up. My background is in ML evaluation and reproducible testing; I'd need your team's review of the hydrology assumptions before settling the scope.

For identical forcing, initial state, catchment parameters and seasonal inputs, changing only the calendar year should preserve outputs at corresponding elapsed timesteps.

For x in {evspsbl, mrro, mrso, snw, canopy}, define r_x(t) = x_shifted(t) - x_control(t). Flux differences have units mm/day; state differences have units mm. Compare row positions after identical spinup.

Also check each run's water budget: R(t) = pr(t) - evspsbl(t) - mrro(t) - [S(t) - S(t-1)]/dt, where S = mrso + snw + canopy and R has units mm/day.

Proposed starting tolerance, pending reference-model validation and physics review:

max_t |r_x(t)| <= 1e-9 * max(mean_t |x_control(t)|, 1 native unit), where a native unit is 1 mm/day for fluxes or 1 mm for states. This provides an explicit 1e-9 native-unit absolute floor near zero. The strict tolerance tests an exact symmetry; it is not a measured result.

The current criterion uses a 1e-12 denominator floor. I'd discuss configurable absolute-tolerance handling and report exact-model deviations before fixing the threshold.

For separate water-budget closure, use |sum_t R(t) dt| / sum_t pr(t) dt <= 0.05. Require positive accumulated precipitation in the scored window; zero-rain cases cannot pass through an undefined denominator.

reference_calendar is the main negative control: its calendar-year-dependent recession factor can conserve mass while changing the paired outputs. It must fail invariance specifically. reference_bucket should pass every declared criterion.

Zero or constant outputs can satisfy equality without useful hydrology. Retain non_degenerate checks and require reference_degenerate to fail there. Verify the declared failure mapping across seeds and reject missing or misaligned output rows. A passing result establishes this symmetry only, not complete physical correctness.

Start from the invariance template. Generate seeded daily precipitation, temperature and PET with wet/dry periods and varied storms. Reuse exactly the same numerical forcing, static attributes, initial state and model random seed for both variants.

Use 4,015 rows: 365 spinup plus 3,650 scored steps. Start calendars on 2000-01-01 and 1972-01-01. The 28-year shift preserves month/day, day-of-year, weekday and leap-day positions over this window; these calendar properties were checked with Python datetime.

Assert equal non-time inputs, row counts, daily intervals, seasonal features and spinup treatment. Exclude year-specific changes in land use, infrastructure or physical parameters. Resolve any adapter-derived date-dependent inputs with the maintainer before interpreting failures.

Deliver one seeded generator, probe.yaml, reproducibility checks, CPU reference validation and a methods/limitations write-up. No model runs or probe implementation have been completed.

Could you confirm availability, the proposed scope under the merged-probe authorship policy, and the intended paper timeline and venue?

### Will you build it?

Yes — assign it to me

SiavashShams:
Closes #10

## What this probe asserts

Adds `mass/time-origin-invariance`: identical generated weather and catchment inputs are replayed from 2000-01-01 and 1972-01-01. The fixed 4,015-day records preserve month, day, weekday, day of year and leap-day positions. Only the calendar origin changes; forcing values, static attributes, initialization contract and model seed stay the same.

After spinup, each required flux or storage must satisfy `max(abs(shifted - control)) / max(mean(abs(control)), 1e-12) <= 1e-9`. Optional groundwater, channel storage and groundwater exchange are compared when reported. The existing `closure` criterion checks the control water budget within 5% of input precipitation, and `non_degenerate` rejects trivial control outputs. No new denominator floor or paired closure criterion is introduced.

Nonfinite values reaching invariance now produce a scored FAIL naming the variable and variant, preserving the rest of the scorecard. Invalid scale factors and finite arithmetic overflow still raise errors. Regression tests cover both direct criterion calls and full probe scoring.

Ten scored years exercise repeated seasons and leap transitions. The README explains this fixed horizon, deterministic replay tolerance, control-only closure, synthetic rain-only scope and limitations. Submitted models receive at least 365 scored days so the annual runoff-ratio guard runs; the runtime allowance is 300 seconds per variant.

## Discrimination

Across all three gate seeds, `reference_bucket`, `flex_lumped`, `flex_topo` and `sacsma_snow17` pass. `reference_calendar` fails invariance, `reference_leaky` fails closure and `reference_degenerate` fails the non-degenerate guard. A regression separately verifies that the calendar model's shifted water budget closes, isolating a fault that closure alone misses.

## Validation

- `ht validate`: 17 probes and 29 models pass validation.
- `pytest -q`: 248 tests pass, including the latest upstream documentation/archive checks.
- `ht gate`: all 17 probe acceptance gates pass.
- Docker: `reference_bucket` passes all three gate seeds over both full-record variants. With a two-CPU quota, the complete six-run probe batch takes 3.081 seconds on local Docker Desktop ARM64; GitHub runner hardware has not been measured.
- Real CLI archive rows added for the new probe: all three physical models and `dhbv2` PASS; `google_flood_forecast` is INCOMPLETE because it lacks required budget outputs. The latter is a contract result, not a model execution failure.
- README, roadmap, contributor credit, archived model counts and site in English, Spanish and Chinese are synchronized. The new flowchart label was visually checked in all three languages.

## Checklist

- [x] Accepted proposal #10 is linked.
- [x] Validation, full local tests and full acceptance gate pass.
- [x] Generator is deterministic and commits no generated data.
- [x] Complete reference probe takes under one minute with a two-CPU Docker quota locally.
- [x] Tolerance, denominator, test horizon and limits are explained.
- [x] Closure-conserving calendar reference fails the new invariance check.

GitHub Actions still needs maintainer approval; local verification does not establish CI success.


chrimerss:
Thanks @SiavashShams , this is a careful piece of work. The probe itself is right: the 28-year shift keeps every calendar feature aligned over this window (I checked the leap days and the 1,461-week arithmetic), the forcing is byte-identical between variants, and `reference_calendar` is caught by exactly the criterion it was built to be caught by. Gate, validate and the full suite all pass for me locally. I'd like one round of revision before merging, mostly to make the PR smaller.

## Requested changes

**1. Drop `paired_closure` and use plain `closure` with `reference_leaky: closure`.**
On this probe `invariance` already requires every closure input (`evspsbl`, `mrro`, `mrso`, `snw`, `canopy`, and the optional `gw`, `channel`, `gwex`) to match between variants to 1e-9 on identical forcing. So `paired_closure` can only fail where `closure` on the control also fails, and `reference_leaky` leaks 15% of rain on every step regardless of calendar, so it trips plain `closure` too. That removes the wrapper, `tests/test_paired_closure.py`, and three smaller issues that come with it (the `variants` default is wrong under `variant_selection: native_and_finer`, `max()` over a NaN residual is order-dependent, and `report.py` only recognises a criterion literally named `closure`, so the archive would record `all criteria pass` instead of the residual). If a future probe with genuinely different variants needs per-variant budgets, the right place is a general variant selector in `evaluate_criteria`, and I'd rather take that as its own PR.

**2. Drop `normalization_floor`, or at least don't ship it here.**
I ran the gate with the floor removed: all four physical models replay bit-identically (max deviation exactly 0.0 on every variable including `snw`), and the must_fail set trips the same criteria. So no baseline exercises it. If we do want an absolute tolerance later, `atol` in the numpy `atol + rtol*scale` form says it directly rather than as a product of two knobs, and it would deserve its own discussion since the module's premise is that invariance has no tolerance to tune towards.

**3. Don't turn non-finite output into a harness ERROR.**
The new `_deviation` raises on NaN/inf, which `run_probe` reports as `verdict FAIL, reason ERROR` with an empty scorecard. On `main`, `inf` was a scored VIOLATION on `invariance` with the other criteria still recorded; on this branch it is indistinguishable from a broken adapter, `ht run` exits 2 (which the model-eval workflow treats as a harness failure), and a `must_fail` entry can never name it. The NaN half is an improvement over the old silent pass, so keep the check, but return a FAIL `CriterionResult` that names the variable and variant rather than raising. The overflow-of-finite-inputs case can stay an exception if you want, though I doubt any real model reaches it.

**4. `max_runtime_s: 60` → `300`.**
The budget is per run. With `min_window_days: 365` a submitted Docker model gets 730 rows per variant, about twice what `catchment-closure` runs under its 60 s, and `google_flood_forecast` already needs ~19 s for 395 rows on a fast machine. Every other paired or year-long probe uses 300; I'd match `area-invariance` here unless you've timed a container on a two-core runner (the checklist item is still unchecked).

## Small things

- `test_event_window_preserves_pairing_and_annual_guard` never looks at `non_degenerate` diagnostics, so it passes even when the annual runoff-ratio check is skipped. Asserting `"runoff_ratio_check" not in diagnostics` for `reference_bucket` would make it test what its name says.
- The README probe table and counts, the `reference_calendar` row ("waiting for an `invariance` probe"), `ROADMAP.md` (still "unclaimed"), `site/index.html` (still "wanted" in three languages), and `CONTRIBUTORS.md` need updating. I can do these on merge if you'd prefer, just say so.
- `period_years: 10` is fine, but a sentence in the README on why ten rather than the three the sibling generator uses would help the next reader.

Once 1 to 3 land this is a three-file probe plus its tests, which is what I'd hoped for. Happy to talk through any of these on the issue if you disagree with the reasoning on (1) in particular.

chrimerss:
@SiavashShams thanks, the probe is merged. One housekeeping request for the paper: could you send me your full name as it should appear, your affiliation (with department and city), an institutional or preferred contact email, and your ORCID if you have one? I'll add the affiliation and ORCID to probe.yaml, CONTRIBUTORS.md and CITATION.cff. Per CONTRIBUTING.md one merged probe is co-authorship on the benchmark paper, and we'll circulate the author list for approval before submission; let me know if you'd rather not be listed.

SiavashShams:
> @SiavashShams thanks, the probe is merged. One housekeeping request for the paper: could you send me your full name as it should appear, your affiliation (with department and city), an institutional or preferred contact email, and your ORCID if you have one? I'll add the affiliation and ORCID to probe.yaml, CONTRIBUTORS.md and CITATION.cff. Per CONTRIBUTING.md one merged probe is co-authorship on the benchmark paper, and we'll circulate the author list for approval before submission; let me know if you'd rather not be listed.

Thanks for merging this and for the helpful review! I’d be happy to be included on the paper.
My full name is Siavash Shams, and my contact email is [ss6928@columbia.edu](mailto:ss6928@columbia.edu). I’m a Columbia University M.S. alumnus, Department of Electrical Engineering, New York, NY, USA

PR #12, opened 2026-09-09
[PROBE: energy] energy/evaporative-partition — the latent-to-sensible shift under a seasonal drydown

licm13:
Part of #3

The second of two, and it depends on #9, which is merged: it reuses that PR's flux vocabulary, its `energy_closure` criterion, its `flux_identity` criterion and its reference models, and adds one more model of its own.

Not on ROADMAP. I am proposing it because of a model I had to build for #9 and then could not catch with it.

## What this probe asserts

Two runs of one seed. Net radiation, air temperature and evaporative demand are byte-identical between them. One of them loses 120 days of summer rain.

Because the radiation did not change,

    sum_W (rn_d - rn_c) == 0    by construction

and conservation forces

    R3 = sum_W [ (hfls_d - hfls_c) + (hfss_d - hfss_c) + (hfg_d - hfg_c) ]
       = 0                                                   [W m-2 day]

The latent heat a drying surface gives up has nowhere to go but the sensible and ground fluxes. This is an identity, not a parameterisation.

#9 asks whether a model's two budgets agree on one number. **This asks whether they agree under a change neither of them has seen.**

`partition_shift` asserts three things, and the second and third are what make it a test rather than an arithmetic restatement of a budget already scored:

    responded    sum_W dLE <= -0.10 * sum_W rn
    conserved    |dLE + dH + dG| <= 0.05 * max(|dLE|, 0.10 * sum_W rn)
    went to air  |dG|           <= 0.20 * max(|dLE|, 0.10 * sum_W rn)

The response gate supplies the closure gate's denominator, so it cannot approach zero. The three are reported separately, because "did not respond", "responded but lost the energy" and "responded but put it in the soil" are different findings about a model. The criterion also reports the change in evaporative fraction, LE / (LE + H), which is the form this literature reads in.

## Why this is a second probe and not a criterion added to the first

Because of `reference_ground_dodge`, new in this PR.

Its latent heat agrees with its evaporation at every step. Its water budget closes to floating point. So does its energy budget. **It passes every criterion of `energy/latent-heat-et-consistency`** — I ran it, it is a clean PASS. Its sensible heat flux is a fixed share of net radiation that never reads the soil, so when the soil dries the ground flux absorbs the entire shift:

    FAIL  partition_shift   hfg absorbed 1.00 of the shift (limit 0.20); the
                            energy went into the soil instead of the air
                            [dhfls -1266, dhfss +0, dhfg +1266 (W m-2 day),
                             sum -2.3e-13, dEF -0.112;
                             intensification stretch (45 steps) dEF -0.204]

Over four months that is a claim about soil heat capacity that no soil has. It is also the dodge available to any model that knows a partition probe is watching only the sum, which is why the criterion bounds the ground share separately. Step-wise coherence and counterfactual coherence are not the same property, and this model has the first without the second.

`max_ground_share = 0.20` is the one bound in either probe that is a physical judgement rather than an identity. Over a four-month window the ground heat store is a small residual of the surface energy budget; G/Rn is a bounded fraction that land surface schemes parameterise explicitly and that varies with cover rather than with soil moisture (Kustas & Daughtry 1990; Yang et al. 1999; Liang et al. 1999). The number is open to argument. Dropping the bound is not: without it `reference_ground_dodge` passes.

## On the name: this is a seasonal drydown, not a flash drought

The probe does not borrow that term, and the README says why. The flash drought literature settled its definition on the *rate of intensification* rather than duration (Otkin et al. 2018), and the field has since organised around the triad of rapid onset, drought development and impacts (Christian et al. 2024). Onset criteria sit at one to six pentads. A 120-day rainfall-free window is on the wrong side of that line.

The window is long because the case is slow, and that is measured:

| Days into the window | Share of the exact model's total shift |
| --- | --- |
| 1 – 60 | under 1 percent |
| by day 82 – 85 | 10 percent |
| by day 105 | 50 percent |
| days 76 – 120 | about 98 percent |

Stable across seeds. The reason is not obvious and is worth recording: the control run keeps receiving rain, so the two runs do not diverge until the drought run has spent its soil buffer. I tried shortening the lag by lowering the soil capacity to 60, 80 and 100 mm and by starting the window in July; it stays at 50 to 70 days.

So the generator marks days 76 to 120 as `_intensification`, and the criterion reports the shift over that stretch alongside the total — turning a four-month number into a rate:

    ok  partition_shift  dhfls -1266, dhfss +1266, dhfg +0, sum -2.3e-13,
                         dEF -0.333; intensification stretch (45 steps)
                         dhfls -1236, dhfss +1236, dEF -0.464

That stretch is reported and never gated. Its placement comes from the reference model — which is how this suite already locates evaluation windows — and a threshold positioned by the reference model is a threshold worth not having.

## The two lessons of #9, applied before review rather than after

**`min_window_days` is 730, and checked rather than assumed.** Any 730-day window of this three-year record contains the whole perturbed stretch, so a submitted model scored on the default 30-day flood window is widened to one that can actually see the shift. On `--window 30` the exact model is scored on 730 days and reports the same `dhfls` of -1266 as it does on the full record, with every criterion green.

**The documentation lands with the probe.** README gains a row in the probes table and a row for `reference_ground_dodge`; the four physical models move from 14 of 15 to **14 of 16, INCOMPLETE on both energy probes** — `reference_bucket`, `flex_lumped`, `flex_topo` and `sacsma_snow17` all return `missing: hfls, hfss, hfg` here too. CONTRIBUTORS.md gains a row.

`reference_ground_dodge` reports `sbl` and `channel` like the other five, and this probe's `flux_identity` block carries the same `sublimation: sbl`, so nothing here infers a phase split either.

**A third, from your review method rather than your findings.** You checked `must_fail` robustness over 20 random seeds; I ran the same check on this probe before opening the PR rather than waiting to be told. `reference_coupled` passes 20 of 20, and `reference_two_head`, `reference_ground_dodge`, `reference_energy_leak` and `reference_constant_lambda` each trip their named criterion 20 of 20. `tests/test_coherence.py` also gains six cases for `partition_shift`, since "no unit tests for `criteria/coherence.py`" would otherwise have been true a second time.

## How a model would pass this while understanding no physics

It would not. Passing requires moving a specific quantity of energy from latent to sensible in response to a water deficit that never occurred, under radiation that did not change. No training-distribution regularity supplies it, because the counterfactual half of the pair does not exist in any record. A model can be fitted to look conservative in sample; it cannot be fitted to a run that was never observed.

This is also why the probe has no real-data counterpart, and I would ask that it not be ported to that track later. The case is not a measurement; it is a comparison between a measurement and something that did not happen.

One consequence beyond the benchmark, offered as motivation rather than as part of the criterion: the latent-to-sensible transfer as soil dries is the mechanism by which the land surface amplifies heat extremes (Zhou et al. 2021; Dong et al. 2020 shows the soil-evaporation-stress operator this probe interrogates is what separates land surface models from one another). A model that fails `partition_shift` is not merely inconsistent — it is structurally incapable of representing a hot drought, however well it fits a hydrograph.

## Related work

**Conservation in learned emulators.** Beucler et al. (2021) enforce analytic constraints in neural networks emulating convection and quantify the residual "physical inconsistency" across mass, momentum, radiation and energy budgets. What is new here is not that a learned model should conserve; it is that the assertion is made about the *response to a counterfactual* rather than about a same-instant residual, which no penalty term evaluated on a validation set can reach.

**Metamorphic testing.** Asserting how an output must change when an input changes in a known way is the software-testing idea this criterion belongs to, and the suite already contains three of them: `mass/causality`, `mass/extreme-rain` and `mass/phase-counterfactual` are metamorphic relations on the water budget. This is the first on the energy partition.

## The case

Three years at PT1D with a year of spinup, 3 seeds, `variants: [control, drought]`, soil capacity 120 mm — smaller than #9's 320 mm, deliberately, since a bucket that cannot empty cannot become water-limited.

**Ninety days was my first attempt and was wrong.** On some seeds the exact model's latent flux fell by as little as 8 percent of the window's net radiation, because the soil crossed no stress threshold in either variant and the removed rain left through runoff and storage. That is a perturbation, not a regime transition. At 120 days the fall is **29 to 43 percent across seeds**, so `min_shift = 0.10` is a gate with a factor of three in hand, set below the measured range rather than fitted to it. The window is sized against the moisture-limited / energy-limited framework of Haghighi et al. (2018), used offline and never entering the scoring.

Holding radiation fixed removes a confounder, not the phenomenon. Real droughts arrive with coupled warming, which would make the shift larger rather than smaller, so the simplification cannot make an honest model look worse.

`_perturbed` and `_intensification` are stripped before the model sees the forcing, so the model is never told which stretch it is judged on.

## Discrimination

`ht gate --probe energy/evaporative-partition` — GATE PASSED, seeds [1551688172, 2058640286, 418108753].

| --- | --- | --- |
| `reference_coupled` | PASS | n/a |
| `reference_two_head` | FAIL | `partition_shift` (response gate) |
| `reference_ground_dodge` | FAIL | `partition_shift` (ground share) |
| `reference_energy_leak` | FAIL | `energy_closure` |
| `reference_constant_lambda` | FAIL | `flux_identity` |

`reference_two_head` fails before the sum is ever examined:

    FAIL  partition_shift   hfls barely moved: +0 against a required fall of
                            422.4 (W m-2 day, 10% of the 4224 the window
                            received); hfss did not rise (+0)

Its energy side has no idea its water side ran out of water, so nothing moves.

The criterion raises rather than fails if the generator ever lets the two variants drift in `rn`. The identity holds only while the driver is held fixed, and a generator that breaks that should be diagnosed as a generator fault, not reported as a model failure.

## Checklist

- [x] `ht validate` passes — 16 probes, 29 models
- [x] `ht gate --probe energy/evaporative-partition` passes, 4.1 s
- [x] `ht gate` (whole suite) passes
- [x] `pytest -q` shows no change against the pristine tree — 18 failed, 107 passed, the same 18 Windows-environmental failures as `main` on this machine
- [x] The generator is deterministic given a seed and variant, commits no data, and asserts in its `__main__` that `tas`, `pet` and `rn` are identical between variants
- [x] The probe runs in under a minute on a two-core runner
- [x] Tolerance and denominator are justified in `probe.yaml`, including why the response gate makes the closure denominator non-vanishing, and why the ground-share bound is a judgement that stays
- [x] I constructed a model that passes closure without doing physics — `reference_ground_dodge` passes every criterion of #9's probe — and `partition_shift` catches it
- [x] Unit tests for `partition_shift` in `tests/test_coherence.py`, the gap named in #9's review: the identity, a model that does not respond, one that lets the ground flux absorb the shift, one whose three changes do not sum to zero, a generator that lets the driver drift, and the intensification stretch being reported without being gated
- [x] must_fail robustness over 20 random seeds, run before submitting rather than after: `reference_coupled` passes 20 of 20 and every declared broken model trips its named criterion 20 of 20


licm13:
CI caught this before a human had to: `tests/test_docs_in_sync.py` from 86a7f3e, which landed after I opened this. Rebased onto it and brought the documents into step — ROADMAP marks the probe merged and the cross-budget section now says two rather than one, the site lists it in all three languages with a flowchart node and a translation each, every model is scored out of 16 rather than 15, and `models/result.csv` gains the five archived verdicts (`flex_lumped`, `flex_topo`, `sacsma_snow17`, `dhbv2`, `google_flood_forecast` are all INCOMPLETE here, the same way they are on the first energy probe).

That test's docstring names the first contributed probe as the reason it exists, which is fair. I could not see these failures locally: on a Windows machine with a CJK locale, `pathlib.read_text()` decodes the UTF-8 templates as GBK and 18 harness tests never run. With `PYTHONUTF8=1` the suite is 206 passed, 0 failed here, `ht gate` green and `ht validate` at 16 probes / 29 models.

PR #27, opened 2026-09-11
[PROBE: mass] Partition added and removed rain among evaporation, runoff and storage

### LINKED ISSUE #18: [PROBE] mass/precipitation-counterfactual: added or removed precipitation must be partitioned among the budget terms, in both directions
I would like to claim and build `mass/precipitation-counterfactual`. The probe runs the same seed four times: as generated, with every scored wet day 20% wetter, 10% wetter, and 20% drier, after an identical spinup. It asks where the water that was added or removed went, and whether runoff moves monotonically along the ladder from drier to wetter. It is scored over the full ten-year record.

I built it from the upstream counterfactual template at revision `45f5c9e` and ran it locally. Every number below is measured with `ht run` and `ht gate` on the three deterministic gate seeds (872466880, 1379418994, 1886371108). The exceptions are marked as inherited from the template, taken from a paper, or proposed here. Where a threshold has no literature behind it, I say so and give the measured margin instead.

Two things need your decision, and are marked as such: raising `max_share` from the template's 0.90 to 0.95, and letting `counterfactual_response` score more than one perturbed variant. Neither is needed for the gate to pass.

| Measured | Value |
|---|---|
| `ht gate`, four physical baselines must_pass, three pinned must_fail | GATE PASSED |
| Shares of the added water at +20%, over four models and three seeds (evaporation, runoff, storage) | 0.140 to 0.222, 0.754 to 0.839, 0.010 to 0.042 |
| Shares of the removed water at −20% | 0.236 to 0.372, 0.593 to 0.745, 0.010 to 0.034 |
| Runoff returned across the whole −20% to +20% ladder | 0.69 to 0.82 |
| Narrowest margin to any threshold | `sacsma_snow17` runoff 0.839 against `max_share` 0.90 |
| Physical baselines on the harness default 30-day window | FAIL, runoff share 1.9 to 2.9; hence `min_window_days: 3650` |
| `reference_in_sample` | PASS at every amplitude; not caught, stated as a boundary |

## Conservation law

Mass.

## Residual equation

The forcing carries precipitation `pr`, air temperature `tas` and potential evapotranspiration `pet`. The model must report `pr`, actual evapotranspiration `evspsbl`, total runoff `mrro` and the absolute stores `mrso`, `snw` and `canopy`. Any further store it reports (`gw`, `channel`) is counted too, and a declared exchange `gwex` enters the single-run budget as a source, as `closure` already does by default. Fluxes are rates in mm/day integrated with the actual step; stores are catchment-equivalent depths in mm.

For variant `v` over the scored window `W`, and for each perturbed variant `p` against the control `c` with `delta X = X_p - X_c`:

```text
P_v  = sum_W pr_v(t) dt          E_v = sum_W evspsbl_v(t) dt
Q_v  = sum_W mrro_v(t) dt        G_v = sum_W gwex_v(t) dt      (zero when not reported)
dS_v = S_v(end) - S_v(start-)    S   = every reported store
R_v  = P_v + G_v - E_v - Q_v - dS_v                                       [mm]

f_E = delta E / delta P     f_Q = delta Q / delta P
f_S = delta dS / delta P    f_X = -delta G / delta P    (signed export of the increment)

R_delta / delta P = 1 - (f_E + f_Q + f_S + f_X)                            [-]
```

`S_v(start-)` is the storage one row before the first scored row, as `Window.storage_initial` defines it; absolute storage is never summed over time. The denominator is the precipitation added or removed, not the control precipitation. For the drier variant `delta P` is negative and the shares keep their meaning: the fraction of the removed water each term gave up. These are absolute incremental fractions, not the relative elasticity `(delta Q / Q) / (delta P / P)`; the elasticities measured here (1.4 to 2.4) exceed one without violating anything, and none is used as an answer key.

Two budgets that each close exactly also close exactly in their difference, so the paired sum is not by itself a proof of physics. Its job is to place the residual against the added water rather than the whole rain. The share bounds carry the rest: a model that ignores the perturbation, or absorbs it in one term, fails whether or not it closes. The ladder adds direction, in both directions of the perturbation.

Five criteria score it, all of them your existing code. Three come from the counterfactual template, two from the registry.

| Criterion | Variants | What it asserts |
|---|---|---|
| `closure` | control | `abs(R_c) / P_c <= 0.05`, `gwex` counted as a declared source |
| `counterfactual_response` | control vs. wetter20 | `f_E`, `f_Q` each `>= min_share`; every share incl. `f_S` `<= max_share`; `abs(f_E + f_Q + f_S - 1) <= sum_tolerance` |
| `monotone_response` | all four, sorted by total `pr` | runoff does not fall between adjacent rungs; no rung adds more runoff than rain; at least 0.1 of the rain added across the whole ladder becomes runoff; no variant runs off more than fell plus what the catchment could hold |
| `forcing_fidelity` | control | the `pr` echoed equals the `pr` given, `rtol 1e-6`, protecting every denominator against a model that rescales its input |
| `state_bounds` | control | `mrso` in `[0, soil_capacity_mm]`, `canopy` in `[0, canopy_capacity_mm]`, `snw >= 0` |

`counterfactual_response` does not include `gwex`, so for `sacsma_snow17` the three-term sum is `1 - f_X`. Measured `f_X` is +0.0042 to +0.0049, so its sum is 0.994 to 0.996 and passes with margin. I recommend leaving `gwex` out of that sum: `closure` already accepts a declared exchange without limit, and the paired sum is the one place a declared export of the increment is still capped, at 10%.

## Tolerance and its denominator

| Setting | Value | Origin | Literature position |
|---|---|---|---|
| `closure` threshold | 0.05 of `sum_pr` | suite-wide rule | loose against model numerics (the four baselines close to 1.7e-16 here; time-stepping mass errors are the subject of Kavetski & Clark [R11]) and tight against observed closure, where residuals of 5% to 25% of precipitation are typical [R9, R10] |
| `min_share` / `max_share` / `sum_tolerance` | 0.03 / 0.90 (**0.95 proposed**) / 0.10 | template defaults | no paper sets them; Budyko gives a physical bracket, below |
| `monotone_response` 0.1 / 1.05 / 0.01 | as in `mass/extreme-rain` | criterion defaults | the inequalities `0 <= dQ/dP <= 1` follow from Budyko [R6] and from the budget itself; the slack values are yours |
| `forcing_fidelity` `rtol` 1e-6, `state_bounds` capacities | criterion default, `static.json` | numerical, physical |

**Measured margins.** Evaporation reaches 0.140 against `min_share` 0.03. Runoff reaches 0.839 against `max_share` 0.90. The sum reaches 0.994 against `sum_tolerance` 0.10. The per-amplitude shares are tabulated under the generator below; the spread across the three seeds is at most 0.02 in any share, and the single worst value is `sacsma_snow17` runoff 0.839 on seed 1379418994. No storage bound is violated, and the echoed forcing matches the input for every model tested. The closure residuals are flagged `suspicious_exact`, as the criterion is designed to do for exact models.

**A physical bracket, since no paper sets the share bounds.** In the Budyko framework the runoff share of an increment is `1 - dE/dP`. Evaluated with Fu's curve as parameterised by Zhang et al. [R7] and the sensitivities of Roderick & Farquhar [R6], at the measured aridity of the gate seeds (PET/P 0.876 to 0.960) and curve parameters 2.0 to 3.0, the expected runoff share is 0.65 to 0.75 and the evaporation share 0.25 to 0.35. The four baselines measure 0.59 to 0.84 and 0.14 to 0.37. The same framework says why no universal bound exists: as PET/P falls toward zero the runoff share tends to one, and as it grows the evaporation share does. The bounds belong to this catchment, whose attributes and generator are frozen with the probe. The measured elasticities lie inside the observed ranges for the United States, 1.0 to 2.5 [R8], and Australia, 2.0 to 3.5 [R12].

**Your decision: `max_share` 0.95.** The template's 0.90 leaves `sacsma_snow17` 0.061 of margin on one seed. In a wetter, energy-limited catchment the soil sits near saturation, runoff is generated by saturation excess, and nearly all added rain becomes runoff; the Budyko limit is a share of one, so 0.90 would fail exact physics there. I propose 0.95. It changes no verdict here, since every physical baseline is below 0.84 and `reference_degenerate` sits at 1.000 and fails `min_share` on runoff as well. The change is for robustness, not to let anything through.

**The denominator cannot approach zero.** The added precipitation over the ten scored years is 1589, 1634 and 1737 mm at +20%, exactly 20.00% of control after the template's six-decimal rounding, half that at +10%, and the same magnitude removed at −20%. The criterion raises rather than passes if the difference is zero. Values are rounded to 1e-6 mm and 3650 rows are summed, so the arithmetic precision budget is of order 1e-2 mm, five orders below the 10% allowance. No absolute floor is needed.

## How would an unphysical model fail this?

A model that reports `evspsbl` from `pet` and `mrro` from `pr` by fixed rules, then sets `mrso` to whatever closes the budget, passes `closure` on every seed forever. That is `reference_cheater`, and it is the model this probe exists for. Measured on the four-variant probe:

| Model | What it does | Shares at +20% | Sum | Verdict |
|---|---|---|---|---|
| `reference_degenerate` | evaporates everything, no runoff, constant stores | 1.000, 0.000, 0.000 | 1.000 | FAIL `counterfactual_response`: runoff below `min_share`, evaporation above `max_share`; also FAIL `monotone_response`, 0.00 returned across the ladder |
| `reference_cheater` | `ET = 0.55 PET`, `Q = 0.35 P`, storage solved as the residual | 0.000, 0.350, 0.650 | 1.000 | FAIL `counterfactual_response`: evaporation below `min_share`; also FAIL `state_bounds`, `mrso` reaches 1870 mm |
| `reference_leaky` | 15% of rain leaves the soil unreported | 0.285, 0.542, 0.024 | 0.850 | FAIL `counterfactual_response`: sum outside tolerance; also FAIL `closure` at 15.0% |
| `reference_in_sample` | exact bucket below 55 mm/day, loses 55% of the surface runoff generated by rain above it | 0.219, 0.741, 0.028 | 0.977 to 0.987 | **PASS** on every criterion |

The three `must_fail` entries are pinned to `counterfactual_response`, so a submission cannot be recorded as failing this probe for a reason another probe already covers. `reference_cheater` moves from the template's `state_bounds` mapping to `counterfactual_response`, because the measured run trips both and the paired criterion is the one that catches a cheater careful enough to keep its invented storage inside the bounds. Its shares are exactly what its code implies: unchanged PET gives no evaporation response, the fixed fraction gives 0.35, and the residual storage takes the rest.

**What the probe does not catch.** `reference_in_sample` passes at every amplitude. Its loss is conditional on daily rain above 55 mm; +20% raises the number of such days from 5 to 7 per record to 10 to 14, and the lost water is 1.3% to 2.3% of the added water, five to ten times its share of control rain but below the 10% line. Tightening `sum_tolerance` to catch it would leave the declared export of `sacsma_snow17` without margin, so I record this as a boundary rather than add the model to `must_fail`. The README states two further limits: a fixed-ratio reporter whose ratios happen to fall inside the bounds passes without dynamics, and equal and opposite errors inside one run cancel in a cumulative budget.

## How is the case generated?

The generator is the upstream counterfactual template [R0], unchanged except that the perturbation is a per-variant multiplier. All random draws happen before the variant branch; each variant scales `pr` on every row after spinup, leaving `tas`, `pet`, wet-day occurrence, event order and the spinup untouched. Variants are control ×1.00, `wetter20` ×1.20, `wetter10` ×1.10 and `drier20` ×0.80; four is the schema's `maxItems` for `case.variants`. `PT1D`, 365 spinup rows and 3650 scored rows per variant, daily timestamps from 2000-01-01 on a 365-day model year.

Measured over the scored decade, the control record carries 794 to 869 mm of precipitation a year on 91 to 95 wet days, mean wet-day depth 8.7 to 9.1 mm, largest daily total 70 to 97 mm, and PET/P 0.876 to 0.960. A third of the precipitation falls below 0 °C, so the treatment is a wetter record rather than a rainier one and snow water must be reported in `snw`. The `flex_lumped` and `flex_topo` adapters have no snow module and treat it as rain, a limitation of those baselines rather than of the probe.

**Why these amplitudes.** The template sets `PERTURBATION = 1.20` and leaves the perturbation design to the contributor [R0]. Two studies use the same proportional grid: Rasouli, Pomeroy and Whitfield perturb precipitation from −20% to +20% in 10-point steps in three headwater basins [R1, section 3.2], and Li and colleagues scale it from 80% to 120% in the same steps around a hybrid model [R2, section 2.3.2]. Neither establishes this probe's thresholds or guarantees that ±20% keeps a catchment in regime, so that was measured, worst seed per cell:

| Amplitude | reference_bucket (E, Q, S) | flex_lumped | flex_topo | sacsma_snow17 |
|---|---|---|---|---|
| −20% | 0.372, 0.593, 0.034 | 0.240, 0.745, 0.014 | 0.263, 0.727, 0.010 | 0.236, 0.725, 0.033 |
| −10% | 0.330, 0.660, 0.009 | 0.219, 0.767, 0.014 | 0.248, 0.742, 0.010 | 0.225, 0.737, 0.033 |
| +10% | 0.243, 0.730, 0.028 | 0.186, 0.793, 0.022 | 0.210, 0.775, 0.016 | 0.159, 0.796, 0.041 |
| +20% | 0.219, 0.754, 0.028 | 0.176, 0.811, 0.013 | 0.206, 0.784, 0.010 | 0.151, 0.803, 0.042 |
| +50% | 0.196, 0.777, 0.028 | 0.164, 0.820, 0.017 | 0.177, 0.814, 0.009 | 0.128, 0.827, 0.042 |

Every cell passes the template's bounds, and so does ±5%, which the README reports along with these. The runoff share rises smoothly with amplitude in all four models, the evaporation share falls, and the storage share barely moves; the differences between models at one amplitude exceed the differences between seeds, so the partition is a property of model structure rather than of the weather draw. The runoff coefficient of `reference_bucket` rises from 0.41 to 0.43 in control to 0.47 to 0.49 at +20%, and that of `flex_topo` stays below 0.57, so ±20% keeps every baseline in the regime it started in.

**The response is monotone but not symmetric, and the probe asks only for monotonicity.** At −20% `reference_bucket` returns 0.593 of the removed water as runoff; at +20% it returns 0.754 of the added water, a difference of 0.16, and the other three differ by 0.06 to 0.08 the same way. That is what saturation-excess structure does: near saturation an increment mostly overflows, while a decrement is partly buffered. Tang et al. [R5] find on 353 observed catchments that the elasticities to precipitation increase and decrease differ, and that the difference depends on aridity, so asymmetry is a property of real catchments and cannot be a failure condition. What must hold in both directions is that runoff does not fall when rain rises and that no rung returns more runoff than rain; that is what `monotone_response` checks. The sign of the asymmetry measured here is not asserted. Temporal concentration is held fixed on purpose, since Lesk and Mankin show that concentration itself changes terrestrial water storage [R3] and would confound a supply perturbation.

**The scoring window must be the whole record.** The harness scores a submitted model on the largest flood event of the window it requested, 30 days by default for a daily model. On this probe that is not a shorter version of the same test:

| Window | reference_bucket (E, Q, S) | sacsma_snow17 | Verdict |
|---|---|---|---|
| 30 days (default) | 0.000, 2.870, −1.870 | 0.000, 1.909, −0.911 | FAIL both |
| 365 days | 0.164, 0.897, −0.061 | 0.158, 0.533, 0.308 | PASS, runoff 0.003 from the limit |
| 730 days | 0.168, 0.911, −0.079 | 0.108, 0.917, −0.030 | FAIL both, `max_share` |
| 3650 days (full) | 0.219, 0.754, 0.028 | 0.151, 0.803, 0.042 | PASS |

Inside a short window the denominator is only the rain added inside it, while the runoff arriving carries the increment stored before it, so for exact physics the runoff share exceeds one and the storage share goes negative. Because the window sits at the largest flood, its position moves with its length, which is why the sequence is not monotone in length; 1825 days also passes. The probe therefore sets `min_window_days: 3650`, the mechanism `mass/phase-counterfactual` uses at 1095. Zhang, Viglione and Blöschl [R4] show that streamflow sensitivity depends on aggregation time through catchment storage; the ten-year value is this probe's measured requirement, not a universal convergence time.

**Spinup.** One year is enough. Prepending 365 or 1460 extra days of independent weather, with the scored decade byte-identical, moves the shares of `reference_bucket`, `flex_lumped` and `sacsma_snow17` by at most 0.001 and those of `flex_topo` by at most 0.009. Lengthening `SPINUP_DAYS` in the generator instead moves them by up to 0.05, because the random stream shifts and the scored weather changes; that comparison is invalid, which is why the prepend design is used.

**The gate.**

```yaml
case:
  variants: [control, wetter20, wetter10, drier20]
  min_window_days: 3650

criteria:
  - closure: {denominator: sum_pr, threshold: 0.05, sinks: [evspsbl, mrro]}
  - counterfactual_response:
      control: control
      perturbed: wetter20          # a list, with the extension below
      driver: pr
      terms: [evspsbl, mrro]
      min_share: 0.03
      max_share: 0.90              # 0.95 proposed above
      sum_tolerance: 0.10
  - forcing_fidelity: {variables: [pr], rtol: 1.0e-6}
  - monotone_response: {driver: pr, variable: mrro, min_share: 0.1, max_share: 1.05, tolerance: 0.01}
  - state_bounds: {mrso: [0.0, soil_capacity_mm], snw: [0.0, null], canopy: [0.0, canopy_capacity_mm]}

baselines:
  must_pass: [reference_bucket, flex_lumped, flex_topo, sacsma_snow17]
  must_fail:
    reference_degenerate: counterfactual_response
    reference_cheater: counterfactual_response
    reference_leaky: counterfactual_response
```

`ht gate --probe mass/precipitation-counterfactual`: GATE PASSED. All four `must_pass` models pass all five criteria, and each `must_fail` model trips the pinned criterion. The worst seed decides, as the harness does. Run cost is twelve ten-year daily runs per model, four variants on each of three seeds, measured at 2 s per model with the subprocess runner and well inside the template budget of 60 s and 2 MB per run.

**Your decision: score every perturbed variant.** As written, `counterfactual_response` takes one `perturbed` name and a probe may not list the same criterion twice, so the partition check runs on the +20% pair only, while +10% and −20% are used by `monotone_response`. I would like the partition checked on all three pairs. The smallest change is to let `perturbed` accept a list, as `response_sign` in the same module already does: each named variant scored against the control with the same three checks, the drier one with a signed `delta P`, the worst pair deciding, and a single name reproducing today's arithmetic exactly. It touches no merged probe, since none uses this criterion yet, and the result is already known, since the per-pair shares are the rows of the amplitude table above. If you would rather leave the criterion alone, the probe works as scored today.

**Deliverables.** `probe.yaml`, `generate.py` and a physics README under `probes/mass/precipitation-counterfactual/`, carrying the equations, the provenance of every number, the full amplitude, window and spinup tables, the Budyko bracket and the stated limits; the README and ROADMAP entries `tests/test_docs_in_sync.py` requires; the evaluation archive rows; and, if you accept it, the `perturbed`-list extension with regression tests showing single-name behaviour unchanged.

## References and parameter provenance

R1 to R12 were checked against the Crossref record of each DOI on 2026-09-10.

**R0. HydroTuring at revision `45f5c9efe10c9b5bab59eb60a02698ccd76f11f2`:** [generator template](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/templates/generate.counterfactual.template.py) (`PERTURBATION = 1.20`, `STATIC`), [probe template](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/templates/probe.counterfactual.template.yaml) (0.03 / 0.90 / 0.10), [response](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/src/hydroturing/criteria/response.py) and [stress](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/src/hydroturing/criteria/stress.py) criteria, [closure and fidelity](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/src/hydroturing/criteria/closure.py), [harness windows](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/src/hydroturing/harness.py), [schema](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/schemas/probe.schema.json) (`variants` `maxItems: 4`), [SAC-SMA adapter](https://github.com/Flood-Lab/HydroTuring/blob/45f5c9efe10c9b5bab59eb60a02698ccd76f11f2/models/sacsma_snow17/ht_adapter.py) (`side = 0.02` exported as negative `gwex`).

**R1.** Rasouli, K., Pomeroy, J. W. & Whitfield, P. H. (2022). The sensitivity of snow hydrology to changes in air temperature and precipitation in three North American headwater basins. *Journal of Hydrology*, **606**, 127460. https://doi.org/10.1016/j.jhydrol.2022.127460. Section 3.2: −20% to +20% in 10-point steps.

**R2.** Li, B., Sun, T., Tian, F., Tudaji, M., Qin, L. & Ni, G. (2024). Hybrid hydrological modeling for large alpine basins: a semi-distributed approach. *Hydrology and Earth System Sciences*, **28**, 4521–4538. https://doi.org/10.5194/hess-28-4521-2024. Section 2.3.2: 80% to 120% in 10-point steps.

**R3.** Lesk, C. S. & Mankin, J. S. (2026). More concentrated precipitation decreases terrestrial water storage. *Nature*, **653**, 425–432. https://doi.org/10.1038/s41586-026-10487-7. Reason to hold temporal concentration fixed; not a source for the amplitude.

**R4.** Zhang, Y., Viglione, A. & Blöschl, G. (2022). Temporal scaling of streamflow elasticity to precipitation: a global analysis. *Water Resources Research*, **58**, e2021WR030601. https://doi.org/10.1029/2021WR030601. Aggregation-time dependence through storage; not a source for the ten-year value.

**R5.** Tang, Y., Tang, Q., Wang, Z., Chiew, F. H. S., Zhang, X. & Xiao, H. (2019). Different precipitation elasticity of runoff for precipitation increase and decrease at watershed scale. *Journal of Geophysical Research: Atmospheres*, **124**, 11932–11943. https://doi.org/10.1029/2018JD030129. Asymmetry on 353 observed catchments; why the probe asks for monotonicity, not symmetry.

**R6.** Roderick, M. L. & Farquhar, G. D. (2011). A simple framework for relating variations in runoff to variations in climatic conditions and catchment properties. *Water Resources Research*, **47**, W00G07. https://doi.org/10.1029/2010WR009826. Budyko sensitivities `dQ/dP = 1 - dE/dP`; source of the bracket and of `0 <= dQ/dP <= 1`.

**R7.** Zhang, L., Hickel, K., Dawes, W. R., Chiew, F. H. S., Western, A. W. & Briggs, P. R. (2004). A rational function approach for estimating mean annual evapotranspiration. *Water Resources Research*, **40**, W02502. https://doi.org/10.1029/2003WR002710. Fu's curve and its parameter range.

**R8.** Sankarasubramanian, A., Vogel, R. M. & Limbrunner, J. F. (2001). Climate elasticity of streamflow in the United States. *Water Resources Research*, **37**, 1771–1781. https://doi.org/10.1029/2000WR900330. Observed elasticity range.

**R9.** Sahoo, A. K., Pan, M., Troy, T. J., Vinukollu, R. K., Sheffield, J. & Wood, E. F. (2011). Reconciling the global terrestrial water budget using satellite remote sensing. *Remote Sensing of Environment*, **115**, 1850–1865. https://doi.org/10.1016/j.rse.2011.03.009. Observed closure residuals of 5% to 25% of precipitation.

**R10.** Lorenz, C., Kunstmann, H., Devaraju, B., Tourian, M. J., Sneeuw, N. & Riegger, J. (2014). Large-scale runoff from landmasses: a global assessment of the closure of the hydrological and atmospheric water balances. *Journal of Hydrometeorology*, **15**, 2111–2139. https://doi.org/10.1175/JHM-D-13-0157.1. Global water-balance non-closure.

**R11.** Kavetski, D. & Clark, M. P. (2010). Ancient numerical daemons of conceptual hydrological modeling: 2. Impact of time stepping schemes on model analysis and prediction. *Water Resources Research*, **46**, W10511. https://doi.org/10.1029/2009WR008896. Numerical mass-balance errors in conceptual models.

**R12.** Chiew, F. H. S. (2006). Estimation of rainfall elasticity of streamflow in Australia. *Hydrological Sciences Journal*, **51**, 613–625. https://doi.org/10.1623/hysj.51.4.613. Observed elasticity range.

## Will you build it?

Yes — assign it to me. It is built and passes `ht gate` locally. I will open the pull request from my fork once this proposal is answered.

yqingypolimi:
Closes #18

## What this probe asserts

One seed, four rainfalls after an identical spinup: the control, and every scored wet day 20% wetter, 10% wetter and 20% drier. For each perturbed variant, the change in evaporation, runoff and storage is divided by the change in precipitation. Evaporation and runoff must each take at least 0.03 of it, no term more than 0.95, and the three shares must sum to one within 0.10. Runoff must also rise monotonically along the ladder from drier to wetter.

A model that solves one budget term as the residual closes on every seed but cannot say where added water went. `reference_cheater` is that model: it closes exactly, yet fails here because its evaporation does not respond to the added rain.

Scoring uses the full ten-year record rather than the default 30-day flood window. On that window the exact physical models fail, because runoff inside it carries water stored before it opened.

## Discrimination

| --- | --- | --- |
| `reference_bucket`, `flex_lumped`, `flex_topo`, `sacsma_snow17` | PASS | n/a |
| `reference_leaky` | FAIL | `counterfactual_response`: shares sum to 0.850 |
| `reference_cheater` | FAIL | `counterfactual_response`: evaporation share 0.000 |
| `reference_degenerate` | FAIL | `counterfactual_response`: runoff share 0.000 |

The narrowest margin is `sacsma_snow17` runoff at 0.839 against 0.95. `reference_in_sample` is not caught and is stated as a limit in the README.

## Validation

`ht validate`, `ht gate --probe mass/precipitation-counterfactual` and `pytest -q` (328 passed) all pass. `models/result.csv` has one row per evaluated model; `dhbv2` ran in its container and fails `state_bounds`.

## Two changes proposed in #18

Neither was answered in the issue, so both are applied here and each can be reverted on its own.

1. `max_share` is 0.95 rather than the template's 0.90: in a wetter catchment the runoff share of added rain approaches one, and no verdict here changes.
2. `counterfactual_response` accepts a list under `perturbed`, as `response_sign` does, so all three pairs are scored. With a single name the output is byte-identical to before (`tests/test_counterfactual_response.py`).

## Checklist

- [x] There is an `accepted` proposal issue and this PR closes it
- [x] `authors` in `probe.yaml` names every author with `name`, `affiliation` and `orcid`
- [x] `ht validate` passes
- [x] `ht gate --probe mass/precipitation-counterfactual` passes
- [x] The generator is deterministic given a seed and commits no data
- [x] The probe runs in under a minute on a two-core runner
- [x] Tolerance and denominator are justified in `probe.yaml`
- [x] I constructed a model that passes closure without doing physics, and a criterion catches it

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -23,6 +23,12 @@ authors:
       Department of Civil and Environmental Engineering,
       The Hong Kong University of Science and Technology, Hong Kong, China
     orcid: "https://orcid.org/0000-0001-9771-1324"
+  - given-names: Qingyi
+    family-names: Yang
+    affiliation: >-
+      Department of Civil and Environmental Engineering (D.I.C.A.),
+      Politecnico di Milano, Milano 20133, Italy
+    orcid: "https://orcid.org/0009-0001-9836-4261"
 repository-code: "https://github.com/Flood-Lab/HydroTuring"
 url: "https://flood-lab.github.io/HydroTuring/"
 license: MIT
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -28,6 +28,7 @@ long.
 | `mass/antecedent-monotonicity` | Zhi Li (CU Boulder) |
 | `mass/phase-counterfactual` | Zhi Li (CU Boulder) |
 | `mass/time-origin-invariance` | Siavash Shams (Columbia University) |
+| `mass/precipitation-counterfactual` | Qingyi Yang (Politecnico di Milano) |
 | `energy/pet-consistency` | Zhi Li (CU Boulder) |
 | `energy/latent-heat-et-consistency` | Changming Li (SCUT) |
 | `energy/evaporative-partition` | Changming Li (SCUT) |
--- a/README.md
+++ b/README.md
@@ -68,15 +68,15 @@ not.
 
 ## The probes
 
-Eighteen: thirteen under mass, four under energy and one under momentum. Each was
+Nineteen: fourteen under mass, four under energy and one under momentum. Each was
 merged only after the acceptance gate saw it pass four physical models, a
 bucket that conserves water exactly, two hand-written FLEX models and the
 NWS's SAC-SMA with Snow-17, and fail a purpose-built broken one on the
 named criterion. A probe that fails a
 physical model is examined before the model is; that is the first thing done
-with any probe pull request. Eleven of the eighteen can be scored on a model
+with any probe pull request. Eleven of the nineteen can be scored on a model
 that reports runoff and nothing else. `ht list` prints them;
-[ROADMAP.md](ROADMAP.md#probes-we-want) has the fourteen more we want, all
+[ROADMAP.md](ROADMAP.md#probes-we-want) has the thirteen more we want, all
 unclaimed.
 
 | Probe | Law | What it asks | The broken model it catches |
@@ -94,6 +94,7 @@ unclaimed.
 | [`mass/antecedent-monotonicity`](probes/mass/antecedent-monotonicity) | mass | The same storm after a dry month and a wet one: the wetter catchment runs off more, and no more than the extra water. | `reference_cheater` |
 | [`mass/phase-counterfactual`](probes/mass/phase-counterfactual) | mass | The same water falling as rain instead of snow: timing moves, the integrated volumes may not. | `reference_sublimating` |
 | [`mass/time-origin-invariance`](probes/mass/time-origin-invariance) | mass | The same weather under a 28-year calendar shift that preserves seasons and leap days: evaporation, runoff and water stores must agree. | `reference_calendar`, `reference_degenerate`, `reference_leaky` |
+| [`mass/precipitation-counterfactual`](probes/mass/precipitation-counterfactual) | mass | The same seed 20% wetter, 10% wetter and 20% drier: the water added or removed must be partitioned among evaporation, runoff and storage, and runoff must rise from drier to wetter. | `reference_cheater`, `reference_leaky`, `reference_degenerate` |
 | [`energy/pet-consistency`](probes/energy/pet-consistency) | energy | Evaporation reaches demand when the model's own soil is wettest, stays below it, and falls when the soil is driest. | `reference_thirsty` |
 | [`energy/latent-heat-et-consistency`](probes/energy/latent-heat-et-consistency) | energy | The evaporation a model reports as water and the evaporation implied by the latent heat it reports: are they the same evaporation? | `reference_two_head`, `reference_constant_lambda`, `reference_sublimation_blind`, `reference_energy_leak` |
 | [`energy/evaporative-partition`](probes/energy/evaporative-partition) | energy | One summer without rain under net radiation that did not change: the latent heat a drying surface gives up has to warm the air. | `reference_two_head`, `reference_ground_dodge` |
@@ -121,12 +122,12 @@ goes untested.
 
 | Model | Kind | What it does | Standing |
 | --- | --- | --- | --- |
-| [`google_flood_forecast`](models/google_flood_forecast) | submitted | The mean-embedding forecast LSTM behind Google Flood Hub, at the published weights. Predicts discharge and nothing else. | **FAIL (INCOMPLETE)**, 5 of 18 probes passed. Runoff-only, so seven probes cannot ask it anything; of the eleven that ask on runoff alone it passes the runoff bounds, memory, area, causality and steady state, and fails step, extreme rain, phase, warming, dry-down, and a 0.18 mm/day dip after an added storm |
-| [`dhbv2`](models/dhbv2) | submitted | δHBV 2.0, the MHPI group's differentiable HBV: neural networks write the parameters of a bucket model that reports its stores and its evaporation. | **FAIL (VIOLATION)**, 11 of 18 probes passed. Its learned regional-groundwater term, declared as `gwex`, closes the budget to 1e-8; what remains is a learned field capacity twice the catchment's, a response to doubled rain above the rain added, a runoff depth that changes with the area it is told, and a third more runoff when snow falls as rain |
+| [`google_flood_forecast`](models/google_flood_forecast) | submitted | The mean-embedding forecast LSTM behind Google Flood Hub, at the published weights. Predicts discharge and nothing else. | **FAIL (INCOMPLETE)**, 5 of 19 probes passed. Runoff-only, so eight probes cannot ask it anything; of the eleven that ask on runoff alone it passes the runoff bounds, memory, area, causality and steady state, and fails step, extreme rain, phase, warming, dry-down, and a 0.18 mm/day dip after an added storm |
+| [`dhbv2`](models/dhbv2) | submitted | δHBV 2.0, the MHPI group's differentiable HBV: neural networks write the parameters of a bucket model that reports its stores and its evaporation. | **FAIL (VIOLATION)**, 11 of 19 probes passed. Its learned regional-groundwater term, declared as `gwex`, closes the budget to 1e-8; what remains is a learned field capacity twice the catchment's, a response to doubled rain above the rain added, a runoff depth that changes with the area it is told, and a third more runoff when snow falls as rain |
 | `reference_bucket` | exact | conserves water exactly by construction | must pass every probe that can ask it anything; INCOMPLETE on the three energy-flux probes, which need fluxes it does not report |
-| [`flex_lumped`](models/flex_lumped) | physical | lumped FLEX/HBV: interception, beta-partitioned unsaturated store, fast and slow reservoirs, triangular lag | must pass every probe that can ask it anything; **PASS**, 15 of 18, INCOMPLETE on the three energy-flux probes |
-| [`flex_topo`](models/flex_topo) | physical | FLEX-Topo: plateau, hillslope and wetland units on real Wark fractions sharing one groundwater store | must pass every probe that can ask it anything; **PASS**, 15 of 18, INCOMPLETE on the three energy-flux probes |
-| [`sacsma_snow17`](models/sacsma_snow17) | physical | the NWS's SAC-SMA with Snow-17 and a gamma unit hydrograph, ported from the legacy Fortran and checked against it | must pass every probe that can ask it anything; **PASS**, 15 of 18, INCOMPLETE on the three energy-flux probes |
+| [`flex_lumped`](models/flex_lumped) | physical | lumped FLEX/HBV: interception, beta-partitioned unsaturated store, fast and slow reservoirs, triangular lag | must pass every probe that can ask it anything; **PASS**, 16 of 19, INCOMPLETE on the three energy-flux probes |
+| [`flex_topo`](models/flex_topo) | physical | FLEX-Topo: plateau, hillslope and wetland units on real Wark fractions sharing one groundwater store | must pass every probe that can ask it anything; **PASS**, 16 of 19, INCOMPLETE on the three energy-flux probes |
+| [`sacsma_snow17`](models/sacsma_snow17) | physical | the NWS's SAC-SMA with Snow-17 and a gamma unit hydrograph, ported from the legacy Fortran and checked against it | must pass every probe that can ask it anything; **PASS**, 16 of 19, INCOMPLETE on the three energy-flux probes |
 | `reference_coupled` | exact | the bucket with snow sublimation and a surface energy budget: every kilogram converted at the latent heat of the phase it actually underwent | must pass every criterion of the three energy-flux probes; supports daily and hourly steps |
 | `reference_two_head` | broken | a water head and an energy head that never meet: both budgets close to 1e-15 and the latent heat implies an evaporation it never reported | caught by `flux_identity` |
 | `reference_constant_lambda` | broken | coherent, but converts every kilogram at the same latent heat of vaporisation | caught by `flux_identity` |
@@ -307,7 +308,7 @@ where that conversation happens, before and alongside the issues.
 
 ## Status
 
-Suite `0.1.0`, pre-release. Eighteen probes, thirteen mass, four energy, one momentum, synthetic track only. More
+Suite `0.1.0`, pre-release. Nineteen probes, fourteen mass, four energy, one momentum, synthetic track only. More
 energy and momentum probes, and the real-data track, are next. The harness runs paired cases and
 scores labelled regimes, so the generalisation probes on the roadmap —
 extrapolation in space and time, counterfactual response, invariance — are
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -153,18 +153,13 @@ one, which is the deployment case the field actually cares about. The work is
 in defending where the hull boundary sits; push one attribute out at a time,
 or you generate catchments no real place resembles and fail honest models.
 
-### `mass/precipitation-counterfactual` &middot; standard &middot; **unclaimed**
-`ht init-probe --template counterfactual`
-
-The same seed twice, once wetter. The added water must appear in the
-difference between the reported budgets, split across evaporation, runoff and
-storage.
-
-*Discriminates:* closure by construction, structurally rather than
-circumstantially. A model that solves for a budget term as the residual closes
-perfectly on every seed forever, and today only its storage bounds catch it. A
-counterfactual asks where the extra water went, which construction cannot
-answer.
+### `mass/precipitation-counterfactual` &middot; **merged**
+The same seed 20% wetter, 10% wetter and 20% drier: the water added or
+removed must be partitioned among evaporation, runoff and storage, no term
+may ignore or absorb it, and runoff must rise from drier to wetter. Catches
+closure by construction: `reference_cheater` closes exactly on every seed,
+yet its evaporation does not respond to added rain.
+Contributed by Qingyi Yang (Politecnico di Milano).
 
 ### `mass/time-origin-invariance` &middot; **merged**
 By Siavash Shams. The same weather under a 28-year calendar shift that
--- a/docs/writing-a-probe.md
+++ b/docs/writing-a-probe.md
@@ -98,7 +98,7 @@ Every one is binary.
 | `non_degenerate` | the partition and the response are non-trivial | one run |
 | `forcing_fidelity` | the model reports back the forcing it was given | one run |
 | `regime_transfer` | closure holds out of range as well as in range | labelled stretches |
-| `counterfactual_response` | added water is partitioned, not absorbed | paired runs |
+| `counterfactual_response` | added or removed water is partitioned, not absorbed; `perturbed` may name one variant or a list, each scored against the control | paired runs |
 | `invariance` | a transform the physics ignores changes nothing | paired runs |
 | `resolution_invariance` | integrated volumes agree between the same weather at two steps | paired runs at different steps |
 | `response_sign` | perturb one driver both ways, hold the rest: each response must point the way physics says, by a real share of the change in demand | paired runs |
@@ -233,8 +233,8 @@ The reference models available today:
 | `flex_lumped` | lumped FLEX/HBV from chrimerss/HydrologicModels; conservative, nonlinear partition | nothing, it must pass |
 | `flex_topo` | FLEX-Topo, three landscape units sharing a groundwater store | nothing, it must pass |
 | `sacsma_snow17` | SAC-SMA + Snow-17 + gamma unit hydrograph, the NWS operational model | nothing, it must pass |
-| `reference_leaky` | hides a silent 15% sink | `closure` |
-| `reference_cheater` | solves for storage as whatever balances the budget; runoff is a fixed share of rain | `state_bounds`, `response_sign` |
+| `reference_leaky` | hides a silent 15% sink | `closure`, `counterfactual_response` |
+| `reference_cheater` | solves for storage as whatever balances the budget; runoff is a fixed share of rain | `state_bounds`, `response_sign`, `counterfactual_response` |
 | `reference_degenerate` | evaporates all precipitation, produces no runoff | `non_degenerate`, `counterfactual_response`, `response_sign` |
 | `reference_in_sample` | exact in range, leaks outside it | `regime_transfer` |
 | `reference_calendar` | recession drifts with the calendar year | `invariance` |
--- a/models/result.csv
+++ b/models/result.csv
@@ -196,3 +196,8 @@ run_date,model,version,suite_version,runner,probe,verdict,reason,window,seeds,de
 2026-09-10,flex_lumped,1.0.0,0.1.0,subprocess,energy/surface-energy-closure,FAIL,INCOMPLETE,not run,,"does not report hfls, hfss, hfg"
 2026-09-10,flex_topo,1.0.0,0.1.0,subprocess,energy/surface-energy-closure,FAIL,INCOMPLETE,not run,,"does not report hfls, hfss, hfg"
 2026-09-10,sacsma_snow17,1.0.0,0.1.0,subprocess,energy/surface-energy-closure,FAIL,INCOMPLETE,not run,,"does not report hfls, hfss, hfg"
+2026-09-11,flex_lumped,1.0.0,0.1.0,subprocess,mass/precipitation-counterfactual,PASS,OK,full record,872466880 1379418994 1886371108,residual 0.000% of driver
+2026-09-11,flex_topo,1.0.0,0.1.0,subprocess,mass/precipitation-counterfactual,PASS,OK,full record,872466880 1379418994 1886371108,residual 0.000% of driver
+2026-09-11,sacsma_snow17,1.0.0,0.1.0,subprocess,mass/precipitation-counterfactual,PASS,OK,full record,872466880 1379418994 1886371108,residual 0.000% of driver
+2026-09-11,google_flood_forecast,0.1.0-828dfc5,0.1.0,docker,mass/precipitation-counterfactual,FAIL,INCOMPLETE,not run,,"does not report pr, evspsbl, mrso, snw, canopy"
+2026-09-11,dhbv2,0.5.4-hbv2ep100.3,0.1.0,docker,mass/precipitation-counterfactual,FAIL,VIOLATION,full record,872466880 1379418994 1886371108,"state_bounds: mrso leaves [0, 320] on 3650 steps (range 600 to 805)"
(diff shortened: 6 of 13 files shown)


chrimerss:
Thank you, @yqingypolimi . The proposal in #18 made this easy to review, and the probe does what it promised. CI is green (gate, container, 328 tests). I went through the diff and Copilot's comment and ran a few checks of my own. Below are answers to your two questions, then four changes I'd like before merging.

**Your two questions from #18: yes to both.**

- `perturbed` as a list: accepted. I scored `main`'s criterion and yours on the same runs with `perturbed: wetter20`, for all seven gate models on the three gate seeds. Status, message, value, threshold and diagnostics are identical on all 21 pairs.
- `max_share: 0.95`: accepted. It changes no verdict here, and `reference_degenerate` still fails at 1.000.

**1. A share with the wrong sign passes (Copilot's comment on `response.py` is right).** `abs(share) < min_share` accepts a term that moves against the perturbation. I edited the last scored row of real `reference_bucket` runs to set the shares of one pair:

- `wetter20` with evaporation −0.10, runoff 0.80, storage 0.30, so evaporation falls when rain is added: passes both `counterfactual_response` and `monotone_response`.
- `drier20` with the same shares, so evaporation rises when rain is removed: passes both.
- `wetter10` with runoff −0.05: passes `counterfactual_response`, and only `monotone_response` catches it.

So in this probe nothing catches evaporation moving the wrong way. The `abs` was already on `main`, but your docstring now defines each share as the fraction of the removed water that term gave up, so the sign is part of what is being asserted. Please:

- compare the signed share, `share < min_share`. I tried this on a copy of your branch: the edited cases fail, `ht gate --probe mass/precipitation-counterfactual` gives the same seven verdicts, and both new test files pass;
- for a negative share, say the term moved the wrong way rather than "barely responds";
- add a regression test with a negative evaporation share, in the style of `test_an_unresponsive_variant_fails_alone`;
- re-run `dhbv2` on the probe and update its row in `models/result.csv` if anything changes.

**2. The window table was measured with different settings from the ones the probe ships.** It was taken at `max_share` 0.90 on the +20% pair alone. With the merged settings (0.95, all three pairs, gate seeds) I get:

| Window | reference_bucket | flex_lumped | flex_topo | sacsma_snow17 |
| 30 d | FAIL | FAIL | FAIL | FAIL |
| 365 d | FAIL, drier20 runoff 0.961 | PASS | PASS | PASS |
| 730 d | PASS | PASS | PASS | FAIL, wetter20 runoff 0.996 |
| 1825 d | PASS | PASS | PASS | PASS |

`min_window_days: 3650` is still the right setting. Please re-measure the table with the shipped settings and bring the `probe.yaml` comment in line with it (365 days also fails a physical model). Please also reword "ten years is this probe's measured requirement": 1825 days passes all four, so the full record is the conservative choice, not the shortest window that works.

**3. `perturbed: []` fails with `max() iterable argument is empty`.** A short check with a clear message would do.

**4. `ruff check` flags I001 in `tests/test_precipitation_counterfactual.py`.** The `hydroturing.harness` import needs wrapping, and `ruff check --fix` handles it.

Everything else checks out. The branch is current with `main`, every place in the `AGENTS.md` merge list is updated, the flowchart has room for the new label in all three languages, and the archive rows cover the same five models as the last two probes.

One thing for the merge: please email me a contact address rather than posting it here. It stays out of the repository, but I need it for the author list.

Once these are in, I'll fast-forward `main`.

chrimerss (changes requested):
Please review my comments

yqingypolimi:
All four requested changes are now in commit 7d85d39 on this branch.

**1. Signed shares.** `counterfactual_response` now compares each share with `min_share` by sign, so a term that moves against the change fails. A negative share is reported as "moves the wrong way", while a small positive share is still reported as "barely responds". `test_a_term_moving_against_the_change_fails` covers your two evaporation cases, `wetter20` and `drier20`, each alone and within the list. In both constructed runs the evaporation share is −0.148 and the difference sits in storage, so only the signed minimum can catch them. Both cases pass under the old `abs` comparison, so the test fails without the fix. Your `wetter10` case with a runoff share of −0.05 now also fails `counterfactual_response`: `wetter10: mrro moves the wrong way (-0.050 of the added pr, minimum 0.03)`.

I re-ran `dhbv2` in its container, and it still fails only `state_bounds` with the same message, so its row in `models/result.csv` is unchanged.

**2. Window table.** I re-measured it with `max_share` 0.95, all three pairs and the gate seeds, and it matches yours cell for cell. That includes `drier20` runoff at 0.961 for `reference_bucket` at 365 days and `wetter20` runoff at 0.996 for `sacsma_snow17` at 730 days. The README now lists all four baselines, and the `probe.yaml` comment notes that 30-, 365- and 730-day windows each fail a physical model. The README now describes the full record as the conservative choice rather than a measured requirement, since 1825 days passes all four baselines.

**3. Empty list.** An empty `perturbed` list now raises `ValueError: counterfactual_response needs at least one variant under 'perturbed'` before scoring, and a test covers it.

**4. Import order.** `ruff check --fix` wrapped the `hydroturing.harness` import, and the three changed Python files are clean under `ruff check --select I,E4,E7,E9,F`.

`pytest -q` gives 331 passed, and `ht validate` and `ht gate --probe mass/precipitation-counterfactual` pass with the same seven verdicts.

`main` has moved by two commits since your review, both touching only `.github/workflows/discord.yml`. I can rebase the branch onto it if that makes the fast-forward easier.

I will email you my contact address separately, as requested.

chrimerss:
Thank you, @yqingypolimi. I re-ran everything against 7d85d39, and all four changes hold up. The edited runs now fail with "moves the wrong way", and the new test fails if the old `abs` comparison is put back. A single `perturbed` name still reproduces `main` on all 21 gate model/seed pairs, the window table matches cell for cell, and `ruff` is clean.

Merged in e8d6b23. `main` had moved by two workflow commits, so I merged rather than asking you to rebase, and your commits are on `main` exactly as reviewed.

I added one small follow-up on top, ec03d0c. The signed minimum labelled every negative share as moving the wrong way, including a share only a hair below zero: `sacsma_snow17` on a 30-day window read `evspsbl moves the wrong way (-0.000 of the added pr)`. The label now needs the share to be at least `min_share` below zero, and anything in between reads as "barely responds". No verdict changes, and `tests/test_counterfactual_response.py` pins it.

Thanks also for your contact details. They stay out of the repository; your name, affiliation and ORCID are in `CITATION.cff` and `CONTRIBUTORS.md`.

Issue #50, opened 2026-09-11
[PROBE] energy: a seven-criterion set for the energy law, and the /io terms it needs


mchen15ouedu:
### Authors

Mengye Chen, School of Civil Engineering and Environmental Science, University of Oklahoma, Norman OK, 0000-0003-2135-1670, @mchen15ouedu

### Conservation law

energy

### Residual equation

This is a proposal for the energy law as a whole rather than for a single residual. It is the criterion set I would want a model to face before I believed its energy budget, written out as seven criteria. Some of them the suite already has in some form, and I say where below; I am proposing the whole set so that the gaps are visible against the parts that are covered, and so the naming stays consistent across them.

**1. `radiative_consistency`** — the radiation the model reports has to be radiation.

```math
r_i = \mathrm{rlus}_i - \left[\varepsilon\sigma\,\mathrm{ts}_i^4 + (1-\varepsilon)\,\mathrm{rlds}_i\right] \qquad [\mathrm{W\,m^{-2}}]
```

with, alongside it, four bounds that are not residuals but hard constraints: every radiative component non-negative; albedo `rsus/rsds` bounded in [0, 1] and strictly higher over a snow-covered surface than over the same surface bare; no reflected shortwave when `rsds = 0`; and, where a model reports no skin temperature, an outgoing longwave that a grey body at a temperature bracketed by air temperature could actually emit.

**2. `surface_energy_balance`** — at the skin, every step:

```math
\mathrm{rn}_i = \mathrm{hfss}_i + \mathrm{hfls}_i + \mathrm{hfg}_i + \mathrm{hfmelt}_i + \frac{\Delta\,\mathrm{canheat}_i}{\Delta t} \qquad [\mathrm{W\,m^{-2}}]
```

The melt term and the canopy heat storage term are the part I am adding. A snow-season or forested step that omits them is not closing the same budget.

**3. `ground_heat_storage`** — `hfg` is stored in a conducting column, not destroyed. Against a reported substrate heat state `esoil` [J/m2], over any interval:

```math
r = \sum_i \mathrm{hfg}_i\,\Delta t - \Delta\,\mathrm{esoil} \qquad [\mathrm{J\,m^{-2}}]
```

and over a climate record it cannot drain energy the column has nowhere to put: the implied mean column temperature must stay physical rather than drifting without bound in either direction.

**4. `snow_energy_consistency`** — melt is not free.

```math
[...]

PR #54, opened 2026-09-12
[PROBE: energy] Check soil heat storage against layer temperature

### LINKED ISSUE #26: [PROBE] energy/soil-heat-storage-consistency: soil heat fluxes and layer temperature must agree
### Authors

Han Wang, Department of Civil and Environmental Engineering, The Hong Kong University of Science and Technology, Hong Kong, China. ORCID: https://orcid.org/0000-0001-9771-1324. GitHub: @cehw.

### Conservation law

Energy.

### Residual equation

A model can close its surface energy budget while reporting a soil temperature that never changes. I propose checking whether the heat retained in a soil layer agrees with its temperature change:

```text
r = G_top - G_bottom - C_A * (T_end - T_start) / dt    [W m-2]
```

The first case uses a fixed, homogeneous layer with constant areal heat capacity C_A [J m-2 K-1], no phase change, heat advection or internal sources. Both fluxes are downward-positive averages over the same interval: G_top is at the actual soil surface and G_bottom at the declared layer depth. T is the mean temperature of that layer in K, and dt is in seconds.

For heating and recovery separately:

```text
mean(abs(r)) <= max(0.05 * mean(abs(G_top - G_bottom)), 1 W m-2)
```

Both phases must pass. Averaging absolute residuals prevents signed cancellation, though brief errors can still fall below the phase allowance. The scale comes from the model's net boundary heat input. Small net input alone is not a physical failure.

These are starting values. In the eight current cases, recovery uses the 1 W m-2 floor, equivalent to about 10–21% of net input. Temperature output at 0.001 K and flux output at 0.01 W m-2 gave ample rounding margin. Coarser output needs checking; mismatched sampling times should be corrected rather than absorbed into the tolerance.

The reference solves a simple conductive heat equation. Two controls keep its fluxes but either freeze the reported soil temperature or halve its change. In eight local synthetic cases, both controls passed the existing cumulative and phase surface-energy criteria but failed this storage check. Formal gate and real-model validation remain to be done.

Passing does not guarantee accurate temperature or thermal inertia. For example, equal top and bottom fluxes with constant temperature pass. Boundary fluxes must come from the model's flux calculations; reconstructing one from the temperature change would make the test pass by construction.

Use 24 hours of equilibrium spinup, 12 hours of positive net radiation and 60 hours of recovery at zero net radiation, all hourly. The reference has a zero-capacity bare skin and no evaporation. Depth, thermal properties and heating vary across seeds.

The lower boundary exchanges heat with a fixed-temperature reservoir through a nonzero conductance. Initial temperature, layer depth, heat capacity, deep temperature and conductance are supplied in the case. Keep the full heating and recovery window.

### Interface and scope

Reuse `hfg` for G_top; propose `hfg_bottom` and a layer-temperature diagnostic, `tsoil_layer`. Each row carries interval-mean fluxes and interval-end temperature; the last spinup row supplies the first scored starting temperature. Flux integration must match the temperature update.

Temperature must stay outside water-storage sums. This diagnostic support could be coordinated with #22, whose skin temperature is a different quantity. Models without matching layer properties, boundaries or thermal outputs cannot be evaluated by this first version.

Would this scope, the proposed allowance and a synthetic thermal reference be suitable? I would also appreciate guidance on declaring the diagnostics and layer settings before implementation.

### Will you build it?

Yes — assign it to me. I will implement it after acceptance and assignment.

cehw:
Closes #26.

Heat retained in a soil layer must agree with its temperature change:

```text
r = hfg - hfg_bottom - C_A * (T_end - T_start) / dt
```

Heating and recovery each require `mean(abs(r)) <= max(0.05 * mean(abs(hfg-hfg_bottom)), 1 W/m²)`. Fluxes, temperature and prescribed heat capacity refer to the same fixed layer. Temperature stays outside water-storage sums.

Following the Noah-MP suggestion, the case supplies incoming radiation and a dry material, allowing native lower-face heat fluxes. This replaces the proposal's prescribed net radiation and fixed lower reservoir; the equation and tolerances are unchanged. The probe now declares its required forcing and layer inputs. Models unable to consume them are N/A (INCOMPATIBLE). The three synthetic references accept incoming shortwave/longwave and explicit layer settings; their former prescribed-Rn fallback is removed.

The normal reference passes; frozen and half-temperature reporting controls fail on all five seeds. Derived and aggregated non-finite values cannot earn a PASS. An inert zero-flux/constant-temperature case still satisfies this necessary budget condition, but is explicitly flagged as exercising no thermal response. No universal minimum warming is assumed. This is the finite-layer test discussed in #50; a whole-column test needs its own complete boundaries.

Validation after rebasing onto `4144386`:

- 814 tests passed; eight POSIX shell mock tests skipped on Windows. All 23 probe gates passed; specs validate for 23 probes and 41 models.
- 107 CLI archive rows generated: every registered model on this probe, and all probes on the three revised references. The complete evaluated-model/reference matrix covers 12 models × 23 probes. Existing upstream results for unchanged pairs are retained.
- All five saved native HRLDAS/Noah-MP outputs still pass; their frozen and half-temperature controls still fail. These are replays of the previous full-model runs, not new simulations. Maximum native phase residual is 0.000834 W/m². [Reproduction, input mapping and limitations](https://github.com/cehw/HydroTuring/blob/codex/soil-heat-storage-consistency/docs/noahmp-soil-heat-validation.md).

Reference tables, diagnostic documentation and author records are included. Native Noah-MP evidence is separate from the synthetic test fixtures; this PR does not register a Noah-MP model.

## Checklist

- [x] Accepted proposal linked above
- [x] Author, affiliation and ORCID match the proposal and author records
- [x] `ht validate` and the acceptance gates pass
- [x] Seeded generator; no committed dataset
- [x] Tolerance, denominator, time and layer conventions documented
- [x] Controls preserve surface closure but fail the storage criterion


chrimerss:
@cehw, please fix the table issue.

licm13 (changes requested):
Hi @cehw, thanks for the continued effort on this. I think the layer-scale soil heat-storage criterion is scientifically useful and provides a valuable complement to the existing surface-energy-closure probe by testing storage within a defined subsurface control volume. There are still a few issues I think we should address before merging.

1. Could you first rebase the branch onto the current `main`? This branch predates the merged radiation/input-contract work and currently conflicts with the protocol, schemas, harness, documentation, result matrix, and tests. During the rebase, please preserve the existing `rlus` support, `ts` diagnostic, and the `requires` / `needs_*` / `uses_*` contract while adding `hfg_bottom` and `tsoil_layer`.

2. There is an important numerical-safety issue in the current criterion. Even with finite raw inputs, flux subtraction or storage calculations can overflow, which could lead to cases such as `inf <= inf` being interpreted as a PASS. Please add finiteness checks for derived quantities and final metrics, together with a regression test covering this case.

3. The reference-model manifest does not currently encode its actual radiation dependency. As a result, the model can appear compatible with a probe and then fail inside the adapter. It would be good to make these compatibility requirements explicit and machine-readable.

4. I think the physical contract between the probe and the model also needs to be made more explicit. The prescribed areal heat capacity and the exact layer/control-volume definition both affect the verdict, but the probe does not currently declare these inputs, and the model metadata does not demonstrate that they were consumed. In particular, a model using a different heat capacity should generally be classified as INCOMPATIBLE rather than as a physical VIOLATION.

5. I also agree with the earlier review comments that are still unresolved. In particular, the zero-flux/constant-temperature case can pass without demonstrating an actual storage response, so some form of non-degeneracy criterion or inert-soil control would strengthen the test. The documentation should also clearly distinguish N/A (INCOMPLETE) from FAIL, and the synthetic reference model should be described as a test fixture rather than as physical validation.

Could you also reconcile the scope with the accepted ground-heat-storage item in issue #50? My preference would be for this implementation to own the layer-scale control-volume test, so that later work does not introduce a second probe with effectively the same physical criterion under a different name.

Once the branch is rebased and these points are addressed, please rerun the full test suite, schema validation, documentation synchronization checks, and the complete model–probe matrix generation, rather than only the soil-heat-specific tests.

Thanks again — I think the scientific direction is good, and addressing these contract and robustness issues should make the probe much easier to maintain and extend.

licm13 (approved):
Hi @cehw , thank you for the careful work on this and for addressing the previous comments so promptly. The revisions have resolved the blocking issues I raised earlier:

* The branch is now rebased and preserves the merged radiation and input-consumption contracts.
* Derived arithmetic and phase metrics are checked for non-finite values, with regression tests covering boundary-flux subtraction, storage-product and reduction overflow.
* The reference manifests now encode their actual radiation and static-input requirements.
* The probe declares the prescribed heat capacity, layer depth and initial temperature on which the control volume depends. Models unable to configure that layer are classified as N/A (INCOMPATIBLE), rather than receiving a physical violation against parameters they did not use.
* The time convention and layer boundary are now explicit in the adapter metadata and documentation.
* The N/A/INCOMPLETE wording, reference-model classification, reference table and archived result matrix have also been corrected.

I am also comfortable with the treatment of the zero-response case. Zero boundary fluxes and constant temperature satisfy this necessary storage identity, and I do not think there is a model-independent minimum ground-flux response that can be justified from incoming shortwave alone. Recording `zero_boundary_fluxes`, `zero_temperature_change`, and “no thermal response exercised,” together with the explicit limitation in the README, seems like the cleaner approach. A thermal-response requirement can be considered separately if needed.

Local verification on this head:

* targeted soil-heat, diagnostics, reporting and documentation tests: 252 passed;
* full suite: 814 passed, 8 skipped;
* validation: 23 probes and 41 models;
* the soil-heat gate passes and separates both declared negative controls;
* diff whitespace validation passes.

Thanks again for the thorough and very responsive revisions. I have no remaining blocking findings.

cehw:
Thank you, Changming (@licm13), for such a detailed and thoughtful review! Your comments have helped us a lot in improving the probe, and I've learned a lot from working through them. I'm really excited to be part of HydroTuring and to keep contributing!

chrimerss:
@cehw I will merge once it passes my local test. Please let me know if you are willing to be a reviewer for this project.

cehw:
Thanks for the invitation, Zhi! I’d be happy to help with reviews, especially on surface energy balance and land-surface processes, as time allows.

PR #70, opened 2026-09-13
[PROBE: momentum] stage-discharge monotonicity and the rating loop

### LINKED ISSUE #53: [PROBE] momentum/stage-discharge-monotonic
Claiming `momentum/stage-discharge-monotonic` from the roadmap.

### Authors

Yuanhang Liu, Independent Researcher, Changchun, China, 0009-0005-9633-6004, github @kawh1111

### Conservation law

momentum

### What it asserts

A model that reports both river discharge `dis` [m3 s-1] and stage `stage` [m] is
reporting one cross-section, and the two outputs must describe one
momentum-consistent rating.

The rating of steady, uniform flow is single-valued and non-decreasing in
discharge: setting the friction slope equal to the bed slope (Manning) fixes the
stage for a given discharge. Unsteadiness makes the rating a loop, traversed so
that at a given stage the rising limb carries more discharge than the falling
one, because the water-surface slope is steeper on the rise and the same depth
therefore carries more velocity. The probe holds the model's two outputs to
that, fitting no functional form.

### Residual equation

Score the window, dropping steps where either output is missing. Write
`Q_t = dis_t` [m3 s-1] and `H_t = stage_t` [m], stage measured above the reach
bed at the outlet.

Bin the steps by discharge into `K = 8` equal-count bins, so every bin holds at
least five steps. In bin `i` let `Hb_i` be the median stage, and let `Hb_i^rise`
and `Hb_i^fall` be the median stage over steps labelled rising / falling by
`sign(dQ_t)`, with a dead-band at the turning points (steps with `|dQ_t|` below
a small fraction of the bin's discharge range are unlabelled).

**Monotonicity (steady envelope).** The bin-median rating must not fall as
discharge rises:

```
R_mono = sum_i max(0, Hb_i - max_{j<i} Hb_j)   [m]
```

normalised by the stage range `dH = max_t H_t - min_t H_t` [m]. The median
removes the loop, which is roughly symmetric about it, so what is left must be
single-valued.

**Loop direction (unsteady).** At a given discharge the rising limb reaches a
lower stage, because a steeper water surface carries more discharge at the same
depth. In every bin where both limbs are present,

```
R_loop = sum_i max(0, Hb_i^rise - Hb_i^fall)   [m]
```

again normalised by `dH`. A model that reports no loop is single-valued and
passes this check (`R_loop = 0`), which is correct; the claim is only that a
loop, if present, points the right way. A model in which the loop is reversed
fails it.

Both residuals are lengths [m] divided by `dH` [m], so both are dimensionless.
Require `R_mono / dH <= 0.05` and `R_loop / dH <= 0.05`. The denominator cannot
vanish: the generator guarantees `dH >= 0.5` m over a large flood, and the
criterion fails a case with `dH < 0.5` m as degenerate rather than dividing by
it. An absolute floor of `0.02` m is added to each residual before
normalisation, against the rounding of the stage output.

* **Two heads.** A model with a water head and a separate stage head (a
  multi-task LSTM, or an adapter that takes `dis` from its physics and `stage`
  from an independent series) closes its water budget -- nothing in that budget
  references stage -- and reports a stage series that looks plausible pointwise
  while implying a non-monotone rating or a reversed loop. `rating_monotonic`
  and `rating_loop` catch it. No water- or energy-budget criterion in the suite
  examines `stage`, so none can: the contradiction lives strictly between two of
  the model's own outputs.
* **An inverted loop.** A model whose rating loops the wrong way (falling limb
  above rising) passes any check on the stage marginal distribution, and fails
  `rating_loop`.
* **A fabricated stage.** A constant stage is caught by `non_degenerate` on the
  stage CV; a stage that carries no discharge information is caught by
  `rating_monotonic` when its bin medians fall.

**Scope.** The probe establishes that the two outputs are physically admissible
together, not that the rating is quantitatively correct; accuracy would need the
reach geometry and roughness the suite does not ask a model to justify.

Synthetic, from a recorded seed. A temperate, rain-dominated record with one
large multi-day flood per seed: a storm cluster is scaled so the reach fills and
empties through most of its range, rising over at least five steps and falling
over at least ten, so both limbs are sampled in every populated discharge bin.
Channel geometry -- `width_m`, `slope`, `manning_n`, `reach_length_m` -- is
drawn per seed into `static.json`, so the reference rating is well posed and no
dataset is needed.

Daily timestep `PT1D`. The loop timescale is the event, not the step: a flood
that rises over days produces a loop that daily sampling resolves, with several
steps on each limb. A model built only for a coarser step, or that cannot
report stage, is scored INCOMPLETE, which is honest.

### Interface it needs

`stage` [m], datum the reach bed, as a new optional output. It is a diagnostic,
neither a water store nor a flux, so it must not be differenced into the water
budget. If the optional `diagnostics` output category proposed in #39 lands,
`stage` belongs there; if not, an equivalent category is needed. Channel
geometry in `static.json`. Reference models see the full record.

### Reference models

| Reference | Expected | Criterion that catches it |
| --- | --- | --- |
| `reference_rating` -- a Muskingum reach, `stage` from reach storage (prism plus wedge), correct loop | PASS | -- |
| `reference_rating_inverted` -- the same reach with the wedge sign reversed | FAIL | `rating_loop` |
| `reference_rating_drift` -- `stage` monotone in a random walk, so its bin medians fall with discharge | FAIL | `rating_monotonic` |
| `reference_flat_stage` -- constant `stage` | FAIL | `non_degenerate` |

### Will you build it?

Yes -- assign it to me.

kawh1111:
Closes #53.

## The physics

A staff gauge is the oldest instrument in hydrology, and it imposes two constraints that no water budget does.

1. **Monotonicity.** Stage is single-valued in discharge: read the reach at the same flow twice and the gauge reads the same level, and a higher flow reads a higher level. A model that computes its stage from something other than the store it claims to describe — a running maximum, a smoothed flow, a depth from a different variable entirely — decouples the gauge from the flow.
2. **The loop.** A flood wave steepens as it arrives, so at the same discharge the rising limb sits *lower* than the falling limb. The sign of the loop is physical, not conventional; a rating read backwards inverts it, and a monotonicity check alone cannot see that.

Neither failure touches the water budget, which is why no mass or energy probe in the suite can catch either one.

## The probe

`momentum/stage-discharge-monotonic`: four years of temperate weather, five multi-day storms, one reach (width, slope, Manning's *n* handed to every model through the static file). The probe requires `channel` as a state and asks for `stage` as a **diagnostic**, in metres.

| Criterion | Asserts |
| --- | --- |
| `rating_monotonic` | binned-median stage never steps below an earlier, higher-discharge bin beyond a span-scaled tolerance (floor: 2% of the rating's span) |
| `rating_loop` | at matching water in transit the falling limb reads higher than the rising one, wherever the reach resolves a loop at all |
| `non_degenerate` | the stage varies with the weather (`cv_variables: [stage]`) |

Design decisions the review will want to check:

- **Limbs are paired on the channel store, not discharge** (`abscissa: store`). The loop is a statement about how long water stays in transit; pairing at equal store compares the limbs at the quantity the hysteresis is about. The limb a step belongs to is still decided by the sign of Δdischarge. A store that never varies is no abscissa, so the exact bucket falls back to discharge.
- **The loop is read as the median of signed per-bin differences, not a sum.** Nearest-neighbour pairing puts single bins on the wrong step wherever the falling limb is sampled sparsely; a sum would fail an honest model for a pairing artifact. A genuine loop has every bin on the same side of zero, so the median is the whole separation.
- **Two "no loop to read" escapes.** A stage that is single-valued in the abscissa (measured by the worst adjacent-pair residual, `rating_scatter_m`) has no hysteresis to read; and a loop smaller than 2% of the rating's span *that is not consistently signed* (< 75% of bins inverted) is pairing noise. A genuinely inverted rating inverts essentially every bin and does not escape.
- **Tolerance floors scale with the rating's own span**, so both criteria are readable on reaches of any depth.

## What changed outside `probes/`

- **`stage` joins `ts` and `tsoil_layer` as a diagnostic, not a state.** The suite's diagnostics channel (`requires_diagnostics` / `emits_diagnostics`, the schema arrays) already carries the radiation and soil-heat readings; this PR adds `stage` to `DIAG_VARS`, gives it metres, and trusts the four rating gauge fixtures as subprocess models. A probe may require a reading and a model may report one, but no criterion can difference one into a closure — summing a length into a volume budget would open every closure test in the suite by exactly the amount the model happened to report.
- **The four physical models report stage** via Manning normal depth, `h = (Q·n / (w·√S))^(3/5)`, built from each model's own routing: `flex_lumped` and `sacsma_snow17` drain the reach's fast path plus its slow gw path; `flex_topo`'s gauge reads its channel store alone (fast stores plus the triangular lag's contents), because its slow reservoir is catchment groundwater feeding the reach over weeks — folding it into the gauge would tie the reading to the seasonal cycle instead of to the flood the reach carries, and the binned rating steps backward across seasons (0.26 m of deficit measured before the slow reservoir was left out). `reference_bucket` also emits `dis` so the fallback abscissa is honest. `flex_topo`'s channel store carries the lag's in-transit water (`cum_gen − cum_out` added back after the per-unit stores), so every event's budget closes on it and the suite's event-closure test holds.
- **Four reference models** pin the probe from both sides:
  - `reference_rating` — the bucket with a triangular unit hydrograph and a Manning stage; it loops the way a flood wave does and pins monotonicity from the passing side.
  - `reference_rating_drift` — stage read off a running maximum of discharge; the rating steps down against its own history → trips `rating_monotonic` (2.08 m deficit).
  - `reference_rating_inverted` — stage read off the channel store, which peaks after the flow does; monotone in discharge, so only the loop sees it → trips `rating_loop` (0.252 m, 12/12 bins inverted).
  - `reference_flat_stage` — a gauge that reports a constant → trips `non_degenerate`.
- **`models/result.csv`** gains one row per evaluated model (9 archived; submitted models that lack `stage` short-circuit to `N/A (INCOMPLETE)` before Docker is ever consulted).
- **Docs synced**: README (probe table, model table, flex standings 18→19, counts twenty-three→twenty-four), CONTRIBUTORS, ROADMAP (`merged`), AGENTS (the `stage` row), `docs/writing-a-probe.md` (criteria + reference-model tables), site (three languages, flowchart's second momentum dot, per-model counts 19 / 19).
- **One `#54` test updated**: `test_thermal_diagnostics.py` used `reference_bucket` as its shipped example of a manifest with no `diagnostics` key; the bucket now reports `stage`, so the default-empty check runs against `reference_leaky` — same kind of in-repo subprocess reference, and its closure-criterion role means it will never carry a diagnostic.

## Gate evidence

```
momentum/stage-discharge-monotonic  (seeds [1670932170, 30400637, 537352751])
  ✅  must_pass  reference_bucket          rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_pass  flex_lumped               rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_pass  flex_topo                 rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_pass  sacsma_snow17             rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_fail  reference_rating_drift    expected `rating_monotonic`, tripped ['rating_monotonic']
  ✅  must_fail  reference_rating_inverted expected `rating_loop`, tripped ['rating_loop']
  ✅  must_fail  reference_flat_stage      expected `non_degenerate`, tripped ['non_degenerate']

✅ GATE PASSED: every probe separates the reference models as declared.
```

Full suite on `123e941` + this PR: `822 passed, 8 skipped in 356s` (same 8 skips as `main`; 821 of those before the one-line `test_thermal_diagnostics.py` fixture swap noted above).

## Credit

`probe.yaml` carries the authorship record: Yuanhang Liu, Independent Researcher, ORCID `0009-0005-9633-6004` — per CONTRIBUTING, one merged probe earns co-authorship.


chrimerss (changes requested):
Please address the comments

kawh1111:
Thanks @chrimerss and the automated review. I have addressed all 12 findings in a new commit (`393d9c2`). CI is re-running; the local gate is green before/after the fix. Below is a per-finding summary with the evidence I collected.

### High-severity items

**P1 — `requires.fluxes` was empty, so the three physical models were being scored against `channel` storage while the reason text claimed "discharge".**
- Fix: `requires.fluxes: [dis]` in `probe.yaml`; added `dis` to `flex_lumped`, `flex_topo` and `sacsma_snow17` adapters and manifests.
- Evidence: after the fix every must-pass model now reports `source dis`, rating spans are physically reasonable (1.06 m / 2.18 m / 4.28 m), and `sacsma_snow17` reaches the real `rating_loop` PASS branch (median loop −0.85 m, 0 of 12 inverted) instead of the single-valued escape.
- The wording now adapts to the actual abscissa: "stage rises monotonically with **discharge**" when `dis` is present, and "with **the channel store**" only for the fallback.

**P2 — `loop = max(0.0, loop_signed)` made a physically correct loop report as "no loop to read", leaving the PASS branch unreachable.**
- Fix: use `loop_size = abs(loop_signed)` for escape detection and diagnostics, keep `loop = max(0.0, loop_signed)` for the directional check, and report `loop_size_m` / `loop_median_m`.
- Evidence: `reference_rating` now passes with a real loop message (`loop_size_m ≈ 0.11 m`) instead of the old "0.00% ... pairing noise" escape.

### Medium-severity items

**P3 — `reference_rating` was not in any baseline list.**
- Fix: added `reference_rating` to `must_pass` in `probe.yaml`.
- Evidence: gate now scores it explicitly; it passes after P2.

**P4 — `_rating_scatter` short-circuited on tied abscissae.**
- Fix: compute `tie = dh[coincident].max()` and gradient residual over the separated pairs separately, return `max(tie, gradient)`.
- Evidence: a synthetic rating with ties plus a real inverted loop still fails; the `tie-only` case no longer masks the loop.

**P5 — `_discharge` docstring said it returned the step-over-step change of the store, but it returned the store itself.**
- Fix: rewrote the docstring to describe the actual behavior and why the fallback returns the store itself; noted that with `dis` required this path becomes dead code.

**P6 — `stage_of` divided storage by `dt`, making stage scale with timestep and folding in the slow groundwater store.**
- Fix: pass the real runoff rate `mrro` (mm/day) to `stage_of`; use `mrro + gw/dt` only where a groundwater term is genuinely routed; drop the slow-store contribution in `flex_topo`.
- Evidence: `stage_of(5.0 mm/day)` gives the same 0.582836 m stage at PT1D, PT1H and PT1M (previously ~78× larger at PT1M).

**P6b — `rating.py` had no unit tests.**
- Fix: added `tests/test_rating.py` with 11 tests covering correct-direction loop, inverted loop, single-valued escape, store fallback, missing-stage failure, monotonic drop detection, abscissa wording, explicit tolerance override, tie handling, and near-zero scatter.
- Evidence: the new tests fail when any of P2/P4/P9/P1 is re-introduced (mutation tested).

### Low-severity items

**P7 — docs table described non-existent `rating_monotonic` implementation.**
- Fix: rewrote the row to the actual running-maximum-deficit-over-bin-medians design, noting the span-scaled floor.

**P8 — `reference_rating` descriptions said UH + Manning, but the model is two linear reservoirs + prismatic depth.**
- Fix: corrected `models/reference_rating/model.yaml`, README.md and `docs/writing-a-probe.md`.

**P9 — explicit `tolerance` could not override the span floor.**
- Fix: `tolerance = tolerance if tolerance > 0.0 else floor`; updated the comment to match.
- Evidence: `tolerance: 0.01` on a 5 m rating is now honoured instead of being widened to 0.10 m.

**P10 — probe README criterion table contradicted `probe.yaml`.**
- Fix: line 33 now says "at matching water in transit"; line 34 says "the stage varies with the weather".

**P11 — `reference_rating_drift` double-counted the step's own outflow.**
- Fix: report `"channel": store` after `store -= routed`.
- Evidence: per-step residual against the water budget went from 0.821 mm to 0.000 mm.

**P12 — docs `rating_loop` row omitted the escape semantics.**
- Fix: added a sentence describing the single-valued/noise escape.

### Verification run

```
$ ht validate
✅ validation passed: 25 probe(s), 46 model(s), suite 0.1.0

$ ht gate --probe momentum/stage-discharge-monotonic
  ✅  must_pass  reference_bucket         rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_pass  flex_lumped              rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_pass  flex_topo                rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_pass  sacsma_snow17            rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_pass  reference_rating         rating_monotonic=pass; rating_loop=pass; non_degenerate=pass
  ✅  must_fail  reference_rating_drift   expected rating_monotonic, tripped ['rating_monotonic']
  ✅  must_fail  reference_rating_inverted expected rating_loop, tripped ['rating_loop']
  ✅  must_fail  reference_flat_stage     expected non_degenerate, tripped ['non_degenerate']
```

I also confirmed that the 8 remaining `test_docs_in_sync.py` failures in my working tree are all due to the still-in-progress #71 probe (`momentum/channel-routing-mass`) and its unmerged docs/archive entries; none of them come from the #70 changes.

Please let me know if you would like the `dis` requirement relaxed or any further wording changes.

chrimerss:
Thanks for the quick turnaround on 393d9c2. Much of the first review is fixed: the tie handling in `_rating_scatter`, the tolerance override in code, the drift model's `channel`, the `reference_rating` descriptions, `flex_topo`'s gauge, and the new unit tests. CI is green and I reproduced it locally (841 tests, full gate). The three physical archive rows reproduce byte for byte on the gate seeds.

Three things still need to change before this can merge. All numbers below come from the PR head.

**1. `flex_lumped` and `sacsma_snow17` still build `dis` and `stage` from a store divided by `dt`.**

The fix passes `mrro` as a rate but keeps `max(gw, 0) / dt`. That is the same store-as-rate the first review flagged, and it is now also written into `dis`. Both models already route that store's release into `mrro`: `generated.append(qf + qs)` in flex_lumped, and SAC-SMA's `tci` includes baseflow. So the term double-counts, and it counts the whole store rather than its release.

On gate seed 1670932170:

| model | mean `dis` | own outflow (`mrro`·area) | ratio | at PT1H |
| flex_lumped | 19.8 m³/s | 3.07 m³/s | 6.5× | `dis` ×20, stage median 0.98 → 6.08 m |
| sacsma_snow17 | 117.8 m³/s | 2.48 m³/s | 47.6× | `dis` ×22, stage median 2.78 → 17.6 m |
| flex_topo | 4.03 m³/s | 4.03 m³/s | 1.00 | unchanged |

sacsma's stage correlates +1.00 with its lower-zone store. Its archived "rising limb sits below the falling limb (median −0.8549 m)" is therefore the seasonal groundwater cycle, not a routing loop. The PT1D/PT1H evidence in the reply only exercised `gw = 0`.

Suggested fix: do what flex_topo does, `dis = mrro · area / 86400` and `stage = Manning(dis)`. I checked this in process over seeds 0–99: both models then pass every criterion on every seed. Their archived rows need regenerating afterwards.

**2. No honest baseline reaches the "the loop runs the right way" branch.**

`reference_rating` on the gate seeds reports:

> no loop to read: the two limbs differ by only 1.27e-02 m (0.03% of the 41.19 m rating) and disagree in sign across bins (0 of 12 inverted), which is pairing noise rather than a loop

It takes an escape on every seed I tried. There are two causes:

- `FLOODPLAIN_RESIDENCE_D = 0.35` is shorter than the probe's one-day step (`k_fast` = 0.94), so at PT1D there is no second time constant to see.
- Stage is the slow store's volume spread over the reach bed, which gives a 41 m rating in an 18 m wide, 2.5 m bankfull reach. The 2%-of-span floor is then 0.82 m, and no loop gets past it.

Once (1) is fixed, sacsma stops reaching that branch too, so only `tests/test_rating.py` would cover it.

The escape branch also hard-codes its wording. It always says "disagree in sign across bins" and always sets `single_valued: True`, even at 0 of 12 inverted and on the noise path. Archived reasons can therefore state things that are not true.

**3. Both sides of `rating_loop` sit too close to the line.**

`ht run` draws three random seeds and one failing seed fails the probe. So this matters for the archive and for every submitted model. The counts below come from 130 seeds (0–126 plus the gate seeds), in process, with cases spot-checked through `ht run --seed`:

- `flex_topo` FAILs `rating_loop` on seeds 42, 83 and 125: the rise sits 0.010–0.034 m above the fall, with 9 of 12 bins inverted. That is about 7% of three-seed runs.
- `flex_lumped` FAILs on 50, 71, 85, 102, 110 and 126 (0.03–0.08 m, 7–8 of 12 inverted), about 13% of runs. Both models report stage as an exact function of `dis`, so this is pairing noise on the store axis, not a rating fault.
- `reference_rating_inverted` escapes on 3 of 103 seeds, for example seed 1: "channel fixes stage to within 1.14e+00 m, so the rating is single-valued — below the 1.7435 m the reach resolves".
- **The inverted control is caught by the 75% rule alone.** Its loop (0.12–0.32 m) is below its own span floor on every seed. If I drop `and not consistent` from the escape, all 11 tests in `tests/test_rating.py` still pass, and the gate then reports `reference_rating_inverted` as tripping nothing.

The FAIL side has `tolerance: 0.0` and no size floor, so a centimetre-scale consistent bias fails. Meanwhile the prism-volume stage inflates the span floor enough to hide a real inversion. A symmetric size floor alone won't fix it: the reference gauges need a physically scaled stage as well.

Please add unit tests for:
- a stage that is an exact function of `dis`, paired on a moving store;
- a small but consistent inversion.

Then report a seed sweep of all eight baselines in the PR.

**Smaller fixes for the same pass:**

- `docs/writing-a-probe.md:119` still says the deficit is compared with "the larger of an explicit `tolerance` and 2% of the rating's span". The code now honours a tighter tolerance.
- The drift model is described as a running maximum that "never comes back down" in README.md:178, docs/writing-a-probe.md:268, the probe README:36 and probe.yaml:74–75. Its gauge actually decays at `PEAK_DECAY = 0.997` per day.
- probe.yaml:20 and generate.py:11 still speak of "same/matching discharge", but the limbs are paired on the store.

**For the record (not blocking):** the implementation departs from the design settled in #53 in these ways:

- limbs are paired on the channel store with the nearest falling step, rather than on discharge bins with no turning-point dead-band;
- 12 bins instead of 8;
- a 2%-of-span floor instead of 5% of dH plus 0.02 m;
- no dH ≥ 0.5 m degenerate case, so a flat stage passes both rating criteria and only `non_degenerate` catches it;
- one fixed reach geometry instead of one drawn per seed;
- two linear reservoirs instead of a Muskingum prism plus wedge;
- a decaying peak memory instead of a random walk.

The PR explains the store pairing. The rest should be stated in the probe README so the maintainers accept them knowingly.

kawh1111:
All three blockers are addressed in `2a1e919`, plus the smaller fixes and the sweep.

**1. `flex_lumped` / `sacsma_snow17` no longer divide a store by the step.**

Both now do what `flex_topo` does: `dis = mrro * area / 86400`, `stage = Manning(dis)`. The store's release is already inside `mrro` (`generated.append(qf + qs)`; SAC-SMA's `tci`), so the extra term double-counted it and read the whole reservoir rather than what leaves it.

On gate seed 1670932170, measured through the adapters directly:

| model | mean `dis` before | mean `dis` after | own outflow | ratio | stage median before -> after |
| --- | --- | --- | --- | --- | --- |
| flex_lumped | 18.9 | 3.076 | 3.076 | 1.000x | 0.949 -> 0.259 m |
| sacsma_snow17 | -- | 2.443 | 2.443 | 1.000x | 2.78 -> 0.184 m |
| flex_topo | 4.03 | 4.141 | 4.141 | 1.000x | unchanged |

The Manning residual between the reported `stage` and `(dis*n/(w*sqrt(S)))^0.6` is `0.00e+00` for all three, so the gauge can no longer depend on the step the model writes at. At PT1H the old /dt blow-up is gone (x0.42 and x0.61, the models' own response to a finer step). The archived rows are regenerated.

**2. An honest baseline now reaches the correctly-signed-loop branch.**

`reference_rating` is rebuilt at a physical scale: the gauge is a Manning depth solved from the *channel's* release, the channel residence is 60 d against the floodplain's 8 d, and 90% of the yield runs over the floodplain. Its span is 0.08-0.30 m rather than 41 m, and over 130 seeds it reaches the "rising limb sits below the falling limb" verdict on **90** of them. The other 40 take the size escape, all PASS — the branch is live rather than dead.

**3. Both sides of `rating_loop` are off the line, and the escapes are honest.**

- The criterion now measures scatter against **both** axes. A rating that is a function of the discharge alone -- `flex_lumped` and `sacsma_snow17`, exactly -- is single-valued and has no loop to show; it escapes instead of failing on store-axis pairing noise. That is what takes `flex_topo`'s ~7% and `flex_lumped`'s ~13% failure rates to zero.
- The size floor is **symmetric** (a separation too small to resolve is not a loop in either direction) and stated in metres -- 2 mm, a staff gauge's reading error -- because the reference gauges are different instruments whose spans differ by two orders of magnitude between seeds while the loop a gauge can resolve does not.
- The escape's wording is branch-dependent: at 0 of 12 inverted it now says the sign *agrees*, and `single_valued` is set from which scatter fired rather than hard-coded.

**4. Seed sweep, 130 seeds (the gate seeds plus 0-126), at the probe's 365-day spinup window:**

| model | `rating_monotonic` FAIL | `rating_loop` FAIL | reaches "the right way" |
| --- | --- | --- | --- |
| reference_bucket | 0 | 0 | 0 |
| flex_lumped | 0 | 0 | 0 |
| flex_topo | 0 | 0 | 0 |
| sacsma_snow17 | 0 | 0 | 0 |
| reference_rating | 0 | 0 | 90 |
| reference_rating_drift | 129 | 28 | -- |
| reference_rating_inverted | 0 | **130** | -- |
| reference_flat_stage | 0 | 0 | 0 (trips `non_degenerate`) |

`reference_rating_inverted` now trips `rating_loop` on every seed with 12 of 12 bins inverted: its gauge solves Manning from the floodplain's release instead of the channel's, which leaves it out of step with the discharge.

**5. Smaller fixes.** `docs/writing-a-probe.md:119` now says the tolerance is an explicit value or 20% of span; the drift model is described everywhere as a decaying peak memory (`peak = max(q, 0.997 * peak)` per day) rather than one that never comes back down; `probe.yaml` and `generate.py` say the limbs are paired on the store, not on discharge.

**6. New unit tests**, as asked: a stage that is an exact function of `dis` paired on a moving store (must read as single-valued, not as a loop), and a small but consistently inverted loop (must fail).

**7. The #53 departures** are now stated in the probe README, including the two load-bearing ones -- the absolute size floor, and the fact that the loop and the monotonicity share are opposite constraints on the same rating, which is why the gauges had to be rebuilt at a physical scale before both could hold at once.

kawh1111:
CI caught a real error in the previous commit. This probe is archived on the **evaluated models**, not on the gate baselines, and that commit had replaced the probe's whole row set with the reference baselines -- so the six N/A rows (cwatm, dhbv2, google_flood_forecast, lisflood, summa, wflow_sbm) went missing. That fails `test_archive_has_every_evaluated_model_on_every_probe`, which in turn fails the gate job before `ht gate` itself runs.

`28b6ac8` keeps PR70's six N/A rows for the models this change does not touch and re-runs only the three whose gauge was rebuilt (`flex_lumped`, `flex_topo`, `sacsma_snow17`). `pytest tests/test_docs_in_sync.py tests/test_report_detail.py` is green again on that commit.

chrimerss:
Thanks. I re-checked 28b6ac8, and all three blockers are fixed:
- `dis` equals each model's own outflow, and it is unchanged at PT1H.
- The physical baselines read as single-valued in `dis` on 130 of 130 seeds, with no failures.
- `reference_rating` reaches the right-way branch on 90 of 130 seeds.
- The inverted control trips on all 130, with loops of 13–100 mm.

CI is green. Locally, pytest gives 843 passed and the full gate passes, and the three regenerated archive rows reproduce byte for byte.

One change before merge: **`MIN_MONOTONIC_FRACTION = 0.20` is looser than anything needs.** Over the same 130 seeds, the largest deficit an honest rating shows is 4.1% of span (`reference_rating`). The smallest for `reference_rating_drift` is 19.7%, so at 20% the drift control passes the probe on one seed.

I also scored the bucket with a shuffled `stage`. At 20% it passes the whole probe on 58 of 130 seeds (9% of three-seed runs). At 8% that falls to 22 seeds (0% of runs), while every honest baseline still passes and drift is caught on all 130. Please set the floor to about 8% and update the docs row and the README paragraph that justify 20%.

A limitation to state in the probe README: a gauge that is hydrograph-shaped but decoupled from the model's own discharge (#53's "two heads") is not caught at any floor. A stage taken from another seed's record passes on 86 of 130 seeds, and one lagged 15 days passes on 125.

Two small things:
- `reference_rating_inverted`'s docstring says it has "the same split" and "the discharge is identical", but its `FLOODPLAIN_SHARE` is 0.5 against 0.9, so the discharge differs.
- The honest loop is 2.0–5.8 mm against the 2 mm floor. That is fine, because it passes either way, but it's worth a sentence in the README.

kawh1111:
All four points are in `c8ca3d4`.

**1. The floor is 8%, and the drift control is caught on every seed.**

`MIN_MONOTONIC_FRACTION` is now `0.08`. The `rating_monotonic` row in `docs/writing-a-probe.md` and the probe README paragraph that justified 20% both cite 8% now.

Over the same 130 seeds (the gate seeds plus 0-126), scored in process against the probe's declared params:

| model | `rating_monotonic` FAIL | `rating_loop` FAIL | `non_degenerate` FAIL |
| --- | --- | --- | --- |
| reference_bucket | 0 | 0 | 0 |
| flex_lumped | 0 | 0 | 0 |
| flex_topo | 0 | 0 | 0 |
| sacsma_snow17 | 0 | 0 | 0 |
| reference_rating | 0 | 0 | 0 |
| reference_rating_drift | **130** | 28 | 0 |
| reference_rating_inverted | 0 | **130** | 0 |
| reference_flat_stage | 0 | 0 | 130 |

Every honest baseline passes every criterion on all 130 seeds, and each control trips on all 130. The largest deficit an honest rating shows is 4.1% of span (`reference_rating`) and the smallest for the drift control is 19.7%, so 8% sits between them with the margin you measured.

**2. The limitation is stated in the probe README.** A new `## Limitations` section says a gauge that is hydrograph-shaped but decoupled from the model's own discharge -- one taken from another seed's record, or lagged a fortnight -- is not caught at any floor, because the criteria only check that the gauge moves with the flow it claims to describe.

**3. `reference_rating_inverted`'s docstring no longer claims the same split.** It now says the split is 0.5/0.5 against the honest model's 0.9/0.1, so the discharge is not identical, and the fault is which path the gauge reads.

**4. The honest loop's margin over the floor is called out** in the same README section: 2.0-5.8 mm against the 2 mm floor.

The three regenerated archive rows (`flex_lumped`, `flex_topo`, `sacsma_snow17`) keep PR70's six N/A rows untouched, `pytest tests/test_docs_in_sync.py tests/test_report_detail.py tests/test_rating.py` is green on the new head, and the gate passes locally.

Issue #30, opened 2026-09-11
[MODEL] SUMMA (Structure for Unifying Multiple Modeling Alternatives)


kawh1111:
## Model name

SUMMA (Structure for Unifying Multiple Modeling Alternatives)

## Model category

physics

## Model code repository

https://github.com/CH-Earth/summa

## Version, release, or commit

v4.0.0 (release, 2026-09-04); current HEAD f787fa5e63d3c67030721a85f2f244d22a8d691e

## Weights repository (optional)

None required.

## Paper or model reference (optional)

https://doi.org/10.1002/2015WR017198 (Clark et al., 2015, WRR)

## Code and weights licenses

Code: GPL-3.0. No weights.

## Supported timestep

PT1H (hourly) - SUMMA's native solver step; can be aggregated to daily.

## Required inputs

Forcing: precipitation, air temperature, shortwave and longwave radiation, wind speed, air pressure, specific humidity.
Static: soil and vegetation parameters per HRU.
Initialization: spinup to equilibrate soil and snow states.

Note, stated plainly: the generated forcing (pr/tas/pet) does not cover SUMMA's full input list. Unless a probe supplies the extra columns (the harness passes extra forcing columns through), the adapter would have to derive radiation, humidity and wind from pet and tas, and the energy verdict would then be about the derivation rather than about SUMMA.

## Outputs produced by the model

pr, evspsbl, mrro, mrso, snw, canopy (SUMMA also reports latent, sensible and ground heat fluxes - hfls/hfss/hfg - which are not among the form's output options but can be declared in model.yaml; it does no routing, so no dis).

## Evaluation window (optional)

Default.

## Packaging status

Help is needed with both Docker packaging and the adapter (Fortran build; the largest packaging job of the models proposed here).

## Installation and inference notes

Fortran + netCDF build (Makefile in build/); documentation at summa.readthedocs.io; test cases ship with the repository. No weights. Runs on CPU; a single lumped HRU at daily output is comfortably within the probe's time budget. Inference needs no network.

## Why this model belongs in the benchmark

[...]

Issue #29, opened 2026-09-11
[MODEL] wflow_sbm (Wflow.jl, Deltares)


kawh1111:
## Model name

wflow_sbm (Wflow.jl, Deltares)

## Model category

physics

## Model code repository

https://github.com/Deltares/Wflow.jl

## Version, release, or commit

v1.0.4 (latest tagged release, 2026-07-24). The GMD paper evaluates v0.7.3; use that tag instead if the evaluation should match the paper exactly.

## Weights repository (optional)

None required.

## Paper or model reference (optional)

https://doi.org/10.5194/gmd-17-3199-2024 (van Verseveld et al., 2024, GMD)

## Code and weights licenses

Code: MIT. No weights.

## Supported timestep

PT1D (daily)

## Required inputs

Forcing: precipitation (mm/day), temperature (degC), PET (mm/day) - the generated forcing covers these directly.
Static: soil and terrain parameters; as a distributed model the adapter would aggregate them to a lumped unit or catchment-mean parameters.
Initialization: state files, or cold start with spinup.

## Outputs produced by the model

pr, evspsbl, mrro, dis, mrso, snw

## Evaluation window (optional)

Default.

## Packaging status

Help is needed with both Docker packaging and the adapter.

## Installation and inference notes

Julia package (`pkg> add Wflow`); TOML configuration and netCDF in/out; no weights. Julia runs in a container from the official images, so the adapter work is configuration templating and unit remapping. CPU-only and fast. Inference needs no network.

## Why this model belongs in the benchmark

wflow_sbm couples the SBM vertical concept - a store-and-flux scheme with an explicit groundwater reservoir - to kinematic-wave channel routing. That makes it the natural candidate for momentum/routing-conservation, the suite's loneliest probe, and its GMD paper demonstrates global-data-to-local-application deployment, so a verdict is informative to a large user base. It is also a Julia model: packaging it exercises the "any language" claim of the /io adapter contract with a second language family.

## Submission confirmation

[x] I have identified a reproducible code revision and have permission to use the linked code and weights for this evaluation. (MIT permits use, reproduction and modification; v1.0.4 is the pinned revision.)
[x] I understand that HydroTuring may report PASS, VIOLATION, INCOMPLETE, INCOMPATIBLE, or ERROR.

Issue #28, opened 2026-09-11
[MODEL] CWatM (Community Water Model, IIASA)


kawh1111:
## Model name

CWatM (Community Water Model, IIASA)

## Model category

physics

## Model code repository

https://github.com/iiasa/CWatM

## Version, release, or commit

1.11 (latest tagged release as of 2026-09-11)

## Weights repository (optional)

None required.

## Paper or model reference (optional)

https://github.com/iiasa/CWatM - CWatM is a participating global
hydrological model in the ISIMIP intercomparison; see the repository for the
model documentation and publication list.

## Code and weights licenses

Code: GPL-3.0. No weights.

## Supported timestep

PT1D (daily)

## Required inputs

Forcing: precipitation, temperature, PET (its PET-driven configuration
matches the generated forcing; a Penman-Monteith configuration additionally
uses radiation, humidity and wind, but is optional).
Static: soil and land-cover maps; for a lumped probe case the adapter would
collapse these to catchment-mean statics.
Initialization: cold start with spinup.

## Outputs produced by the model

pr, evspsbl, mrro, dis, mrso, snw

## Evaluation window (optional)

Default.

## Packaging status

Help is needed with both Docker packaging and the adapter. (CWatM has no
PyPI package - checked 2026-09-11 - and no ready Dockerfile, but it is pure
Python, so both jobs are small.)

## Installation and inference notes

No PyPI package (both `cwatm` and `CWatM` return 404 on PyPI as of
2026-09-11). Install from the tagged source tree instead:

  pip install git+https://github.com/iiasa/CWatM.git@1.11

or clone at tag 1.11 and `pip install .`. Pure Python on the standard
scientific stack; upstream recommends a conda environment. A run is driven
by a single ini settings file, and example setups ship with the repository,
so the adapter is settings templating plus unit conversion. Inference needs
no network. CPU-only.

## Why this model belongs in the benchmark

CWatM is a global hydrological model in the ISIMIP intercomparison with
explicit irrigation-abstraction and groundwater modules, written in Python.
Two reasons. First, alongside LISFLOOD (issue #20) it is the other
production model best placed to be wired to the mass/human-abstraction case
(issue #19), and having two independent implementations of "the model was told
about the withdrawal" is what turns a probe from a single-model anecdote
into a comparison. Second, its groundwater compartment is a declared,
inspectable store - the variable most AI rainfall-runoff models get wrong
without anyone noticing - so its budget probes verdicts are informative
[...]

Issue #20, opened 2026-09-10
[MODEL] LISFLOOD (EC JRC)


kawh1111:
## Model name

LISFLOOD (EC JRC)

## Model category

physics

## Model code repository

https://github.com/ec-jrc/lisflood-code

## Version, release, or commit

v5.0.0 (latest tagged release as of 2026-09-11; the PyPI package
lisflood-model 5.0.0 matches)

## Weights repository (optional)

None required. An official Docker image exists (`jrce1/lisflood`, verified on
Docker Hub) and the repository README documents it for dependency-free runs.

## Paper or model reference (optional)

https://ec-jrc.github.io/lisflood/ - model documentation; LISFLOOD is the
hydrological engine of the European and Global Flood Awareness Systems
(EFAS/GLOFAS).

## Code and weights licenses

Code: EUPL-1.2. No weights.

## Supported timestep

PT1D (daily) - its operational configuration; hourly is also supported.

## Required inputs

Forcing: precipitation (mm/day), temperature (degC), PET (mm/day) - its
standard meteorological set, which the generated forcing covers directly.
Static: soil and land-use maps; for a lumped probe case the adapter would
collapse these to the catchment-mean statics the harness provides.
Initialization: cold start with spinup.

## Outputs produced by the model

pr, evspsbl, mrro, dis, mrso, snw

## Evaluation window (optional)

Default.

## Packaging status

Help is needed with both Docker packaging and the adapter. (The upstream
project ships an official image, so the container is a thin layer over it;
but no HydroTuring Dockerfile or /io adapter exists yet, so both are still
to be written.)

## Installation and inference notes

Per the repository README: `docker pull jrce1/lisflood:latest`, then
`docker run -v <case>:/input jrce1/lisflood /input/settings/cold.xml`.
Alternatively `pip install lisflood-model` (5.0.0 on PyPI; needs pcraster
and gdal from conda-forge). Test catchments ship in tests/data/. A run is
driven by one XML settings file; the adapter templates it and remaps paths.
Windows is explicitly unsupported upstream, which the container route
sidesteps. Inference needs no network. CPU-only.

## Why this model belongs in the benchmark

LISFLOOD is the engine behind the European Flood Awareness System, and it
ships explicit water-use modules - irrigation abstraction, domestic and
industrial withdrawal - as first-class budget terms. That makes it the
operational model best placed to be wired to the case mass/human-abstraction
(issue #19) prescribes, rather than one structurally blind to it, though the
[...]

PR #73, opened 2026-09-13
[PROBE] mass/spinup-cycle-invariance: periodic forcing should converge to a repeatable seasonal cycle

### LINKED ISSUE #63: [PROBE] mass/spinup-cycle-invariance: periodic forcing should converge to a repeatable seasonal cycle
### Authors

Kaihao Long, School of Geography and Planning, Sun Yat-sen University, Guangzhou 510006, China, [ORCID: 0000-0001-7578-3378](https://orcid.org/0000-0001-7578-3378)

### Conservation law

mass

### Residual equation

I would like to propose `mass/spinup-cycle-invariance`.

The probe asks whether a model driven by the same complete annual weather cycle eventually produces the same annual hydrologic response, or whether the answer still depends on how many copies of that cycle preceded the evaluation year.

For a physical catchment, a seasonal climate normally leads to a periodic seasonal orbit. Soil water, snow, groundwater, evapotranspiration, and runoff can vary strongly within a year, but after arbitrary initialization effects decay, the same day of the next year should have the same state and response. The proposed comparison is therefore an evaluation year after $N$ cycles versus the same evaluation year after $N+3$ cycles:

$$
Q_N(t) \approx Q_{N+3}(t),\qquad E_N(t) \approx E_{N+3}(t),\qquad S_N(t) \approx S_{N+3}(t).
$$

The probe does not require a seasonal hydrograph to become constant. It tests whether the seasonal orbit is repeatable. It targets an internal state that continues to evolve with elapsed cycles even though the physical forcing has returned to the same annual pattern.

The physical accounting law for each model run is

$$
\frac{dS}{dt}=P+G-E-Q,
$$

where $S$ is total reported physical water storage, $P$ is precipitation, $G$ is declared external exchange (positive into the catchment), $E$ is actual evapotranspiration, and $Q$ is total runoff leaving the modeled system.

$P$, $G$, $E$, and $Q$ are water depths per unit time, for example mm day$^{-1}$. $S$ is an equivalent water depth, for example mm. Soil water, snow water equivalent, canopy interception, groundwater, surface water, and routing/channel stores belong in $S$ whenever the model declares them as physical stores. Transfers among those stores are internal transfers and are not external sources or sinks.

The proposed forcing is periodic:

$$
X(t+365\ \mathrm{d})=X(t),
$$

where $X$ contains at least precipitation, temperature, and potential evapotranspiration. During either selected evaluation year, forcing is the same. Thus, for the paired difference between the two runs, $\Delta P(t)=0$, and the static catchment attributes and declared external drivers are also the same.

Integrating the conservation law over one evaluation year gives

$$
S(T)-S(0)=\int_0^T(P+G-E-Q)\,dt.
$$

At a periodic seasonal orbit, the storage on a given calendar day repeats in the next year, so $S(T)=S(0)$. The annual water balance then becomes

$$
\int_0^T(P+G)\,dt=\int_0^T(E+Q)\,dt.
$$

This conservation law is a precondition, not the entire new claim. A model can obey it exactly and still carry an internal clock, counter, or latent state whose value changes after every repeated year. The probe asks whether that extra state continues to change observable hydrologic behavior after the physical seasonal state should have repeated.

There are two related residuals, and keeping them separate is important.

First, each member of the pair must have an ordinary water-budget residual. For run $j$ over an evaluation year $[0,T]$,

$$
R_{\mathrm{closure},j}
=\sum_t(P_{j,t}+G_{j,t}-E_{j,t}-Q_{j,t})\Delta t
-[S_j(T)-S_j(0)].
$$

$R_{\mathrm{closure},j}$ has units of mm and should be approximately zero. This guard prevents a model from passing cycle comparison by simply losing or creating water.

Second, the proposed invariance residual compares the same day of the two phase-aligned evaluation years. For any reported flux or store $Y$,

$$
R_Y(t)=Y_{N+3}(t)-Y_N(t).
$$

For a flux, $R_Y(t)$ has units of mm day$^{-1}$; for a storage, it has units of mm. The score uses the largest absolute daywise departure:

$$
R_{\mathrm{cycle}}=\max_Y\max_{t\in[0,T]}|R_Y(t)|.
$$

Variables should include runoff, actual evapotranspiration, every required reported store, and total reported storage. Individual stores are compared as well as their sum: otherwise an increase in soil water could be hidden by a decrease in groundwater while the total appeared unchanged.

The probe compares outputs at the same position within the annual cycle. It does not compare unequal forcing or arbitrary days. A difference is evidence only when the corresponding day-by-day forcing is identical.

The suite uses a 5 percent engineering rule. For each variable $Y$, define

$$
r_Y =
\frac{
  \max_{t}\left|Y_{N+3}(t)-Y_N(t)\right|
}{
  \max\left(
    \frac{1}{n}\sum_{t=1}^{n}\left|Y_N(t)\right|,
    Y_{\mathrm{floor}}
  \right)
}.
$$

The proposed acceptance rule is

$$
\max_Y r_Y\le0.05.
$$

The denominator is the typical magnitude of the response after the shorter spin-up, because the question is whether three additional identical cycles materially change that response. It can approach zero: snow, canopy, or runoff may be absent for part of a synthetic case. An absolute floor is therefore required. A provisional floor is 0.05 mm $$day^{-1}$$ for fluxes and 1 mm for stores.

These floors prevent division by zero and prevent a near-zero baseline from receiving an accidental pass or an unstable infinite score. The 5 percent value, floors, choice of $N$, and the added three cycles should be calibrated against physical reference models. The case is suitable only when those references have reached a repeatable seasonal orbit; the probe must not impose an arbitrary universal spin-up length on every real catchment.

onsider a conservative bucket model with one additional internal variable $h$. The bucket accounts for every millimetre correctly, keeps all reported stores within their physical bounds, and closes the ordinary budget exactly. However, after each annual cycle it updates a hidden clock:

$$
h_{k+1}=h_k+\epsilon.
$$

This hidden variable does not create water and need not grow without bound; it can be a bounded phase or a counter reduced modulo a fixed range. It only needs to change the next year's runoff coefficient or evaporation partition to affect the output.

After $N$ cycles and after $N+3$ cycles, the model receives the same annual weather sequence, but $h$ has a different value. It therefore produces a different seasonal hydrograph or reported ET while $R_{\mathrm{closure}}$ remains near zero and all reported stores remain valid. This model fails $R_Y(t)$ even though ordinary `mass/catchment-closure` passes.

This is the central discrimination. The open [`mass/multi-decadal-drift` proposal](https://github.com/Flood-Lab/HydroTuring/issues/13) follows one long repeated-forcing run and catches storage that slowly leaves its physical bounds. A bounded hidden clock can avoid that drift. `mass/steady-state` uses constant forcing and asks for a constant terminal state; it does not establish that a seasonal orbit is invariant to extra annual spin-up. `mass/time-origin-invariance` changes the calendar origin, whereas this probe changes the amount of identical prior history.

The new probe is aimed at the case where all water budgets close, all explicit stores look plausible, and only the initialization-dependent seasonal response reveals the defect. A failure does not by itself prove that every long memory is unphysical: a real catchment may have a slow aquifer, snowpack, reservoir operation, land-use change, or another external driver omitted from the case. Such a driver must be declared or the case lengthened before treating the result as a violation.

The first case should be synthetic and deterministic so that the same experiment can be reproduced from a seed. The generator would:

1. draw one 365-day weather cycle containing precipitation, temperature, and potential evapotranspiration;
2. repeat its numerical values for nine consecutive cycles, so the forcing satisfies $X(t+365\ \mathrm{d})=X(t)$ exactly;
3. use the same model, initial state, and static catchment attributes in both variants;
4. designate the sixth cycle as the evaluation year after $N=5$ spin-up cycles and the ninth cycle as the evaluation year after $N+3=8$ spin-up cycles;
5. verify that the two selected evaluation years have identical precipitation, temperature, PET, timestep, and declared external forcing;
6. keep evaluation labels on the host side only and remove them before the adapter receives forcing, so a model cannot branch on `short` or `long`.

The initial version should be rain-dominated and above freezing. That makes the first conditioning question the convergence of soil-water, runoff, and evaporation memory rather than a simultaneous choice of a snow-memory timescale. A separate snow-specific case can be proposed later.

Before fixing the horizons and threshold, `reference_bucket`, `flex_lumped`, `flex_topo`, and `sacsma_snow17` should all agree between the sixth and ninth cycles. If they do not, the case is poorly conditioned and its forcing or spin-up length must be revised. Anti-degeneracy and state-bound checks are also needed so a constant-zero or incomplete output cannot pass by ignoring seasonal forcing.

Please review whether this periodic-attractor expectation, normalization, and conservative hidden-clock counterexample provide a useful and sufficiently distinct test of initialization 

### Will you build it?

Yes — assign it to me

longkaihao:
## Summary

This PR adds `mass/spinup-cycle-invariance`.

The probe tests whether a model reaches a repeatable seasonal state under periodic forcing. A physical model may need several cycles to approach its seasonal attractor, but once that state has converged, additional identical spin-up cycles should not change the following evaluation-year response.

The probe is designed to expose hidden initialization state or hidden clocks that continue changing with the number of prior cycles, even when the model reports a closed water budget and remains within all declared state bounds.

## Case construction

The generator creates one deterministic 365-day forcing cycle from the seed and repeats it for three variants:

- `short`: starts in 2001
- `plus3`: starts in 1998
- `long`: starts in 1997

All three variants score the same calendar year, 2007. Their visible evaluation-year forcing and static case metadata are therefore identical, while their prior histories differ.

The history offsets are 1096, 1461, and 365 days. These offsets have no common divisor greater than one, so a hidden clock with a short integer period cannot align with every pairwise comparison.

This construction avoids the leap-phase problem in the earlier shared-record design. The variants do not compare 2007, 2010, and 2011 directly. Instead, each variant has its own absolute start date and all three compare the same 2007 evaluation cycle.

## Criteria

The probe requires:

- precipitation, evaporation, and runoff fluxes;
- reported soil moisture, snow water equivalent, and canopy storage;
- optional groundwater and channel outputs when the model exposes them.

The preconditions check:

- closure over the complete post-spin-up record of every variant;
- state bounds over the complete post-spin-up record of every variant;
- forcing fidelity;
- non-degenerate hydrologic behavior.

The paired criterion compares evaporation, runoff, reported storage, and optional reported stores during the shared 2007 evaluation year.

The engineering tolerance is 5%. Fluxes use a 0.05 mm/day conditioning floor. Storage uses a 1.0 mm fallback floor with calibrated per-variable values:

- `mrso`: 1.0 mm
- `snw`: 1.0 mm
- `canopy`: 0.01 mm

The criterion validates `state_floors_mm` names and rejects unknown names instead of silently falling back to the scalar floor.

## Verdict interpretation

A model that reports all required outputs but produces different evaluation cycles receives `FAIL (VIOLATION)`.

This verdict records the observed inconsistency in the model's own outputs. The signature does not by itself identify whether the cause is a persistent hidden state, an internal clock, or a physical store that has not yet converged. A longer spin-up experiment can distinguish those cases for a particular model, but it does not change the observation made by this probe under its prescribed configuration.

`INCOMPLETE` is reserved for models that do not report enough information for the probe to ask its question.

## Reference controls

The gate requires the following physical references to pass:

- `reference_bucket`
- `flex_lumped`
- `flex_topo`
- `sacsma_snow17`

The negative controls are:

- `reference_restless`: hidden clock / non-convergent internal state;
- `reference_leaky`: closure failure;
- `reference_cheater`: state-bound failure;
- `reference_degenerate`: degenerate hydrologic response.

The leap-phase design was independently checked over seeds 0–199. The four must-pass physical references produced no failures, with SAC-SMA's worst departure at 0.056%. The hidden-clock reference was detected for all 200 seeds.

## Archive and model results

The archive was regenerated after rebasing on the current upstream main.

Docker-capable submitted adapters were rerun on r830. The current results include:

- CWatM: PASS
- SUMMA: PASS
- Wflow-SBM: PASS
- LISFLOOD: PASS
- δHBV2: FAIL because its reported `mrso` violates the declared `[0, 320]` state bound
- Google Flood Forecast: N/A because the adapter does not report the required probe variables
- MODFLOW6: N/A because it does not expose the paired hydrologic interface required by this probe

The model cards, README, roadmap, site tables, contributor records, and archive rows were updated for the current main branch and current probe count.

## Validation

- `ht validate`: 32 probes, 72 models
- `ht gate --probe mass/spinup-cycle-invariance`: passed
- Full test suite: 1165 passed, 8 skipped
- Regression coverage includes leap-phase alignment, coprime offsets, hidden clocks, long-only failures, all-variant aggregation, per-variable storage floors, invalid floor names, and report-detail preservation.


chrimerss (changes requested):
please address some minor changes request

chrimerss:
if you are ready, please request review. Thanks

longkaihao:
Thanks for the detailed review. I addressed the findings in commits `04c477a`, `3139fed`, and `7e49ab5`.

### Criterion and case correctness

- **Strict forcing fidelity:** `_same_forcing` now compares the complete visible forcing-column set in both evaluation windows and uses exact elementwise equality. The time column remains excluded deliberately because the two evaluation years have different absolute dates.
- **Total reported storage:** `spinup_cycle_invariance` now adds `total_reported_storage`, formed from every `STATE_VARS` column reported by both runs, and scores it with the state floor in addition to each individual store.
- **Leap-day alignment:** the generator now emits nine complete Gregorian years (3287 scored rows). February 29 reuses February 28's synthetic-cycle value and later leap-year dates are shifted back by one cycle index, so every repetition starts on January 1 with the same month/day forcing. `period_days` and the tests were updated accordingly.
- **Initial closure state:** the case now emits a real 365-day spin-up year. This gives `make_window` a state row immediately before the scored record and avoids the `spinup_days: 0` off-by-one.
- **Both selected cycles:** closure and state-bounds are explicitly configured as per-variant, phase-scoped preconditions. The harness evaluates them on the labelled evaluation cycle for both `short` and `long`, using the preceding row as the initial state.
- **Forcing echo:** the standard `forcing_fidelity` criterion now checks the model's reported `pr` against the supplied precipitation.

The paired form is retained for compatibility with the existing harness and now asserts that the two variants select different evaluation rows. The visible adapter inputs remain identical, while the host-side phase labels select different cycles.

### Documentation and archive

- Added `spinup_cycle_invariance` to `docs/writing-a-probe.md`.
- Updated the `reference_restless` descriptions in the README and authoring guide.
- Corrected the ROADMAP generalisation wording.
- Shortened the Spanish flowchart label to `ciclo inicial`.
- Documented the threshold calibration using the observed reference spread.
- Re-ran all ten archive entries for this probe on 2026-09-14 using the remote Docker environment.
- The archive now records four physical references passing, CWatM/LISFLOOD/SUMMA/Wflow passing, DHBV2 failing `state_bounds` as expected, and Google Flood Forecast as `N/A (INCOMPLETE)`.
- SUMMA now passes this probe after the evaluation-phase bounds check; its overall HydroTuring standing remains determined by its other probes.

### Validation

- `ht validate`: 24 probes, 41 models
- `ht gate --probe mass/spinup-cycle-invariance`: passed
- Targeted tests: 213 passed
- Full regression suite: 842 passed

chrimerss (changes requested):
Thanks for working through the earlier review. Most of it is fixed and checks out when run. One thing the leap-day fix introduced is blocking.

## Blocking

**The leap-day fix gives the forcing a 4-year period, and the scored pair sits at two different leap phases.**

`_cycle_day` repeats a value on 29 February, so 2004 and 2008 each get an extra day. The record now repeats every 1461 days, not every 365. The short evaluation year (2007) is three years after a leap day; the long one (2010) is two years after. Once a model has converged, its evaluation year matches the year four years later, not three years later. A threshold model therefore fails even though it converged.

SAC-SMA/Snow-17 is a `must_pass` model, and it shows this directly:

| seed | 2007 vs 2010 (this PR, K=3) | 2007 vs 2011 (K=4) | 365-periodic values, cycle 6 vs 9 |
| --- | --- | --- | --- |
| 134 | mrro **130.75%** | 0.015% | 0.016% |
| 177 | mrro **5.44%** | 0.005% | 0.006% |
| 8 | 2.95% | 0.00% | 0.00% |

In a 50-year run on seed 134, SAC-SMA converges by about 2006. After that, every year equals the year four years later to 1e-4, and years one leap phase apart differ by 13–131%.

Over seeds 0–199, SAC-SMA fails `spinup_cycle_invariance` on 2 seeds (134, 177). The gate seeds happen to miss them. With 1.5% failing seeds, about 4.5% of random 3-seed `ht run`s would FAIL a physical reference.

**Suggested fix:** `EXTRA_CYCLES = 4`, with `period_days: 3652`, `period_years: 10` and `min_window_days: 3652`. That scores 2007 against 2011. I checked this in a scratch copy over the same 200 seeds:
- SAC-SMA's worst departure is 0.06%, with 0 failures of this criterion.
- flex_lumped, flex_topo and the bucket are at 0.00%.
- `reference_restless` is still caught on 200/200 seeds (minimum 73%; 1461 mod 30 = 21 days of clock shift).

The tests (`2010`, `PERIOD_DAYS`), the README/ROADMAP wording ("five and eight cycles") and all ten archive rows would need to follow.

Related: the docstring and probe README say 29 February repeats 28 February's value. `_cycle_day` actually gives it index 59, which is 1 March's value.

## Should fix

1. **The `suspicious_exact` flag is lost on this probe.**
   - The `all_variants` branch in `evaluate_criteria` (harness.py:300ff) nests closure diagnostics under `variants`.
   - `run_probe` reads `result.diagnostics.get("suspicious_exact")` at the top level (harness.py:604), so the flag never fires.
   - Evidence: reference_bucket's report has `flags: []` here, although both per-variant diagnostics say `suspicious_exact: true`. The same model gets `suspicious_exact:mass/catchment-closure` on catchment-closure.
   - Over 200 seeds, full-record closure would flag every exact baseline; the official path flags none.
   - Fix: hoist `suspicious_exact = any(per-variant)` into the aggregated diagnostics, or have `run_probe` look inside `variants`.
2. **SAC-SMA fails `non_degenerate` on seed 69** (runoff ratio 0.0134 < 0.02). The generated climate is semi-arid for it: about 650–690 mm of rain against about 1000 mm of PET, giving runoff ratios of 1.3–5% across seeds. This is rarer (1/200) and independent of the K fix. Lower the floor or wet the cycle slightly.

## Nits

- **Runs are duplicated.**
  - `short` and `long` produce byte-identical outputs: closure residuals agree to every digit for summa and dhbv2. Every model therefore runs twice for one measurement. The author kept this on purpose; it is the maintainer's call.
  - `reference_cheater` escapes every criterion on 6/200 seeds (43, 103, 162, 164, 167, 186). The gate seeds are fixed, so the gate is stable.
- **Stale numbers.**
  - The probe README says "0.38% for LISFLOOD" and the probe.yaml comment says "Gate references span 0.00% to 0.38%", but the archive says 0.36%. LISFLOOD and CWatM are also not gate references.
  - The model cards still say 18: cwatm "14 of 18", wflow_sbm "15 of 18", lisflood "15 of 18", sacsma_snow17 "18 of 18". README and site say 19. Nothing enforces this.
  - The PR description is stale: 839 tests, summa FAIL.

## Verified

- **Branch state:** base main 123e941, 0 commits behind, mergeable.
- **Local checks:**
  - `ht validate`: 24 probes, 41 models.
  - pytest: 842 passed.
  - `ht gate --probe mass/spinup-cycle-invariance` and the full `ht gate` both pass.
- **Archive rows:** all ten reproduce byte for byte with `--gate-seeds` (4 subprocess; summa, dhbv2, cwatm, wflow_sbm, lisflood in Docker on arm64; google_flood_forecast N/A).
- **summa and dhbv2 without phase scoping:**
  - summa's PASS also holds with `state_bounds`/`closure` on the full 3652-row record, so phase scoping is not what flipped it.
  - dhbv2 fails full-record bounds the same way (mrso 595–861 against 320).
- **Earlier findings, all fixed:** strict forcing equality with a both-schema check; `total_reported_storage`; a real 365-day spin-up and prior-state row; per-variant closure/bounds; `forcing_fidelity`; a distinct-rows assertion; the docs catalogue row; the `reference_restless` entries; ROADMAP wording; the Spanish label; the calibration comment.
- **Harness change:** opt-in. No other probe uses `phase`/`all_variants`, and the full gate is unchanged.

longkaihao:
Thanks for the detailed review. I addressed each point in the follow-up commits:

- **Leap-phase alignment (blocking):** the generator now maps February 29 to the February 28 phase and uses a 3652-day, 10-year Gregorian record. The paired runs compare `N` and `N + 4` complete cycles, so the evaluation year has the same calendar/leap phase in both runs. The tests now assert the 2011 evaluation year and the four-cycle period.
- **`suspicious_exact` aggregation:** the all-variants aggregate now preserves this diagnostic whenever any variant reports it. I added a regression test covering the aggregate path.
- **SAC-SMA seed 69 non-degeneracy:** the previous 0.02 runoff-ratio floor rejected a physically valid low-runoff case (`r ≈ 0.0134`). I calibrated the floor to 0.01 and documented why; seed 69 now passes all criteria, including `non_degenerate`.
- **Archive and documentation:** I regenerated the spin-up rows on r830 from the repaired code. For this probe, the fresh results are PASS for reference_bucket, flex_lumped, flex_topo, SAC-SMA, CWatM, Wflow, LISFLOOD, and SUMMA (SUMMA's failures on other probes are unchanged); DHBV2 remains a real FAIL only on its reported `mrso` bounds (672–861 versus [0, 320]), while its spin-up invariance check passes; Google Flood Forecast remains N/A because the required variables are not reported. The model cards and suite documentation now use the current 19-probe count.
- **Validation:** the complete remote test suite passes: **843 passed**. The targeted probe/report/document tests pass: **214 passed**. Repository validation reports 24 probes and 41 models.

chrimerss (changes requested):
Thanks. The blocker and both should-fix items from the last review are fixed, and they hold up when run. What is left is small: a few stale numbers, one mechanism question, and CI.

## Fixed and verified

- **Leap phase.**
  - `EXTRA_CYCLES = 4` scores 2007 against 2011, and 29 February now takes 28 February's value.
  - Over seeds 0–199, none of the four must-pass models fails any criterion. SAC-SMA's worst departure is 0.06%; seed 134 was 130.75% before.
  - `reference_restless` is caught on 200/200 (minimum 73%).
- **`suspicious_exact`.** It now fires on every exact baseline, on 200/200 seeds, and the new regression test covers it.
- **`non_degenerate` floor 0.01.**
  - SAC-SMA's lowest runoff ratio over the 200 seeds is seed 69 at 0.0133, which now passes.
  - `reference_degenerate` is still caught on 200/200.
- **Checks.**
  - `ht validate`: 24 probes, 41 models.
  - pytest: 843 passed.
  - The probe gate and the full `ht gate` both pass.
  - All ten archive rows reproduce byte for byte with `--gate-seeds`: 4 subprocess; summa, dhbv2, cwatm, wflow_sbm and lisflood in Docker on arm64; google_flood_forecast N/A.
- **Model cards.** cwatm, wflow_sbm, lisflood and sacsma_snow17 are updated.

## Still open

1. **`phase: evaluation` leaves 8 of the 10 scored years unchecked by `closure` and `state_bounds`.**
   - Both variants give the adapter identical inputs, so `all_variants` alone would score the whole record twice at no extra cost.
   - `phase` keeps only 2007 and 2011, so a leak or an out-of-bounds store in any other year passes this probe.
   - Deleting the two `phase: evaluation` lines still passes the gate. Over the same 200 seeds, full-record closure and bounds fail no must-pass model and catch `reference_leaky` 200/200 and `reference_cheater` 191/200.
   - I rescored the five Docker models' kept outputs on the full record: no verdict moves. Only dhbv2's message changes, from 365 to 3652 steps.
   - If per-year scoping is intended (#63 defines closure over an evaluation year), please say in the probe README that the other years go unchecked.
2. **The `all_variants` aggregate can report the passing variant** (harness.py, the `all_variants` branch).
   - The worst variant is picked by `value`, with `None` ranked lowest, and `value=max(values)` skips `None`.
   - A failing variant with no value, next to a passing one, reports the passing variant's value, threshold and name.
   - Fix: pick it the way `run_probe` picks the worst seed, `min(..., key=(passed, -abs(value or 0)))`.
   - A test with a long-only failure would cover this path. docs/writing-a-probe.md should also document `all_variants` and `phase`.
3. **Stale figures.**
   - The probe README says "0.02% for CWatM, and 0.36% for LISFLOOD", and probe.yaml says "0.00% to 0.36%". The archive now says CWatM 0.00% and LISFLOOD 0.47%. Neither model is a gate reference.
   - `models/summa/README.md:507` still says "8/21 probes passed". With this probe it is 9/22.
   - `models/lisflood/README.md`: on a fast host the model would pass 17 of 19, not "16 of 19". It was 16 of 18 before this probe added a pass.

## Nits

- A hidden clock whose period divides the 1461-day offset goes undetected. On the gate seeds, 3 d and 487 d give 0.00%, 365 d gives about 0.6%, and 73 d is caught on one seed only. The 30-day reference is caught at 95–113%. This is worth a line under "What it does not claim".
- `reference_cheater` escapes every criterion on 4/200 seeds (43, 103, 164, 186). The gate seeds are fixed, so the gate is stable.
- There is an f-string with no placeholders at `criteria/base.py:133`.
- The README's reason for `total_reported_storage` doesn't hold, because the sum is scaled by the mean total, which `mrso` dominates.

## Process

- The fork CI has not run on any of the six pushes; every run is `action_required`.
- The maintainer's CHANGES_REQUESTED is still in place.

cehw:
Thanks for noting the long-transient limitation. I ran a small check with the native CWatM 1.11 single-cell adapter using this probe’s three gate seeds.

The default settings passed all three cases. I then changed only the groundwater recession coefficient from 0.0069 to 0.001 day⁻¹, slowing groundwater drainage while keeping the forcing and criteria unchanged. This value is within CWatM’s published parameter-map range.

Groundwater storage differences between the two evaluation years reached 10.0–10.3%, above the 5% threshold; all other criteria passed. Keeping the same parameters and adding 40 years of preceding repeated weather restored a pass in all three cases. This was a recovery check, not a claim that 40 years is necessary.

Could we also carry the README’s caveat into the reported result? For example: “Outputs still differ after the prescribed spin-up; insufficient spin-up is one possible cause, so this result alone does not establish a physical violation.”

chrimerss:
@cehw thanks, this reproduces exactly. I copied CWatM with only `recessionCoeff` changed from 0.0069 to 0.001 per day and ran it at 0e5e30e on the gate seeds. `gw` differs by 10.31%, 10.14% and 10.00%, and every other criterion passes. 0.001 is inside the 10th–90th percentile band of CWatM's own 30′ maps (0.0002–0.061); at 0.0002 the difference is about 26%. With 40 extra years of the same weather in front of the record, 0.001 passes at 0.00%.

We'll keep the verdict as FAIL. The probe asks whether a model reaches the same year after the spin-up it is given, and this one hasn't. The reported result should still say what the README already says, so nobody reads the FAIL as proof of a hidden state.

@longkaihao could you:

1. Add the caveat to the `spinup_cycle_invariance` failure message (`criteria/spinup.py`, the "differs by" line), so a failure reads:
   > 'gw' differs by 10.31% between the N- and N+K-cycle evaluations (limit 5.00%). Outputs still differ after the prescribed spin-up; insufficient spin-up is one possible cause, so this result alone does not establish a physical violation.
2. Assert that sentence in a test. The `reference_restless` test already produces a failure to check.
3. Under "What it does not claim" in the probe README, note that a FAIL can come from a slow physical store that hasn't settled yet, with this CWatM case as the example.

None of the archived rows fails this criterion, so the message change needs no re-runs.

The branch now conflicts with main after #74 and #70 merged: `README.md`, `models/result.csv`, `site/index.html`, and the LISFLOOD and SAC-SMA model READMEs. After rebasing, the standings become:
- 21 of 21 for the physical models
- dhbv2 11/20
- wflow_sbm 17/20
- summa 9/23
- cwatm 16/20
- lisflood 17/20

The open items from my last review still apply.

longkaihao:
Thanks for the detailed reproduction. I have rebased the branch onto the latest `main` (`e5c87e2`) and resolved all merge conflicts. The rebased branch is currently at `26107e3`.

For the slow-groundwater case:

The `spinup_cycle_invariance` verdict remains FAIL when the model has not reached the same evaluation-year state after the prescribed spin-up.

The failure message now includes:

> “Outputs still differ after the prescribed spin-up; insufficient spin-up is one possible cause, so this result alone does not establish a physical violation.”

I added a regression assertion for this exact sentence.

The probe README now documents the CWatM slow-groundwater example and explains that a FAIL may indicate insufficient spin-up for a slow but physically valid store, rather than proving a hidden state or physical violation.

The earlier review fixes from `55de36d` were also carried forward:

- Removed `phase: evaluation` from the closure and `state_bounds` preconditions, so both now cover the complete post-spinup record.
- Kept the evaluation phase only for the paired comparison of the two selected spin-up cycles.
- Ensured that the row immediately before the post-spinup record is used as the initial state.
- Fixed `all_variants` aggregation so a failure in the long variant cannot be hidden by a passing short variant, even when the failing result has no numeric value.
- Added a regression test for this long-only failure case.
- Added documentation describing the intended use of `all_variants` and `phase`.
- Updated the spin-up probe README with the calibrated reference-model values, the limitations of the discriminator, and the explanation of `total_reported_storage`.
- Updated the stale probe-count summaries in the CWatM, Wflow-SBM, and SUMMA model cards.
- Removed unnecessary `f` string prefixes where no interpolation was used.

Validation completed:

- Spin-up probe tests: 15 passed.
- Documentation synchronization tests: 166 passed.
- Full test suite: 835 passed, 8 skipped. The one local failure is the existing `window_days: 0` validation test when using the temporary local `jsonschema` shim; it is unrelated to these changes.

The CWatM verdict remains intentionally FAIL for the shorter prescribed spin-up. The new wording distinguishes “not converged under this test configuration” from “proven physically invalid.”

Barbhuiya12 (commented):
I ran the gate, reproduced the discrimination, and went to the spin-up literature on the one question the probe's own README flags as open. Everything below is verified on `pr73` locally unless I say otherwise.

The probe fills a real gap. `mass/steady-state` asks whether constant forcing produces constant output; nothing in the suite asked whether *periodic* forcing produces a *repeatable* output, which is the property an initialised physical model actually has and a hidden state need not. The framing as a state-space test rather than a claim about spin-up length is the right one, and the README is unusually honest about its own limits — the `reference_cheater` seed disclosure especially. What follows is one substantive suggestion and some smaller things.

## What I verified

```
gate, all four must_pass and all four must_fail as declared   PASS

worst departure over the three gate seeds
  reference_bucket     0.00%          reference_restless   113.11%
  flex_lumped          0.00%          threshold              5%
  flex_topo            0.00%
  sacsma_snow17        0.00%
```

The separation is four orders of magnitude, not a tuned margin. `reference_restless` fails specifically on `spinup_cycle_invariance`, and `reference_cheater` trips `state_bounds` as pinned.

## The main point: the alias is disclosed, but it is cheap to close entirely

The README says a hidden clock "whose period divides 1461 days (four Gregorian years) can return to the same phase in both selected years", and suggests extending or phase-shifting the case "if a model exposes evidence of that kind of calendar lock".

Two things make me think that deserves more than a caveat.

**First, the blind spot is not exotic.** Injecting a hidden state into `reference_bucket` that modulates `mrro` by 10% with period *p* cycles, on gate seed 107008012:

```
   p   2007%p  2011%p   verdict   departure
   2        1       1   PASS       0.00%
   3        0       1   FAIL       7.82%
   4        3       3   PASS       0.00%
   6        3       1   FAIL       7.45%
   8        7       3   FAIL      10.79%
```

Periods 2 and 4 are not near the threshold — they are **exactly 0.00%**, indistinguishable from a perfectly converged model. And an alternating state or a short counter is the simplest hidden state a recurrent model can carry, not a pathological one.

**Second, the suggested mitigation is circular.** "If a model exposes evidence of that kind of calendar lock" — this probe is the thing that would expose it. Nothing else in the suite looks at repeated identical cycles, so a period-2 lock has no route to being noticed at all.

**The fix looks cheap.** A hidden clock of period *p* survives a comparison at offset *k* exactly when *p* divides *k*. So it survives a *set* of offsets only when *p* divides their gcd:

```
  offsets     gcd    periods that survive
  (4,)         4     2, 4
  (3,)         3     3
  (3, 4)       1     none — only a genuinely converged state escapes
```

Adding one variant at **+3 cycles** alongside the existing +4 leaves gcd = 1, and no periodic hidden clock can sit at the same phase in all three evaluations. I checked the calendar constraint holds: 2007 and 2010 are both non-leap, so the +3 pair satisfies the same leap-phase requirement the +4 pair was chosen for, and `_same_forcing` verifies the evaluation forcing is identical at runtime anyway.

The cost is one extra run per seed, +50% runtime. Given `max_runtime_s: 300` and a 4017-row record, that is a real cost for the Docker adapters and it is the maintainer's call whether it is worth paying. But it converts "this probe can be evaded by a period-2 clock" into "nothing periodic evades it", which seems like a large return for one variant.

## The spin-up length: the literature supports the README's caveat more strongly than the README does

The README says the four references passing is "evidence that five cycles are enough for this generated catchment, not a claim that five years is a universal spin-up length". That is exactly right, and I think it is worth citing, because the numbers are not marginal:

- [Seck, Welty & Maxwell (2015, WRR)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2014wr016371) report **2, 9, 6 and 5 years** to reach 1% convergence in groundwater storage, unsaturated storage, CLM water storage and root zone respectively, for ParFlow.CLM on the Monocacy basin.
- [Cosgrove et al. (2003, JGR)](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2002JD003316) found Mosaic reached fine-scale total and root-zone soil moisture equilibrium in about **5.5 years**, while **VIC and Noah in some cases took longer than 10 years**.
- [Ajami et al. (2014, WRR)](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1002/2013WR014258) use a criterion of groundwater storage change below 1% of potential recharge, and find subsurface storage to be the binding variable for system-wide equilibrium.

The probe compares cycle 5 with cycle 9. That interval sits squarely inside the window in which several widely used models are still converging, and the binding store in every one of those studies is the one this probe declares *optional* (`gw`). The four `must_pass` references reach 0.00% because they are fast conceptual stores without a deep groundwater reservoir — which is a property of the reference set rather than evidence the interval is safe for the submitted models the probe exists to judge.

**This creates a concrete tension I would like settled before merge.** The criterion's own failure message says the result "does not establish a physical violation", and the README repeats it. But a criterion FAIL is archived as:

```
dhbv2,FAIL,VIOLATION
```

and counts against the model's standing. A criterion that disclaims establishing a violation should probably not produce one. Two options that seem consistent with the suite: raise the spin-up so that a FAIL is defensible as physics, or have the criterion report `N/A (INCOMPLETE)` when it cannot distinguish a slow store from a hidden clock. I do not think the current combination — disclaiming in prose while archiving `VIOLATION` — should ship, and I would say the same about my own probe.

## Smaller

**The `phase` window is added to shared code and never used.** `make_window(run, probe, phase=None)` gains ~20 lines, and `closure` and `state_bounds` both pass `params.get("phase")`. No probe in the tree sets `phase:`, including this one, and I found no test exercising it (`grep -c "phase=" tests/test_spinup_cycle_invariance.py tests/test_harness.py` → 0, 0). The comment says "A periodic probe may ask for closure on each selected evaluation cycle", which this probe does not do — it uses `all_variants: true` and scores the full record. Either this probe should use it, or it should come out until something does. Shared code with no caller and no test is the kind of thing that drifts.

**The README contradicts itself on CWatM.** Line ~100: "CWatM with a slower groundwater recession differs by more than 5% after the probe's spin-up." Acceptance gate section: "departures are 0.00% for the exact and FLEX references, 0.00% for CWatM". The first is presumably a hypothetical reconfiguration and the second the archived run, but as written they read as the same model with two answers. Worth disambiguating, because that paragraph is doing load-bearing work in explaining what a FAIL means.

**The threshold justification mixes populations.** `probe.yaml` says "Gate references span 0.00% to 0.47%". I measure all four gate references at exactly 0.00%; the 0.47% is LISFLOOD, a submitted model. The margin is enormous either way, so this is accuracy rather than substance.

**`reference_cheater` escaping all criteria on 4 of 200 seeds** is disclosed, and I appreciate that it is. With `n_seeds: 3` the gate is fine on the derived seeds today. I would only ask that the disclosure say what happens if it is not — the gate is deterministic, so this is stable unless the seed derivation or the probe id changes, and it might be worth a line saying so.

**`bool(params.pop("all_variants", False))`** reads a quoted `"false"` as True. YAML booleans parse natively so this cannot bite from `probe.yaml` today; I only mention it because I hit exactly this in `_as_bool` on #67 last week.

## Recommendation

The physics is sound and the implementation is careful. My one substantive ask is the third variant, or an explicit decision from @chrimerss that the period-2/4 alias is acceptable and the README should say plainly that it is unmitigated rather than "extend the case if a model exposes evidence". The `VIOLATION`-versus-disclaimer tension I think does need resolving either way.

Happy to re-run the injection test against a third-variant version if you add one.

Barbhuiya12 (commented):
Re-ran everything against `2a7b877`. The coprime offsets close the alias, confirmed rather than assumed:

```
equal-amplitude (±5%) hidden clock injected into reference_bucket, all three gate seeds

   p   seed1  seed2  seed3   verdict
   2   24.7%  22.1%  22.2%   FAIL      (was 0.00% PASS on the +4-only design)
   3   16.8%  15.0%  15.1%   FAIL
   4   12.4%  11.1%  11.1%   FAIL      (was 0.00% PASS)
   6   24.7%  22.1%  22.2%   FAIL
```

Nothing marginal — the weakest is 11% against a 5% threshold. `gcd(3, 4) = 1` does what it should. Gate still passes on all eight baselines with the third variant, and the threshold comment now separates the gate references from the LISFLOOD observation.

One correction to my own review while I'm here: I initially measured period 4 at 5.59%, close enough to the line to look like a weak catch. That was my test, not your probe — my modulation was `(y mod p)/p`, which is half the amplitude at p=4 as at p=2. With equal amplitude it is 11–12%. Nothing to look at.

The 40-year CWatM diagnostic is a better answer than the one I asked for. I raised the spin-up interval from the literature — [Seck et al. 2015](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2014wr016371) needing 9 years for groundwater storage, [Cosgrove et al. 2003](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2002JD003316) over 10 for VIC and Noah — and you answered it by running the actual packaged model out to 40 years and showing it sits at 0.02%. That settles it for this adapter in a way a citation cannot, and the 100-year timeout is worth recording exactly as you have.

Two things still open, both small.

**The `VIOLATION` tension moved rather than resolved.** The disclaimer is gone from the criterion message, which is an improvement, but the README still says a FAIL "by itself does not establish a physical violation or a hidden state" and a criterion FAIL still archives as `FAIL,VIOLATION`. With the 40-year evidence in hand, I think the README sentence is now the thing that is out of date rather than the verdict: you have shown the prescribed spin-up is sufficient for the one model where it was in doubt. Saying that plainly — that a FAIL is a real result unless a model can show convergence at longer spin-up, as CWatM did — would be a stronger claim than the current hedge and is now supported.

**Line 102 reads as contradicting the archive.** "CWatM with a slower groundwater recession differs by more than 5% after the probe's spin-up" against "CWatM completed all three variants and passed with a worst departure of 0.02%" and `cwatm,PASS` in `result.csv`. I take the first to mean a hypothetical reconfiguration rather than the packaged adapter, but as written the same model has two answers in one document. Worth a clause.

**On the `phase` window**, the test added covers it and that is enough for me — I withdraw the "no test" half. It is still unused by every probe in the tree, so it remains shared code carried for a caller that does not exist. Your call and @chrimerss's whether that is worth keeping now or dropping until something needs it; I do not think it blocks anything.

Nothing else from me. The probe does what it says and the alias fix is clean.

longkaihao:
Thanks @Barbhuiya12 for the detailed physics review. I have addressed the substantive hidden-clock concern and the related interpretation issues in the current branch.

### Hidden-clock alias

The original `N` versus `N+4` comparison could miss hidden clocks whose period divides four. I added a third selected cycle at `N+3`, so the probe now compares:

- `short`: `N`
- `plus3`: `N+3`
- `long`: `N+4`

The criterion performs all three pairwise comparisons. Because `gcd(3, 4) = 1`, a short integer-period hidden clock cannot remain phase-aligned in every comparison.

I also added deterministic regression coverage for hidden-clock periods 2, 3, 4, 6, and 8. The generator, probe README, and `probe.yaml` now document this design.

This change is included in commit `0221ab7`.

### Slow physical stores and the meaning of FAIL

The slow-groundwater case raised an important interpretation issue. A model can be physically conservative but still fail the prescribed spin-up because its groundwater store has not reached the periodic attractor.

The criterion therefore still returns `FAIL` when the selected evaluation years differ, but the failure message now explicitly states:

> Outputs still differ after the prescribed spin-up; insufficient spin-up is one possible cause, so this result alone does not establish a physical violation.

I added a regression assertion for this exact message and documented the limitation in the probe README.

### Long-spin-up experiment

I ran a separate 40-year repeated-history diagnostic for CWatM. The default packaged CWatM configuration passed, with a worst pairwise departure of approximately `0.02%` in groundwater storage. The previously reproduced slow-groundwater configuration also converged after 40 years, whereas it differed by about 10% under the shorter prescribed spin-up.

This suggests that a sufficiently long spin-up can provide stronger evidence about the interpretation of a failure: if a model still fails after a carefully chosen long spin-up, the result is more likely to indicate persistent hidden-state dependence, an unreported store, or another physically problematic mechanism. However, the appropriate threshold should be determined more cautiously rather than assumed to be exactly 40 years, for example by calibrating it against experiments with physical reference models and considering the class of catchments represented by the benchmark.

The formal probe remains at the shorter `N`, `N+3`, and `N+4` configuration for now. A long-spin-up version would also face substantial execution limits:

- the current probe sets `max_runtime_s: 300` seconds per model invocation;
- the probe schema currently permits at most 600 seconds;
- the GitHub gate job has a 15-minute workflow limit;
- the Docker container check has a 20-minute job limit;
- the on-demand model-evaluation workflow has a 45-minute job limit.

A model with expensive initialization or long per-step runtime may therefore time out before completing a 40-year case. Such a timeout would be an execution error, rather than evidence that the model violates conservation.

@chrimerss, could you please advise whether the current `FAIL` semantics should be retained for models that do not converge under the prescribed spin-up, together with the explicit caveat above, or whether the project prefers a different existing classification for this situation?

### Other review fixes carried forward

The branch also retains the earlier fixes for:

- leap-phase alignment;
- a real 365-day initial spin-up row;
- full-record closure and state-bound checks;
- strict forcing fidelity;
- total reported storage;
- correct `all_variants` aggregation when a failing variant has no numeric value;
- documentation of `phase` and `all_variants`;
- the criterion catalogue and `reference_restless` documentation.

The local validation and acceptance gate pass with the three-variant implementation.

Barbhuiya12:
@longkaihao you asked @chrimerss whether `FAIL` is the right classification when a model does not converge under the prescribed spin-up. I went looking for whether an existing reason could absorb it, and I think the answer is in `scoring.py`'s own docstring:

> VIOLATION says the model reported its budget and the budget did not close.
> INCOMPLETE says the model never reported enough to be checked at all.
> ...
> Neither INCOMPLETE nor INCOMPATIBLE is a failure. A probe stopped by either never asked the model anything.

That rules `INCOMPLETE` out cleanly. A model here reports everything the probe requires; the probe asked its question and got an answer. Reclassifying it would make `INCOMPLETE` mean two different things — "did not report enough" and "reported enough but I cannot attribute the cause" — and the whole point of that docstring is that the reasons stay distinct. Same for `INCOMPATIBLE`, which is about consuming the case.

Read the other way, though, I think the same sentence argues `VIOLATION` is already right. The analogue of "reported its budget and the budget did not close" is "reported two identical years and they were not identical". That is a disagreement between the model's own outputs under forcing the criterion verifies is byte-identical. The *cause* may be a slow store rather than a hidden clock, but the *observation* is a real inconsistency in what the model reported, which is exactly what VIOLATION is defined to record.

If that holds, then the thing to change is not the verdict but the README sentence that says a FAIL "by itself does not establish a physical violation or a hidden state" — and your own 40-year run is the reason it can change. The one model where slow convergence was actually suspected converged at 0.02% well inside the prescribed window once you tested it rather than assumed it. The remaining hedge is now doing less work than the evidence you gathered.

A wording that would survive both readings, for whatever it is worth:

> A FAIL means the model did not reproduce the same evaluation cycle after different amounts of identical prior history. It does not identify the mechanism: a hidden state and a store that has not settled produce the same signature here. A model can distinguish the two itself by demonstrating convergence at longer spin-up, as CWatM does at 40 years.

That keeps the honesty about mechanism while dropping the part that reads as disowning the verdict the probe just issued.

On spin-up length generally, I would not put a number on it from the literature. [Seck et al. 2015](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2014wr016371) needed 9 years for groundwater storage in ParFlow.CLM and [Cosgrove et al. 2003](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2002JD003316) over 10 for VIC and Noah, but both are real catchments with real subsurface depth, and this is a generated lumped case whose stores you have characterised directly. Your empirical calibration against the four references is better evidence for *this* case than either paper, and your caution about not fixing 40 years as a universal threshold is the right instinct. The runtime ceilings you listed seem to settle it in any case.

Nothing further from me on the technical side — the coprime offsets close the alias and I have verified that independently.

longkaihao:
Thanks for the clarification. I kept the result as `FAIL (VIOLATION)` and did not change the framework-level classification.

The model reports all required variables, the criterion is evaluated successfully, and the model fails to reproduce the same evaluation cycle after different amounts of identical prior history. Therefore this is a scored criterion failure, not `INCOMPLETE`.

I revised the spin-up criterion message so that it states the observed failure directly while keeping the mechanism distinction explicit:

> The model did not reproduce the same evaluation cycle after different amounts of identical prior history. This signature does not identify the mechanism: it may reflect a hidden state or a physical store that has not yet settled.

The probe README now explains the same distinction. The 40-year CWatM run is documented as a diagnostic showing convergence for that specific slow-groundwater configuration; it is not treated as a universal 40-year requirement.

The coprime `N+3` and `N+4` variants remain unchanged. They close the short-period alias identified in the review. The existing regression tests cover the period-2, 3, 4, 6, and 8 hidden-clock cases.

Validation after this update:

- Spin-up and documentation tests: 225 passed
- Full test suite: 1043 passed, 8 skipped
- No framework-level scoring or verdict logic was changed
- No archived model verdicts or numeric results changed

Barbhuiya12 (dismissed):
Re-ran everything on the final head.

```
ht validate   29 probes, 60 models
ht gate       PASSED, all eight baselines
pytest -q     1051 passed

alias regression, +/-5% hidden clock injected into reference_bucket
  p=2  24.7%   p=3  16.8%   p=4  12.4%   p=6  24.7%   p=8  16.0%    all FAIL
```

The coprime offsets still close the alias after four more commits, which is the part I was most concerned would drift.

Keeping `FAIL (VIOLATION)` is the right call, and I think it matters past this probe. Any probe whose failure has more than one possible mechanism runs into the same question, and the pattern here is a good one to have set: report the observation, say plainly that it does not identify the mechanism, and do not disown the verdict the probe just issued. The new message does exactly that.

On @chrimerss's three items, I checked that `phase:` is gone from the probe, that the `all_variants` worst-variant selection is now keyed on `(passed, -abs(value))`, and that the four stale figures are no longer in the tree. That was grep plus a gate run rather than a line-by-line read of the harness change, so treat it as a physics approval, not an implementation one.

Two limits I would not want an approval to paper over, both of which your README already states: a clock with a period longer than the tested offsets still escapes a finite test, and the sufficient spin-up length stays case-specific rather than being settled at 40 years. The probe claims neither, and that is the right scope for it.

Approving. Thanks for the 40-year run in particular — I raised the spin-up interval from the literature and you answered it by measuring the actual packaged model, which is the better answer.

chrimerss (changes requested):
Thanks for the rebase and for working through everything, and thanks @Barbhuiya12 for the hidden-clock review. Most of it holds up when run. The new `plus3` variant, though, brings back the leap-year problem from the first review.

## Blocking: `plus3` scores 2010, which sits at a different leap phase

The three evaluation years are 2007, 2010 and 2011. February 29 repeats a day, so the forcing repeats every four calendar years, not every year. A converged model's 2010 therefore legitimately differs from its 2007 and its 2011, and a threshold model shows it. SAC-SMA/Snow-17 is a `must_pass` model, and it fails on two of seeds 0–199:

| seed | worst pair | departure |
| --- | --- | --- |
| 171 | `short<->plus3:mrro` (2007 vs 2010) | 60.02% |
| 177 | `plus3<->long:mrro` (2010 vs 2011) | 5.49% |

- `ht run --model sacsma_snow17 --probe mass/spinup-cycle-invariance --seed 171` reproduces the first row.
- With 1% of seeds failing, about 3% of random 3-seed runs would FAIL a physical reference. The gate seeds happen to miss both.
- The generator docstring says the leap-day mapping is now "tested rather than hidden". It is the other way round: pairs at different leap phases are not comparable.

Within one shared record, leap alignment needs offsets that are multiples of four years. Coprime offsets and leap alignment therefore can't both hold there. Two ways out:

1. **Drop `plus3`** and compare N against N+4, as at 0e5e30e. The README then states that a hidden clock with a period of 2 or 4 cycles escapes.
2. **Keep the 3- and 4-cycle offsets, but give each variant its own start year and have all three score the same calendar year.**
   - For example, all three score 2007, with records starting in 2001, 1998 and 1997, so they carry 6, 9 and 10 prior years.
   - I ran SAC-SMA in-process over the same 200 seeds: 0 failures, worst 0.056%.
   - The variants would no longer share visible dates or row counts. `test_adapters_receive_the_same_visible_case_metadata` would change, and it needs checking that the harness accepts per-variant dates and lengths, including the time-based window it applies to LISFLOOD.

## Fixed and verified

- **Rebase:** the branch is on 57d1d41 and mergeable.
- **Checks:**
  - `ht validate`: 29 probes, 60 models.
  - pytest: 1051 passed.
  - `ht gate --probe mass/spinup-cycle-invariance` and the full `ht gate` both pass.
- **Archive rows:** all ten reproduce byte for byte with `--gate-seeds`: four subprocess; summa, dhbv2, cwatm, wflow_sbm and lisflood in Docker on arm64; google_flood_forecast N/A.
- **Preconditions:** `closure` and `state_bounds` now cover the full post-spin-up record in every variant. Over 200 seeds they catch `reference_leaky` 200/200 and `reference_cheater` 191/200, and pass every must-pass model.
- **`all_variants` aggregation:** a failing variant wins even without a value, and the new test covers it. `suspicious_exact` still fires on every exact baseline, and quoted booleans parse correctly.
- **Failure message:** it states the observation and says it doesn't identify the mechanism, with a test.
  - In an `ht run --csv` row, `reference_restless`'s detail keeps the whole sentence.
  - So does a slow-recession CWatM copy (`recessionCoeff` 0.001): it fails at 10.31% (`short<->long:gw`), and its `--csv` row carries the whole caveat.
- **Hidden clocks:** `reference_restless` is caught on 200/200 seeds (minimum 91%).

## Nits

- **Stale comments and text:**
  - `generate.py`'s docstring has a broken sentence ("The scored comparison is / The selected years are…").
  - probe.yaml says closure covers "both selected cycles", and provenance says the cycle is "repeated nine times".
  - The comments in `bounds.py`/`closure.py` still describe the phase window, which no probe uses now. `make_window`'s phase branch has no caller.
  - The probe README calls `mass/multi-decadal-drift` an open proposal; it merged as #14.
- **The long-spin-up section:** it reports the default CWatM at 40 years (0.02%). The case that raised the question is the slow-recession one: about 10% at the prescribed spin-up, passing after 40 years. Put that result there.
- **Cost:** each seed now runs the model three times on identical inputs.
- **PR description:** it still describes two runs (N and N+4) and 843 tests. It also still says "comparing runs three years apart can place them on different leap-day phases", which is the problem `plus3` brings back.

longkaihao:
@chrimerss  Thanks for identifying the leap-phase problem. I fixed it in commit `e841470`.

The three variants now start in 2001, 1998, and 1997, respectively, while all score the same non-leap calendar year, 2007. Their visible dates differ intentionally, but their evaluation-year `pr`, `tas`, and `pet` forcing values and static metadata are identical. This preserves the N+3 and N+4 history offsets without comparing different leap phases.

The existing probe contract requires a fixed record length, so I kept the 3652-day record length and avoided framework changes. The regression test now verifies the different start dates, common 2007 evaluation year, identical scored forcing, and hidden phase labels being absent from adapter inputs.

Validation completed: 225 targeted tests passed, 1043 full tests passed with 8 skipped, `ht validate` passed for 29 probes and 60 models, and the spin-up gate passed all declared baselines.

kawh1111 (commented):
Thanks @longkaihao. I ran the probe at `e841470` locally before reading the README, building the case for all three variants on seed `20260912` and running `reference_bucket` and `reference_restless` through the subprocess runner. Most of it holds up, so I will start there.

- **The gate separates, and only the intended criterion fires.** `reference_bucket` reports `0.0` on `spinup_cycle_invariance`; `reference_restless` reports `1.36` and fails on that criterion alone, with `closure` at 4e-17, `state_bounds` at 0.0 and `non_degenerate` at 0.29. A hidden clock that conserves water exactly and stays inside every bound is caught here and nowhere else, which is the claim the probe makes.
- **The phase labels do not reach the model.** The staged forcing CSV the adapter reads is exactly `time, pr, tas, pet`, and the three evaluation-year forcings compare equal column by column.
- **The coprime argument is right.** The scored windows begin at rows 2191, 3287 and 3652 of the 4017-row record. The differences between them are 1096, 365 and 1461, and no integer greater than one divides all three, so no short integer clock can be aligned in every comparison.
- **The `reference_cheater` note is a stable property, not a lucky draw.** This probe's gate seeds are `107008012`, `613960126` and `1120912240`, none of which is among the four you list, so the gate's result on `state_bounds` does not depend on the seed draw.

Two things I measured that I could not find in the probe README.

## 1. The 1 mm storage floor, not the 5 percent rule, is what governs the snow and canopy channels

The criterion takes `scale = max(mean_t |Y|, floor)` with `state_floor_mm: 1.0`. On the scored year of this case that single constant does very different work on each reported store:

| variable | floor | mean \|Y\| on the scored year | scale | departure that fails |
| `mrso` | 1.0 mm | 118.5 mm | 118.5 mm | over 5.93 mm |
| `snw` | 1.0 mm | **0** | 1.0 mm | over 0.05 mm |
| `canopy` | 1.0 mm | **0.0565 mm** | 1.0 mm | over 0.05 mm |
| `evspsbl` | 0.05 mm/day | 1.75 | 1.75 | over 0.088 |
| `mrro` | 0.05 mm/day | 0.72 | 0.72 | over 0.036 |

`snw` is zero everywhere, and the reason sits upstream of the criterion: `generate.py` clips `tas = np.maximum(2.0, seasonal + noise)` and `snow_threshold_degC` is 0.0, so nothing ever falls as snow and the coldest day in the scored year is 4.54 °C. The README already says the floors keep "empty snow or channel stores well-defined", so the emptiness is known and I am not reporting it as a surprise. What I would add is its consequence: `requires.states: [mrso, snw, canopy]` obliges every model to declare a store that the criterion can then never see a difference in. It is a joining requirement with no discriminating power, and `reference_restless` demonstrates it — its `snw` and `canopy` departures are exactly `0.0`.

`canopy` is the channel that is *not* empty and is still governed by the floor. Its mean over the scored year is 0.0565 mm, so the 1 mm floor is 17.7 times the signal, and the check on that channel becomes "an absolute difference over 0.05 mm" — 88 percent of the channel's own mean — rather than five percent of it.

I could not find a calibration for 1.0 mm in `probe.yaml` or in the README. The `runoff_ratio: 0.01` a few lines above it carries one ("SAC-SMA can legitimately produce about 1.3% runoff"); the floor does not. If 1.0 mm is chosen for `mrso`, a floor per variable would say so out loud. As written, one constant picked for a 118 mm store also decides two channels holding 0 mm and 0.06 mm.

## 2. The separation is carried by one flux channel; the storage channels land just under the line

From the same run, the largest departure `reference_restless` shows per variable. The criterion compares `departure / scale` against `threshold = 0.05`:

| variable | max departure | scale | departure / scale | against the 0.05 limit |
| `mrro` | 0.983 | 0.723 | **1.36** | 27 times the limit |
| `evspsbl` | 0.0857 | 1.755 | 0.0488 | **97.7 percent of the limit** |
| `mrso` | 5.709 | 118.7 | 0.0481 | **96.2 percent of the limit** |

So the `1.36` the criterion reports is `mrro`, and the two remaining channels that are not vacuous miss by 2 and 4 percent. `WOBBLE = 0.3` sits right at the edge of what this criterion can see in the stores. It works today. But the margin is thin enough that it is worth writing down which constant decides it: the `0.05` threshold, the `1.0 mm` floor and the case's own weather each move `mrso` and `evspsbl` across that line, and neither would be noticed, because the gate only asks whether the baseline trips at all.

## 3. A code path with a test and no caller

`closure` and `state_bounds` now take `params.get("phase")` and pass it to `make_window`. No probe sets a `phase:` key — a grep over `probes/` finds none — and this probe's own `probe.yaml` uses `all_variants: true` for both criteria instead. So the phase-scoped window is a third window mode with a unit test but no caller, and it re-implements the "find the contiguous labelled block, take the row before it, reject a non-contiguous one" logic that `criteria/spinup.py:_evaluation` already implements, down to near-identical error strings. Two implementations of one rule is the thing I would remove before merge.

## 4. Text

README lines 173 to 175 break mid-sentence — a stray `floors` sits on its own line.

The cycle count also excludes the spin-up year. The scored windows start at rows 2191, 3287 and 3652 of a record whose first row is already day one of the cycle, so a model has been through **six, nine and ten** repetitions of the same year when it is scored, not five, eight and nine. `EVALUATION_CYCLES = {short: 5, plus3: 8, long: 9}` is the offset inside the ten-year period, and the spin-up year is one more. It changes no result and not the coprime argument, only the sentence.

## 5. The description

It still describes spinning up for `N` cycles and for `N + 4` cycles, with four extra cycles serving as a leap-phase guard. The branch now runs three variants at `N`, `N + 3` and `N + 4`, and handles the leap phase by mapping 29 February onto 28 February's phase instead of by shifting the offset. There is also no `ht gate` line for the current head, and the merge checklist is not filled in.

---

Nothing here says the probe does not discriminate — on `mrro` it does, decisively. Items 1 and 2 are about how wide the rest of the check actually is; items 3 to 5 are cleanup and text. I would want item 1 answered with a number before merge, because it decides what `snw` and `canopy` are doing in `requires.states` at all.

longkaihao:
Thanks  @kawh1111 .

Thanks for the additional calibration review. I addressed the probe-local items in commit `8c5a7fd`.

The storage criterion now supports explicit per-variable floors. This probe uses 1.0 mm for `mrso`, 1.0 mm for the empty `snw` channel, and 0.01 mm for `canopy`; the scalar 1.0 mm value remains the fallback for other stores and for total reported storage. The README now records the independent-seed means and explains that the snow floor is an absolute conditioning bound, while the canopy floor keeps the 5% relative rule active.

The README also now documents that 5, 8, and 9 are period offsets, while the preceding spin-up year means that 6, 9, and 10 complete prior cycles are present before the shared 2007 evaluation year. The broken `floors` line and the diagnostic margin discussion have been corrected.

I kept the generic phase-window path unchanged because it is framework-level code and is outside this probe change. No current probe configures that path; removing it should be handled as a separate framework cleanup.

Validation after the change: 226 targeted tests passed, 1044 full tests passed with 8 skipped, `ht validate` passed for 29 probes and 60 models, and both the spin-up gate and full `ht gate` passed.

kawh1111 (dismissed):
Verified at `8c5a7fd` — I built all three variants on seed 20260912 and ran the reference models locally rather than reading the diff.

**The per-variable floor does what the review asked for.** For `reference_bucket` the scored-year means are 118.5 mm (`mrso`), 0 (`snw`) and 0.0565 mm (`canopy`), and the floor that now applies to each is 1.0 / 1.0 / 0.01. The channel that changes is canopy: its scale is the signal itself, 0.0565, instead of the 1 mm floor, so the departure that fails it drops from 0.05 mm to **0.0028 mm** — 17.7 times tighter, which is exactly the gap between the floor and the signal the review was pointing at. `snw` keeps its 1 mm bound and the README now says why: an explicit absolute conditioning bound for a store that cannot fill.

**Nothing else moved.** `reference_restless` still reports 1.3595 against a 0.05 threshold — identical to the value I measured on `e841470` — and still trips only `spinup_cycle_invariance`. The gate separates all eight baselines as before, including `reference_cheater` tripping both `state_bounds` and `spinup_cycle_invariance`, which it also did on `e841470`, so the tighter canopy floor did not introduce that. `ht validate` reports 29 probes and 60 models; the 265 tests across the spin-up, docs and report files pass.

**The new test pins the behaviour rather than the number.** `test_storage_floors_can_be_calibrated_per_variable` builds canopy at 0.05 and 0.09, where the scalar 1.0 floor scores a departure of 0.04 and passes; only the per-variable 0.01 fails it. Dropping the `state_floors_mm` support turns it red. The criterion also records the floor it actually used in its diagnostics, which is what made this check quick.

The two text corrections are in as well: the `floors` line no longer breaks mid-sentence, and the cycle count now reads as period offsets of five, eight and nine with six, nine and ten complete prior cycles including the spin-up year.

**Declaration.** This probe scores `wflow_sbm`, `summa`, `cwatm` and `lisflood`, which I proposed as #29, #30, #28 and #20. I did not write those adapters and the clause does not bind a reviewer here, but it is the tie the section exists to surface, so I would rather state it than leave it unsaid.

**One thing I would still change, and it does not block this.** `state_floors_mm` keys are not validated. `state_floors.get(var, state_floor)` means a misspelled variable name — `canopyy: 0.01` — falls back to the scalar in silence, and the probe would then be scored with the floor the author was moving away from. `routing_conservation` refuses unknown parameter names and `event_water_closure` does the same; checking the keys against the state variables plus `total_reported_storage` would fit the shape the repo already uses. A follow-up commit, not a re-review.

One note on mechanics rather than content: the branch is `dirty` against main, so a merge will move the head, and if stale review dismissal is on this approval will go with it. I am satisfied with the probe itself and will re-approve after the merge if that is what it takes.

chrimerss (changes requested):
Thanks @longkaihao, and thanks @kawh1111 for the floor calibration. The leap-phase blocker from my last review is fixed.

## Fixed and verified at 8c5a7fd

- **Leap phase.** All three variants now score calendar year 2007, starting in 2001, 1998 and 1997.
  - Over seeds 0–199, none of the four must-pass models fails any criterion, and SAC-SMA's worst departure is 0.056%.
  - Seeds 171 and 177, which failed at 60.02% and 5.49% on 27f0037, are now at 0.014% and 0.005%.
  - `reference_restless` is caught on 200/200 seeds (minimum 91%), and `reference_leaky` and `reference_degenerate` on 200/200.
  - The history offsets are 1096, 1461 and 365 days, which share no common divisor.
  - LISFLOOD's 3652-day window is cut by row, not by date, so it covers every variant's full record.
- **Per-variable storage floors.** Canopy is now scored against its own 0.0565 mm signal rather than a 1 mm floor, and the new test fails without `state_floors_mm`. The tighter canopy floor causes no failure in the same 200-seed sweep.
- **Checks.**
  - `ht validate`: 29 probes, 60 models.
  - pytest: 1052 passed.
  - `ht gate --probe mass/spinup-cycle-invariance` and the full `ht gate` both pass.

## Before merge

1. **Regenerate the archive rows.**
   - Neither e841470 nor 8c5a7fd touched `models/result.csv`, so all ten rows for this probe still come from the 27f0037 design (2007/2010/2011 in one record, 1 mm canopy floor).
   - `ht run --gate-seeds` at 8c5a7fd no longer reproduces them. Every verdict is the same, but the recorded worst pair and departure differ. For example:
     - `sacsma_snow17`: `short<->plus3:gw 0.02%` archived, `short<->long:gw 0.00%` now.
     - `flex_topo`: `plus3<->long:mrro` archived, `short<->plus3:channel` now.
   - In Docker, cwatm and wflow_sbm drift the same way. For cwatm, `short<->plus3:gw 0.03%` is archived and `short<->long:gw 0.00%` is what the run gives now.
   - summa, dhbv2, lisflood and google_flood_forecast still reproduce, so six of the ten rows are stale.
   - AGENTS.md asks for these rows to come from `ht run --model <name> --probe mass/spinup-cycle-invariance --gate-seeds --csv models/result.csv`. Please regenerate all ten after the rebase below.
2. **Rebase on main.**
   - #79 and #100 merged, and the branch now conflicts in `README.md`, `models/result.csv` and `site/index.html`.
   - After the rebase, every "N of M" for a model this probe scores moves up by one from main's new totals. `tests/test_docs_in_sync.py` checks them.
3. **Update the PR description.**
   - It still describes two runs, N and N+4, and says the extra four cycles are "a leap-phase alignment guard".
   - It also says the probe keeps "the forcing and evaluation window identical between the two runs".
   - The branch now runs three variants with different dates and relies on a shared calendar year for alignment.

## Smaller items (fine as a follow-up)

- `state_floors_mm` keys aren't checked. A misspelt key such as `canopyy: 0.01` silently falls back to the scalar floor, and a flux name is silently ignored. `event_water_closure` and `total_storage_drift` already refuse unknown parameter names, and the same check would fit here.
- `make_window`'s `phase` branch still has no caller, and it duplicates `criteria/spinup.py:_evaluation`. I'm fine with removing it in a separate framework cleanup, as you suggest.

## Review status

- @kawh1111's approval on 8c5a7fd counts under GOVERNANCE.md's pool rule.
- @Barbhuiya12's approval on 27f0037 was dismissed by the later pushes.
- The probe still needs a second pool approval on the final head, and a green acceptance gate. The fork CI has not run on any push yet.

Barbhuiya12:
The `plus3` leap bug was mine and I should say so plainly, because the history reads as though @longkaihao introduced it.

I suggested the coprime offsets and checked the calendar constraint as "2007 and 2010 are both non-leap". That was the wrong invariant. February 29 repeating a day makes the forcing repeat on a four-year period, not annually, so what mattered was the leap *phase* of each evaluation year and not whether each was individually a leap year. 2007 sits at leap+3 and 2010 at leap+2, which is exactly the comparison the original `EXTRA_CYCLES = 4` existed to avoid.

The verification failure is the part I want to record. I ran three gate seeds and reported the alias closed. @chrimerss ran two hundred and found SAC-SMA failing at 60.02% on seed 171. Three seeds could not have found a one-percent failure rate; I had no basis for the confidence I expressed, and a `must_pass` physical model would have been failed by roughly three percent of random seed draws had it gone in.

I have reproduced the fix independently on `8c5a7fd`, in-process over seeds 0-199:

```
sacsma_snow17   0 failures   worst departure 0.056%
seed 171        60.02%  ->  0.014%
seed 177         5.49%  ->  0.005%
```

That matches @chrimerss's 0.056% on different hardware. Giving each variant its own start year and scoring one shared calendar year is a better answer than what I proposed: the evaluation forcing is now identical by construction rather than verified after the fact, and the 1096/1461/365-day offsets keep the coprimality without the calendar fighting it.

I am not re-approving yet. The three pre-merge items are outstanding, six of the ten archive rows still come from the `27f0037` design, and the branch does not currently merge, so any approval now would be against a head that is about to change — which is what happened to the last one. Once the rebase and the regenerated rows are in, I will re-run the 200-seed sweep and the hidden-clock regression against the final head and approve there if it holds.

On the two smaller items, for whatever my view is worth: validating `state_floors_mm` keys seems right to me, since a misspelt store silently falling back to the scalar floor is the kind of thing that passes review twice. The `phase` branch I would leave to the separate cleanup as you both suggest.

longkaihao:
Thanks for the detailed reviews. The final branch is now rebased on the current upstream `main` (`4441476`) and the final head is `5f0dae9`.

I addressed the blocking leap-phase issue by giving each variant its own absolute start date while scoring the same calendar year:

- `short`: starts in 2001
- `plus3`: starts in 1998
- `long`: starts in 1997
- shared evaluation year: 2007

The resulting history offsets are 1096, 1461, and 365 days, which have no common divisor greater than one. This preserves the coprime hidden-clock check while ensuring that all compared evaluation forcing is in the same leap phase. The variants no longer compare 2007, 2010, and 2011 from one shared calendar record.

The seed-171 and seed-177 SAC-SMA failures from the previous design are therefore removed. The independent 0–199 seed sweep reported during review found zero failures for all four must-pass references, with SAC-SMA's worst departure at 0.056%. The same sweep detected `reference_restless` on all 200 seeds, while `reference_leaky` and `reference_degenerate` were also detected on all 200 seeds.

The archive rows were regenerated after rebasing on current `main`. The Docker-capable submitted models were rerun on r830 rather than being manually fabricated. The current archive reflects the new three-variant case and current model versions. SUMMA now passes this probe. δHBV2 remains FAIL because its `mrso` output exceeds the declared state bound; this is a `state_bounds` failure, not a failure of the spin-up comparison itself. Google Flood Forecast and MODFLOW6 remain N/A because they do not expose the required probe interface.

The storage-floor calibration has also been completed. The probe now uses:

- `mrso: 1.0 mm`
- `snw: 1.0 mm`
- `canopy: 0.01 mm`

The criterion records the actual floor used in its diagnostics. It also rejects unknown keys such as `canopyy`, so a spelling mistake cannot silently fall back to the scalar floor. Regression coverage was added for both calibrated per-variable floors and invalid floor names.

The documentation and metadata were updated to match the final implementation:

- probe README;
- top-level README;
- roadmap;
- model cards;
- site tables and translations;
- contributor and citation records;
- current probe/model counts;
- PR description;
- archive rows.

The verdict remains `FAIL (VIOLATION)` when a model reports all required outputs but fails to reproduce the same evaluation cycle after different amounts of identical prior history. The failure message reports the observed disagreement and explicitly states that the signature does not identify whether the cause is a hidden state or a physical store that has not yet settled. `INCOMPLETE` is not used because the probe did receive the required outputs and completed its comparison.

The unused `make_window` phase branch was intentionally left for the separate framework cleanup discussed in the review. Removing it would be unrelated framework refactoring rather than part of this probe.

Final validation on the rebased head:

```text
ht validate
32 probes, 72 models

ht gate --probe mass/spinup-cycle-invariance
GATE PASSED

pytest -q
1165 passed, 8 skipped

**Review by Barbhuiya12 — APPROVED** [review]
Re-reviewed at `14608a8`, as I said I would once the branch was rebased and the
archive rows regenerated. Both are done: the branch is 0 commits behind `main`,
and all 14 archive rows for the probe are dated 20-23 September — none is left
from the `27f0037` design. `ht validate` (33 probes, 76 models), the full `ht
gate` and `pytest -q` (1218 passed) all pass here.

## The 200-seed sweep

All eight baselines over seeds 0-199, recording every criterion per seed rather
than only the verdict, so that a control catching the right thing can be told
apart from one catching the wrong thing:

```
model                  role        n     result
reference_bucket       must_pass  200    0 failures   spin-up worst 0.0000%
flex_lumped            must_pass  200    0 failures   spin-up worst 0.0000%
flex_topo              must_pass  200    0 failures   spin-up worst 0.0000%
sacsma_snow17          must_pass  200    0 failures   spin-up worst 0.0561%
reference_restless     must_fail  200    0 escapes    isolated on every seed   spin-up min 91.12%
reference_leaky        must_fail  200    0 escapes    isolated on every seed
reference_degenerate   must_fail  200    0 escapes    isolated on every seed
reference_cheater      must_fail  200    passes outright on 4 seeds — see below
```

`sacsma_snow17`'s worst departure is 0.0561%, the same 0.056% @chrimerss and I
measured on `8c5a7fd`. Seed 171 — the 60.02% failure I did not catch the first
time — is now 0.0144%. The one-shared-year design removed the leap-phase fault by
construction rather than by tuning, and it holds across the draw.
`reference_restless`, the model the probe exists for, is caught on all 200 seeds
at a minimum of 91% against a 5% threshold, isolated to
`spinup_cycle_invariance` every time.

## The hidden-clock regression

The suite passes (28 in `test_spinup_cycle_invariance.py`). Because the period
test builds its own three-variant configuration rather than reading the probe, I
also checked the design on the real cases, with the real start dates and the
probe's own parameters. The variants begin on 2001-01-01, 1998-01-01 and
1997-01-01, giving 2191, 3287 and 3652 days of history before 2007 and pairwise
offsets of 1096, 365 and 1461 days, whose greatest common divisor is 1.

Sweeping a hidden clock counted in days since the run began, at 10% amplitude,
over every period from 2 to 60 days plus 73, 137, 365, 487, 731, 1096 and 1461,
the only clocks that stay under the 5% threshold are those whose period divides
365. That is structural and exact: 1096 = 3·365 + 1 and 1461 = 4·365 + 1, so
against any period dividing 365 the variants are shifted by exactly one day. The
prediction for a one-day shift, $2A\sin(\pi/T)$, against what the criterion
measured:

```
T = 73 d    predicted 0.0086   measured 0.008604   (third pair aligned to 1e-15)
T = 365 d   predicted 0.0017   measured 0.001721   (third pair aligned to 1e-15)
```

I read this as correct behaviour rather than a gap. A clock periodic in the
forcing's own 365-day cycle is a seasonal cycle, which is what a converged model
looks like; the one day between variants is the leap day, and the probe is right
not to call that a hidden state. Every clock that accumulates across cycles is
caught.

One small thing in the period test, not blocking. Its two assertions of the form

```python
assert any("short<->plus3:mrro" in key for key in result.diagnostics["deviations"])
```

only check that the key exists, and `deviations` records every pair whether or
not it deviated. At `period=3`, cycles 5 and 8 agree mod 3, so that pair measures
exactly 0.0 and the assertion still passes. `assert not result.passed` is the
only line doing any work. Asserting on the value — that the pair a given period
cannot alias exceeds the threshold — would make the test say what its docstring
claims.

## `reference_cheater` passes outright on 2% of seeds

On seeds 43, 103, 164 and 186 it trips nothing and the probe returns PASS. That
is the remainder of what @chrimerss listed as a nit on 14 September — six seeds
then, four now, since `spinup_cycle_invariance` has since started catching 162
and 167.

The cause is exact. The cheater does no hydrology: it sets
`soil += 0.65·pr − 0.55·pet` and lets the storage absorb whatever closes the
budget. It is caught only if that drift carries the storage out of
`[0, soil_capacity]` within the record, or makes the evaluation years disagree.
Measuring the drift per cycle as a share of half the soil capacity:

```
seeds where it passes     max 0.009   (n = 4)
seeds where it is caught  min 0.011   median 0.381   (n = 196)
```

The separation is clean. On the four seeds the climate happens to balance
0.65 P against 0.55 PET, so the invented storage is bounded *and* periodic — it
closes, stays in range and converges. On those draws it violates nothing this
probe can observe. I do not think that is a defect in the probe; it is the
control failing to exhibit its defect on a climate it was not written for. The
gate seeds catch it every time, which is why @chrimerss rated it a nit, and I
agree. If it is worth closing, the place is the cheater's fractions — a drift
guaranteed to be non-zero on this climate — rather than any criterion here.

One consequence for #134's new isolation check: `reference_cheater` trips
`spinup_cycle_invariance` as well as `state_bounds` on 190 of 200 seeds, and I
would not ask for that to be isolated. A storage that drifts without bound is
exactly a state that never reaches a periodic attractor, so this probe's own
criterion is correctly finding a second real defect in the same model. That is
unlike #102, where the double trip hid which criterion the control demonstrated.

## The earlier items

- **Leap phase:** closed by giving each variant its own start date while scoring
  one shared year, as above.
- **`suspicious_exact`:** survives the per-variant aggregation;
  `test_exact_closure_flag_survives_per_variant_aggregation` pins it.
- **SAC-SMA on seed 69:** the `non_degenerate` runoff floor is now 0.01, clearing
  its 0.0134.
- **Storage floors:** calibrated per variable (`mrso` 1.0, `snw` 1.0, `canopy`
  0.01 mm), recorded in the diagnostics, and a misspelt name such as `canopyy` is
  now rejected rather than silently falling back to the scalar floor.
- **The FAIL message** no longer disowns its own verdict; it says the signature
  cannot tell a hidden state from an unsettled store, which is the honest
  statement and the one the 40-year CWatM run supports.

## Approving

The fault that cost the last approval is fixed by construction and holds over 200
seeds; the design catches every hidden clock that accumulates across cycles; the
three controls built for this probe are caught and isolated on every seed; and the
archive reflects the final case. The cheater's four seeds are a property of that
shared control on this climate, not of the probe, and are worth recording here for
whoever picks it up.

PR #79, opened 2026-09-13
[PROBE: mass] Groundwater-river exchange and storage close the water budget

### LINKED ISSUE #25: [MODEL] MODFLOW 6
### Model name

MODFLOW 6

### Model category

physics

### Model code repository

https://github.com/MODFLOW-ORG/modflow6

### Version, release, or commit

MODFLOW 6.7.0

### Weights repository (optional)

_No response_

### Paper or model reference (optional)

_No response_

### Code and weights licenses

Code: CC0-1.0; weights: N/A (no pretrained weights).

### Supported timestep

PT1D (daily)

### Required inputs

Forcing: groundwater recharge (L/T); surface-water stage or other time-varying boundary conditions where applicable (L)
Static: model grid and layer geometry, top and bottom elevations (L), hydraulic conductivity (L/T), storage properties (specific yield and/or specific storage), and boundary-condition configuration
Initialization: initial groundwater head (L); optional spin-up to establish a dynamically consistent initial state

### Outputs produced by the model

dis — river discharge

### Evaluation window (optional)

_No response_

### Packaging status

Help is needed with both Docker packaging and the adapter

### Installation and inference notes

MODFLOW 6 is a compiled, stand-alone groundwater modeling program and does not require pretrained weights. Precompiled executables are available with the official release, and the source code can also be compiled following the MODFLOW 6 build instructions. A simulation is run by executing the MODFLOW 6 executable (mf6) in a directory containing the required simulation input files. No network access is required at runtime once the executable and model inputs are available. For a small synthetic groundwater test case, computational and memory requirements are expected to be modest.

### Why this model belongs in the benchmark

MODFLOW 6 is a widely used, physics-based groundwater flow model developed by the U.S. Geological Survey. Including it would extend HydroTuring's model set beyond rainfall-runoff models and provide a well-established process-based reference for evaluating groundwater behavior. In particular, MODFLOW 6 explicitly represents groundwater heads, storage changes, and fluxes across hydraulic boundaries, and can represent groundwater–surface water exchanges through relevant packages. Benchmarking MODFLOW 6 would help assess whether HydroTuring's physical verification framework can be generalized to groundwater states and fluxes, which are not fully represented in the current standard output vocabulary.

### Submission confirmation

- [x] I have identified a reproducible code revision and have permission to use the linked code and weights for this evaluation.
- [x] I understand that HydroTuring may report PASS, VIOLATION, INCOMPLETE, INCOMPATIBLE, or ERROR.

### LINKED ISSUE #40: [PROBE] Net groundwater–surface water exchange consistency
### Conservation law

mass

### Residual equation

```math
R = Q_{\mathrm{GW\rightarrow SW}} - Q_{\mathrm{SW\rightarrow GW}} - Q_{\mathrm{GW,net}}
```

where $Q_{\mathrm{GW\rightarrow SW}}$ is gross groundwater discharge to surface water, $Q_{\mathrm{SW\rightarrow GW}}$ is gross surface-water leakage to groundwater, and $Q_{\mathrm{GW,net}}$ is the independently reported net groundwater contribution to surface water. All terms have units of water volume per unit time ($L^3T^{-1}$), or equivalent depth per unit time ($LT^{-1}$).

The relative residual is evaluated against the gross groundwater–surface water exchange and must remain below the suite's 5% engineering tolerance. Gross exchange is used as the denominator because large opposing groundwater–surface water fluxes may produce a small net exchange. Normalizing by the net flux would become unstable when gaining and losing fluxes nearly cancel. An absolute tolerance is used when gross exchange approaches zero.

A model can close its overall water budget while reporting internally inconsistent groundwater–surface water exchange components. For example, it may report 100 units of groundwater discharge to surface water and 30 units of surface-water leakage to groundwater, but report a net groundwater contribution of 100 rather than 70.

Such a model could still satisfy the catchment-scale water budget if the discrepancy is compensated by another storage or boundary term. A conventional global budget-closure test could therefore pass it, whereas this probe would fail because

$$ 100-30-100=-30. $$

Other deliberately broken models can reverse the sign convention or omit one direction of exchange while preserving the global water balance. The exchange-consistency residual catches these failures.

Generate a small synthetic coupled groundwater–surface water system using a fixed random seed. The case contains both gaining and losing surface-water reaches so that groundwater-to-surface-water and surface-water-to-groundwater fluxes are simultaneously nonzero.

The forcing and boundary conditions are chosen to produce appreciable exchange in both directions while keeping the system numerically simple. The model is asked to report gross groundwater-to-surface-water exchange, gross surface-water-to-groundwater exchange, and net groundwater contribution to surface water over the same spatial domain and time window.

No observational data or prescribed "correct" exchange magnitude is required. The probe tests only the mass-accounting identity among the three model-reported quantities.

### Will you build it?

Yes — assign it to me

yaji-wang:
Adds the probe `mass/gw-sw-exchange-consistency` (closes #40) and the MODFLOW 6 model (closes #25). Both were proposed by Yaji Wang (University of Illinois Urbana-Champaign).

## The probe

A two-year daily aquifer case, after a 90-day spinup, driven by seeded groundwater recharge (`gw_recharge`, mm/day) and river stage (`sw_stage_m`, m). The model reports:
- `gw_sw_exchange`: the net river-aquifer exchange, positive into the aquifer;
- `gw_to_sw <= 0` and `sw_to_gw >= 0`: its signed components;
- `gw`: the aquifer store;
- `gw_boundary` (optional): a boundary term acting on the aquifer, such as a GHB or a well.

| Criterion | Asserts |
| --- | --- |
| `exchange_components` | `gw_to_sw + sw_to_gw = gw_sw_exchange` (relative 1e-6, absolute 1e-4 mm), and neither component sits on the wrong side of zero |
| `groundwater_balance` | `gw_recharge + gw_sw_exchange + gw_boundary` equals the change in `gw` on every step and over the record. The step allowance is 1% or a floor of `max(2e-4 mm, min(1e-5 × max|gw|, 0.05 mm))`; the record allowance is `max(1% of recharge, 2 × floor)` |
| `state_bounds` | `gw >= 0` |
| `exchange_directions` | the aquifer loses water to the river and gains water from it by at least 0.1 mm each over the scored record |

`groundwater_balance` and `exchange_components` report the worse residual divided by its allowance, against 1. `case.min_window_days: 180` keeps a submitted model's window across two stage cycles.

## Contract changes

- **New flux names** in `FLUX_VARS`, `UNITS` and the AGENTS.md table: `gw_sw_exchange`, `gw_to_sw`, `sw_to_gw` and `gw_boundary`.
  - River-aquifer exchange happens inside a catchment, so it is never `gwex`.
  - A boundary flow on the aquifer that also crosses the catchment boundary is reported in both `gwex` and `gw_boundary`. No budget adds the two.
- **New criteria** in `src/hydroturing/criteria/groundwater.py`, listed in `docs/writing-a-probe.md`.
- **`must_pass` exception** in `docs/writing-a-probe.md`: a probe about a process that none of the four physical models has may gate on an exact reference, with an archived physical model (here modflow6) as the independent check.

## Baselines

- **`must_pass`: `reference_exchange_exact`.** A trusted, Docker-free bookkeeping reference that closes its groundwater budget exactly.
- **`must_fail`: `reference_exchange_sign_error` → `exchange_components`.** It reports `gw_to_sw` with its sign flipped on alternating steps, and trips nothing else.

## Models

- **`modflow6` (6.7.0-adapter.1).** MODFLOW 6.7.0 run through flopy in an amd64 Docker image, with one river cell. Its RIV and STO budget terms are reported as signed exchange and storage. It is archived as a submitted model (PASS, 1 of 1 probe scored) and is not in the gate.
- **Every other evaluated model** is N/A (INCOMPLETE) here, because none of them reports the exchange components.

## Archive

`models/result.csv` has gate-seed rows for:
- modflow6 on all 29 probes, plus its adapter-contract row;
- both references;
- every evaluated model on this probe.

## Checks at the merge head

- `ht validate`: 29 probes, 63 models
- `pytest -q`: 1084 passed
- `ht gate`: all 29 probes pass
- modflow6's archived row reproduces through Docker

Closes #40
Closes #25


chrimerss (changes requested):
@yaji-wang Please refer to the auto review and check if they make sense to your case.

yaji-wang:
> @yaji-wang Please refer to the auto review and check if they make sense to your case.

Hi @chrimerss, thanks for the detailed review. Good catch on the `gwex` sign. It was really worth catching before it propagated further. 

Just pushed a follow-up commit `7ba019e` addressing the review point by point. Fixed the `gwex` sign inversion (now matches `AGENTS.md`/`closure.py`: `recharge + gwex = dS`, `sw_to_gw >= 0`, `gw_to_sw <= 0`), closed out the `AGENTS.md` merge checklist, added `min_window_days: 180`, rewrote `reference_exchange_sign_error` to trip only `exchange_components`, reverted `reference_bucket`, and fixed `modflow6`'s storage term to use MODFLOW's own STO budget (residual now ~3e-10 mm).

I left review point 4 open for discussion — `must_pass` is still `modflow6` alone, since none of the four standard physical baselines consume `gw_recharge`/`sw_stage_m`. I didn't want to make that design call without checking with you first.

`ht validate`, `ht gate --probe mass/gw-sw-exchange-consistency`, and `pytest -q` (823 passed) all green. Happy to adjust further if anything's off.

chrimerss:
Thanks for the quick revision, @yaji-wang. 7ba019e fixes most of the first review, and the sign convention is now right in code. I checked the head locally:

- `ht validate` and all 831 tests pass.
- `groundwater_balance` now checks `recharge + gwex = dS`, and modflow6 maps RIV `q > 0` to `sw_to_gw`.
- `reference_exchange_sign_error` trips only `exchange_components`. That holds on the 5 gate seeds and 45 random seeds, over the full record and the 180-day window.
- Every new archive row reproduces byte for byte with `--gate-seeds`. modflow6's PASS row reproduces under amd64 (see 2).
- modflow6 now reads storage from MODFLOW's own STO budget, and its balance closes to 3.7e-10 mm.

Three things still need to change before this can merge.

**1. `gw_to_sw` and `sw_to_gw` cannot be components of `gwex`.**

In the rest of the suite, `gwex` is water crossing the catchment boundary. Eight models emit it:

- `flex_lumped`, `flex_topo`, `reference_bucket` and `cwatm` report a prescribed withdrawal.
- `sacsma_snow17`, `lisflood` and `wflow_sbm` report deep loss or leakage, plus any withdrawal.
- `dhbv2` reports regional groundwater exchange.

`closure.py` counts `gwex` as an outside source. River-aquifer exchange is different. In a catchment model it moves water between `gw` and `channel`, and both stores are inside the control volume. "Into the aquifer" matches "into the catchment" only when the aquifer is the whole model. That holds for modflow6 and for no other model here.

The conflict appears once a catchment model reports the components. Take a day with 0.5 mm of baseflow to the river and 0.1 mm of groundwater abstraction:

- Reported honestly, `gw_to_sw = -0.5` and `gwex = -0.1`. `exchange_components` then fails by 0.4 mm.
- Forced to fit, `gwex = -0.5`. `mass/catchment-closure` then counts 0.5 mm of internal flow as an outside sink and misses the 0.1 mm withdrawal, a 0.4 mm error.

So a correct catchment model cannot pass both probes. No archived row is wrong today, because only modflow6 and the reference emit the components. But `FLUX_VARS` and the AGENTS.md table are the contract every adapter follows. Please:

- Add a separate net flux for aquifer-river exchange, for example `gw_sw_exchange`, positive into the aquifer. Define `gw_to_sw` and `sw_to_gw` as its components.
- Give `exchange_components` a `net` parameter in place of the hardcoded `gwex` (`exchange.py:22, 29`). Point the `exchange` parameter of `groundwater_balance` at the new flux.
- Emit the new flux from modflow6 and the reference instead of `gwex`. Update AGENTS.md, `spec.py` and the probe README to match.

**2. The modflow6 image does not run on arm64.**

The Dockerfile starts from `python:3.12-slim` without a platform, then installs the x86-64 `mf6.7.0_linux.zip`. On an Apple Silicon host the image builds as arm64. mf6 then stops with `rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2`, and `ht gate` fails. MODFLOW 6.7.0 ships no Linux arm64 build, so please pin the platform as `models/lisflood/Dockerfile` does:

```dockerfile
FROM --platform=linux/amd64 python:3.12-slim
```

With that one line, both this probe's gate and the full 24-probe gate pass on my machine.

**3. `must_pass` needs a baseline that runs without Docker.**

On your open question: I agree the four physical models cannot be put to this probe. But `ht gate` runs on every probe and harness pull request. With modflow6 as the only `must_pass`, each run downloads a 134 MB release and installs flopy before it scores anything. If GitHub Releases or PyPI is slow or down, the gate fails for every pull request.

Your reference already holds the answer. With the sign flip removed, it passes all four criteria on the same 50 seeds, over the full record and the 180-day window. Please:

- Add that model as a trusted exact reference, for example `reference_exchange_exact`, in `TRUSTED_SUBPROCESS_MODELS`. Make it the `must_pass`.
- Keep modflow6 as a submitted model with its archive rows, outside the gate. List it with the submitted models in the README and on the site, not with the physical references.

**Smaller fixes**

- `spec.py:42-43` still gives the old signs ("gw_to_sw is positive into surface water").
- AGENTS.md, `probe.yaml:55` and the probe README say models without the components are `N/A (INCOMPATIBLE)`. The harness and the archive give `N/A (INCOMPLETE)`, because a missing output is checked first. INCOMPATIBLE applies only to a model that reports the components without declaring the inputs.
- modflow6 builds a fixed 1 km² grid (`NROW`, `DELR`, `DELC`) but converts fluxes with `area_km2`. With `area_km2 = 2`, reported exchange and storage would cover twice the area the recharge fell on. Please size the grid from `area_km2`, or convert over the grid's own area.
- `aquifer_storage_coefficient` reaches MODFLOW as specific storage in 1/m, because `storagecoefficient=False` (`ht_adapter.py:102`). Please set it to `True` or rename the key, and list every static key with its units in the probe README.
- The 2e-4 mm floor in `groundwater_balance` is tighter than common output formats. `gw` sits near 2000 mm, so a closed budget written with 6 significant figures fails on every seed, by up to 49 times the allowance. `%.6f` passes. Please state in the README that these columns need full precision, or scale the floor with the store.
- modflow6's `model.yaml` still has `version: "0.1.0"`, `authors: [Local HydroTuring contributor]` and a free-text licence. Please follow the other submitted models: a version naming MODFLOW 6.7.0 and an adapter revision, real authors, `license: CC0-1.0` as in #25, and a `reference:` link. The version keys all 24 archive rows, so fix it before regenerating them.
- `models/result.csv` now starts with a UTF-8 byte-order mark. Please strip it.
- `README.md:82` now says "nine more we want". ROADMAP still lists ten unclaimed probes, so it should stay "ten".
- `CITATION.cff` needs your ORCID (item 3 of the AGENTS.md merge checklist).
- Five files still lack a trailing newline: the modflow6 `Dockerfile`, `generate.py`, `probe.yaml`, `exchange.py` and `groundwater.py`.
- The reference's sign error trips both arms of `exchange_components` on the same steps, so deleting either arm keeps the gate green. Please add one unit test per arm: swapped component columns (sign arm only) and a net that omits one component (residual arm only).
- Optional, but closer to #40: with one river cell, `gw_to_sw` and `sw_to_gw` are never both nonzero on the same step. The "100 − 30 reported as 100" case from #40 therefore never arises. A second river cell at a different stage would exercise it, if the adapter splits each cell's `q` by sign before summing.
- Optional: `rbot` moves with the stage (stage − 1 m), which caps river leakage at 0.15 mm/day whatever the stage. modflow6's `sw_to_gw` reaches exactly 0.15. A fixed riverbed elevation below the lowest stage would be more physical. Otherwise, please note the simplification in the README.

After these changes, please regenerate the archive rows with `ht run --gate-seeds --csv models/result.csv`. CI has not run on your fork yet. I will start it after your next push.

yaji-wang:
@chrimerss Thanks for the thorough follow-up review. I agree with all three of your main points from the previous round: the first was a genuine misunderstanding on my part about what `gwex` actually means in this repo (a boundary-crossing exchange). The other two (the arm64 platform pin and a Docker-free  `must_pass` baseline) I also fixed as you described. Pushed all the fixes to the same branch. The tip is now `90f3d68`, which should cover everything from this round: the sign comment in `spec.py`, the INCOMPATIBLE/INCOMPLETE wording, `modflow6`'s grid now sized from `area_km2`, `storagecoefficient` set correctly, the `groundwater_balance` floor widened to clear common output precision, `modflow6/model.yaml`'s version/authors/license brought in line with the other submitted models (and all 24 archive rows regenerated with the plain `ht run --model <name> --gate-seeds --csv` form), the BOM and trailing newlines cleaned up, the probe count fixed in README, a unit test for each arm of `exchange_components`, and my ORCID added to CITATION.cff.

The two optional items (a second river cell, a fixed riverbed elevation) I left as documented simplifications in `models/modflow6/README.md` rather than changing the physical setup.

`ht validate`, `ht gate --probe mass/gw-sw-exchange-consistency`, and `pytest -q` (827 passed) all green on `90f3d68`. Ready for CI whenever you'd like to kick it off.

chrimerss (changes requested):
Thanks, @yaji-wang. 90f3d68 fixes all three blocking points and nearly every smaller one. I checked the head locally:

- `ht validate` and all 835 tests pass.
- This probe's gate passes in 1.3 s with no Docker, and the full 24-probe gate passes.
- `reference_exchange_exact` passes all four criteria on 50 seeds, over the full record and the 180-day window. `reference_exchange_sign_error` still trips only `exchange_components`.
- `gw_sw_exchange` is fully separate from `gwex`. Nothing outside the new criteria reads the three new columns, so a catchment model can report river exchange and a withdrawal side by side.
- All 81 new archive rows reproduce byte for byte with `--gate-seeds`, including modflow6's PASS row from the pinned image.

One thing still needs to change before this can merge.

**1. The 2e-2 mm floor lets `groundwater_balance` pass a leaking aquifer.**

The floor went from 2e-4 to 2e-2 mm. Mean recharge in this case is about 0.02 mm/day, so the floor now sets the allowance on every step and the 1% term never acts. On the 5 gate seeds, all of these pass:

| Unreported loss from `gw` | Share of total recharge | `groundwater_balance` |
| --- | --- | --- |
| 0.0199 mm/day | 89–104% | PASS 5/5 |
| 0.01 mm/day | 45–52% | PASS 5/5 |
| 0.005 mm/day | 22–26% | PASS 5/5 |
| 10% of each day's recharge | 10% | PASS 5/5 |

The criterion already computes the cumulative residual as its `value` (`groundwater.py:42`). For the first row that value is 0.89–1.04 against a threshold of 0.01. But `ok` (`groundwater.py:41`) only checks single steps.

Rounding and a leak behave differently over the record. Rounding a stored state cancels out when the steps are summed, while a leak adds up. Honest `gw` output at 6 significant figures has a cumulative residual of 0.0. Please gate on the cumulative residual as well. Two ways, both checked on the gate seeds:

- Pass only if every step is within the current allowance **and** `|Σ residual| <= rel_tol × Σ recharge`. This catches every row in the table, and honest output at full precision or 6 significant figures still passes.
- The same, with the per-step floor scaled to the store, `max(abs_tol_mm, 1e-5 × max|gw|)`, and a cumulative allowance of `max(rel_tol × Σ recharge, 2 × floor)`. This catches the same leaks. It also passes an honest model whose `gw` sits 10,000 mm higher and is written at 6 significant figures. That model fails today. States are absolute storages, so the datum is the model's choice.

Please update the tolerance paragraph in `probes/mass/gw-sw-exchange-consistency/README.md:80-88` to match.

**Smaller fixes**

- The site still lists modflow6 as `physical` (`site/index.html:699`), `físico` (`:909`) and `物理模型` (`:1119`). Please mark it as submitted and move it after lisflood, as in the README.
- `exchange_components` uses `abs_tol: 1.0e-6` (`probe.yaml:38`). That fails an honest two-reach model, the #40 case, once each column is written at ordinary precision. With a second reach that gains and loses on the same step, `%.6f` output fails on 5 of 5 gate seeds and `%.6g` on 1 of 5. At `abs_tol: 1.0e-4` both pass, while `reference_exchange_sign_error` and a net that omits `sw_to_gw` still fail. Please raise it, and drop the sentence in the probe README that says the precision concern does not apply there (`:86-88`).
- The unit tests pin each arm, but not the two halves of the sign arm (`exchange.py:37`). The swapped-columns test breaks both halves at once. So deleting `(gw_to_sw > 1.0e-9).any()` alone keeps every test green, and the gate stays green too. Please add two zero-residual tables: one with a positive `gw_to_sw` only, and one with a negative `sw_to_gw` only. The tests also call the criterion with `{}` instead of the probe's parameters (`test_gw_sw_exchange_components.py:67, 81, 96`), and `make_window` is imported but unused (`:21`).
- `groundwater_balance` computes `recharge + exchange - dS` with no other term. A MODFLOW model with a GHB or WEL boundary would report that flow as `gwex` and fail. Please either add an optional `sources` list, as `closure` has, or state in the probe README and AGENTS.md that `gwex` must be zero or absent on this probe.
- The probe README says `reference_exchange_exact` is not exercised by CI (`:71-73`). The `gate` job runs `ht gate`, which runs it as the must_pass. The modflow6 command on line 73 also needs `--gate-seeds --csv models/result.csv`.
- `models/result.csv` now has 49 duplicated (model, version, probe) keys, 26 of them byte-identical rows:
  - `reference_exchange_exact`: its 23 N/A rows appear twice.
  - `reference_exchange_sign_error`: its 23 N/A rows from 2026-09-13 are still there next to the 2026-09-14 copies.
  - `flex_lumped`, `flex_topo` and `sacsma_snow17`: two rows each on this probe.

  Please keep one row per key.
- `models/modflow6/ht_adapter.py:17` still writes version `0.1.0` into `run.json`. The manifest says `6.7.0-adapter.1`.
- The `#40` link in `models/modflow6/README.md:31` is relative and resolves to a 404. The other model READMEs use the full issue URL.
- `generate.py:17` emits `river_stage_reference_m`, which is not in the static-key table. Please document it or drop it.
- `reference_exchange_exact` declares `aquifer_storage_coefficient` in `needs_static` but never reads it. Its head is a moving average of the stage, not a reservoir. Please describe it as a bookkeeping reference, not a linear-reservoir aquifer, in its `model.yaml`, the probe README and `README.md`.

After these changes, please regenerate the archive rows once. I will start CI on your next push.

yaji-wang:
Hi, @chrimerss, sorry for the late reply. Thanks for the detailed suggestions! Here is my update:

## Main fix

Added a cumulative-residual gate to `groundwater_balance` (option 2 from the review): the per-step floor is now scaled to the store, `max(abs_tol_mm, 1e-5 × max|gw|)`, and the record must also satisfy `|Σ residual| <= max(rel_tol × Σ recharge, 2 × floor)`. `abs_tol_mm` was restored to `2e-4` since the storage-rounding term now carries that job. Verified on the gate seeds: all four leak scenarios from the review now fail, and an honest model passes at full precision, at 6 significant figures, and with `gw` offset by +10,000 mm at 6 significant figures. Updated the tolerance paragraph in `probes/mass/gw-sw-exchange-consistency/README.md:80-88` to match.

## Also addressed in [`b7d6c8a`](https://github.com/yaji-wang/HydroTuring/commit/b7d6c8af932bcef291313ab2903d5e3288646d14)

- `exchange_components`: `abs_tol` raised `1e-6 → 1e-4` so honest rounding no longer fails a two-reach model.
- `groundwater_balance`: added an optional `sources`/`gwex` term for GHB/WEL-style boundaries.
- Added tests pinning each half of the sign-arm check independently; fixed tests to use real probe params instead of `{}`; dropped the unused `make_window` import.
- Site now shows `modflow6` as submitted, after `lisflood`, in all 3 languages.
- Fixed `modflow6`'s version string, its dead `#40` link, and described `reference_exchange_exact`/`sign_error` as bookkeeping references instead of linear-reservoir aquifers.
- Corrected the README's CI/baseline claims and archive command; dropped the unused `river_stage_reference_m` key.
- Removed 49 duplicate rows this branch had added to `models/result.csv`.

All 829 tests pass; baselines and `ht validate` verified unchanged.

Please let me know if there is anything else needed from myside.

chrimerss (changes requested):
Thanks, @yaji-wang, and no worries about the timing. b7d6c8a does everything the last review asked for. On that commit:

- `ht validate` and all 837 tests pass, and so do this probe's gate and the full 24-probe gate.
- All 81 archive rows reproduce byte for byte with `--gate-seeds`, including modflow6's PASS row through Docker. No duplicate rows remain.
- The leak cases from my last review now fail. Honest output passes at full precision and at 6 significant figures.
- `reference_exchange_sign_error` still trips only `exchange_components`, and the site lists modflow6 as submitted in all three languages.

Thanks also for merging main in 207d8dd. Unfortunately that merge kept this branch's side of every conflict, so it deletes what main had added in those places. Three things need to change before this can merge. Two of them come from what I suggested last round, so they are on me rather than on you.

**1. Please redo the merge with main.**

At 207d8dd, `ht validate` fails, pytest reports 220 failures and 32 errors, and `ht gate` stops before it scores anything. The conflict resolution removed these from main:

- `src/hydroturing/spec.py`: seven entries in `TRUSTED_SUBPROCESS_MODELS` (`reference_snyder_router`, `reference_instant_router`, `reference_inverse_router`, `reference_rating`, `reference_rating_drift`, `reference_rating_inverted` and `reference_flat_stage`). Their manifests are now rejected, and that is what stops `ht validate` and `ht gate`.
- `CITATION.cff`: four probe authors (Shunan Zhou, Binlan Zhang, Siddik Barbhuiya and Bing Li).
- `models/result.csv`: 73 of main's rows. These include every model's rows on the five probes merged since your base and all of `flex_lumped`'s 1.1.0 rows.
- `README.md`: the 28-probe paragraph and the current model standings, which are back to their old counts.
- `site/index.html`: the `ungauged` and `drift` flowchart entries, the `ungauged` labels, and the current model counts.

Please reset the branch to b7d6c8a and merge main locally, with `upstream` pointing at Flood-Lab/HydroTuring:

```bash
git reset --hard b7d6c8a
git fetch upstream
git merge upstream/main
```

In each conflict, keep main's lines and add yours. For `TRUSTED_SUBPROCESS_MODELS` and `CITATION.cff` that means the union of both sides. For `models/result.csv`, keep all of main's rows and append yours. For the README and site, start from main's text and add this probe to it. Then please:

- Restate the probe counts for 29 probes in the README and in all three site languages. `tests/test_docs_in_sync.py` points to each one.
- Add an N/A row for `flex_lumped` 1.1.0 on this probe. Main bumped it from 1.0.0.
- Add modflow6 rows for the five probes merged since your base: `energy/snowmelt-energy-water`, `mass/multi-decadal-drift`, `mass/ungauged-basin-closure`, `momentum/routing-lag-consistency` and `momentum/stage-discharge-monotonic`.

`ht validate`, `pytest -q` and `ht gate` should all pass on the merged head before you push it with `git push --force-with-lease`. If you would rather not redo the merge, tell me and I can push a corrected one to your branch, since maintainer edits are enabled.

A heads-up on #94: it adds its own `src/hydroturing/criteria/exchange.py`, so whichever PR merges second hits an add/add conflict. The criterion names do not clash. The simplest way out is to move `exchange_components` into `groundwater.py`, next to `exchange_directions`, and drop this PR's `exchange.py`.

**2. The store-scaled floor lets a model set its own tolerance.**

I suggested scaling the per-step floor with `max|gw|`. That was a mistake. `gw` is an absolute storage, so a model can raise its own floor by reporting its storage higher. On the 5 gate seeds, with the storage shifted up and everything else unchanged:

| `gw` shifted up by | Hidden fault | `groundwater_balance` |
| --- | --- | --- |
| 1e6 mm | a 0.0199 mm/day leak, about all of the recharge | PASS 5/5 |
| 1e6 mm | a 5 mm one-step storage jump that no flux explains | PASS 5/5 |
| 2e5 mm | a 0.005 mm/day leak | PASS 5/5 |
| 1e5 mm | 10% of each day's recharge dropped | PASS 5/5 |

Without the shift, every one of these fails. Please cap the storage term, for example `storage_floor = min(1e-5 × max|gw|, 0.05)` (`groundwater.py:53`). With that cap, all four rows fail. Honest output still passes at full precision, at 6 significant figures, and with `gw` shifted up by 10,000 or 100,000 mm at 8 significant figures. At 6 significant figures, a `gw` of 10,000 mm or more would then fail. Please ask in the probe README for `gw` at full precision, or at least 8 significant figures.

**3. `sources: [gwex]` charges a catchment `gwex` to the aquifer.**

My suggestion to reuse `closure`'s `sources` list was too loose. In AGENTS.md, `gwex` is anything that crosses the catchment boundary, and a withdrawal counts wherever it is taken from. On this probe, every `gwex` is now added to the `gw` budget (`probe.yaml:53`). On the 5 gate seeds, with `reference_exchange_exact` and a 0.05 mm/day term:

- An outflow taken from `gw` and reported as `gwex` passes, as it should.
- The same outflow left unreported fails, as it should.
- A withdrawal taken from the channel and reported as `gwex` fails `groundwater_balance`, although the model is correct. `gw` does not change, so the probe sees 0.05 mm/day it cannot explain.

No archived model is affected today, but this is exactly the kind of catchment model #40 asks for. One way to fix it, though the naming is up to you:

- Add an aquifer-only boundary flux to `FLUX_VARS`, `UNITS` and the AGENTS.md table, for example `gw_boundary`, positive into the aquifer, for GHB, WEL and regional exchange acting on `gw`.
- Default `sources` to `[]` (`groundwater.py:36`) and set `sources: [gw_boundary]` in `probe.yaml`.
- Update the probe README and the `groundwater_balance` docstring, which now describe `gwex` in two different ways.

If you prefer to keep `gwex`, please state in AGENTS.md and the probe README that on this probe `gwex` must act on `gw` alone.

**Smaller fixes**

- `groundwater_balance` has no unit tests, so this round's main fix is not pinned. Deleting the cumulative check keeps pytest and the gate green, and so does ignoring `sources`. Please add tests that use the probe's parameters:
  - exact output passes;
  - a 10%-of-recharge leak fails on the cumulative check;
  - an alternating ±0.5 mm error per step fails on the per-step check;
  - a shifted `gw` hides none of the faults in the table above;
  - whatever you choose for item 3 behaves as documented.
- The new `abs_tol: 1.0e-4` is not pinned either. Reverting it to 1e-6 keeps every test green. Please add the two-reach table from my last review: a second reach that gains and loses on the same step, written at `%.6f`. It should pass with the probe's parameters and fail with `abs_tol=1e-6`.
- `groundwater_balance` still reports `value = |Σ residual| / Σ recharge` against `threshold = rel_tol` (`groundwater.py:71-72`), but the pass rule has changed. An alternating ±0.5 mm error fails with a value near zero. A run can also pass with a value above 0.01 whenever `2 × floor` sets the cumulative allowance. Please report the threshold that was actually applied, for example `cumulative_allowed / Σ recharge`.
- The probe README (`:71`), `models/reference_exchange_exact/model.yaml:10` and its adapter docstring (`ht_adapter.py:14`) still call the exact reference "physical". The README table lists it as `exact`, and the same files now call it a bookkeeping reference.
- Both reference manifests say `aquifer_storage_coefficient` is declared "only because the probe's case provides it". It is declared because `probe.yaml` requires it, and without it the must_pass would be INCOMPATIBLE. No criterion reads that value. Dropping it from `requires.static` would be cleaner, but it changes every N/A detail on this probe. Either choice is fine once the wording is right.

I have not started CI on 207d8dd, since `ht validate` already fails there. I will approve the run on the corrected merge.

yaji-wang:
Thanks @chrimerss for the prompt review. I just pushed `09161d9` on top of a redone merge. Here is the update:

**Merge:** Reset to `b7d6c8a`, redid it against `upstream/main` keeping main's side of every conflict and adding this probe's content alongside it. Probe counts restate 29 everywhere `test_docs_in_sync.py` checks, `flex_lumped` has its 1.1.0 N/A row, `modflow6` has rows on the five newly-merged probes. `ht validate`, `pytest -q`, and `ht gate` are all green.

**Store-scaled floor:** Capped at `min(1e-5 * max|gw|, 0.05)`. Your four leak/jump cases fail again; honest output still clears full precision, 6 and 8 sig figs. README now asks for `gw` at full precision or at least 8 significant figures.

**`sources: [gwex]`:** Added `gw_boundary` as an aquifer-scoped flux, defaulted `sources` to `[]`, pointed `probe.yaml` at `gw_boundary`. Your three scenarios all come out right, including the channel-withdrawal case.

**Smaller fixes:** `groundwater_balance` and `exchange_components` both have unit tests now, built off the probe's own declared parameters, each checked to actually fail against the behavior it pins. Two-reach precision case is in at `%.6f`. `threshold` reports the cumulative allowance actually applied. `reference_exchange_exact` says "exact" not "physical"; both manifests explain `aquifer_storage_coefficient` correctly now.

**#94:** Moved `exchange_components` into `groundwater.py`, dropped this PR's `exchange.py`.

Let me know if anything still looks off.

cehw (commented):
Thanks, Yaji. The `gw_boundary` split is the right shape for Zhi's third point. It gives an aquifer-scoped boundary term a home, so the balance no longer has to guess which store a catchment-level `gwex` left. I ran his three scenarios at `09161d9` and they come out as he asked: a withdrawal from `gw` reported as `gw_boundary` passes, unreported it fails at 0.05 mm per step, and one taken from the channel and reported as `gwex` passes. His four datum-shift cases fail now too, the 1e6 mm datum with a 0.0199 mm/day leak drifting 14.53 mm against 0.14 allowed.

One thing I would raise before merge. `AGENTS.md` describes `gw_boundary` as a regional groundwater exchange, unlike `gwex`, and the probe README says a model reports it rather than `gwex`. But nothing outside this probe reads `gw_boundary`. `closure` defaults its sources to `["gwex"]`, and so do the stress, limits, degeneracy and event-closure criteria.

So take a catchment model that follows the README, with a 0.3 mm/day regional loss from the aquifer and a 0.1 mm/day channel withdrawal. It closes here and fails `mass/catchment-closure` at 13.3% of the rainfall against a 5% limit. It passes both only if the aquifer-crossing part sits in both columns, `gwex` at −0.4 and `gw_boundary` at −0.3, which double counts nothing since neither criterion reads the other's variable.

To me the variable looks right and the trouble is in the word "unlike". Would a component rule work, where `gw_boundary` is every flux across the aquifer's own boundary and whatever part of it also crosses the catchment boundary stays inside `gwex` too? A pumping well returning water to soil inside the catchment would then be `gw_boundary` and not `gwex`, which is worth its own clause. No archived row changes today, since no model emits it yet.

Worth settling with #94 while both are open. Its `exchange.py` calls the token-cheat escape the price of not splitting the head-driven part of `gwex` into its own variable, which is what `gw_boundary` does for the aquifer. A trial merge gives five conflicts, all adjacent added lines, so the wording looks like the only real decision.

Two small ones. The probe README line 81 still calls `reference_exchange_exact` the physical baseline. And the ask for 8 significant figures does not hold at a 1e6 mm datum, where honest output fails.

Validation at this head: `ht validate` 29 probes, 1059 tests, this probe's gate and the full gate passed, and native modflow6 passed on the five gate seeds and the two declared-input variants. I did not build the Docker image.

PR #82, opened 2026-09-14
[PROBE: momentum] Test geomorphic routing-lag consistency

### LINKED ISSUE #8: [PROBE] momentum/routing-lag-consistency
### Conservation law

momentum

### Residual equation

τ = t_peak(Q) - t_peak(P)  [hours] 
where τ is the peak lag time (time difference between the rainfall centroid and the runoff peak).
τ_observed / τ_GIUH ∈ [0.5, 2.0]
where τ_GIUH is the theoretical routing time estimated from catchment geomorphological parameters (area, slope, drainage network length) using the Geomorphologic Instantaneous Unit Hydrograph (GIUH) or hydraulic geometry scaling laws

The tolerance for τ is a relative error of ±50% (i.e., τ_observed / τ_GIUH ∈ [0.5, 2.0]). When τ_GIUH < 1 hour, an absolute tolerance of ±0.5 hour is applied instead.

An unphysical model could "cheat" in the following ways:

Zero‑lag response: The model produces an instantaneous runoff response to a rainfall event (τ ≈ 0), completely ignoring routing time. This is physically impossible – water takes time to travel from hillslopes to the catchment outlet.

Anti‑physical scaling: The model yields shorter lag times for larger catchments than for smaller ones, violating the positive monotonic relationship between routing time and catchment area / drainage network length.

Corresponding checks: lag_time_bounds checks whether the lag time falls within the physically plausible range predicted by GIUH; scaling_monotonicity checks whether the lag time is monotonically increasing with catchment area.

The generator constructs a set of synthetic catchments with varying geomorphological characteristics:

Generate 3–5 synthetic catchments with areas ranging from 10 km² to 10,000 km², and set slope, drainage length, etc. according to physical scaling laws.

For each catchment, generate an identical design storm (e.g., 50 mm over 3 hours).

Run the model on each catchment and extract the peak lag time τ.

Test whether τ increases monotonically with catchment area and whether its magnitude falls within the GIUH‑based plausible range.

All cases are generated on‑the‑fly with random seeds to prevent the model from memorising specific examples.

### Will you build it?

Yes — assign it to me

binbinlan:
Closes #8

## What this probe asserts

The same isolated one-day storm and byte-identical weather are applied to four
synthetic catchments with areas of 30, 300, 3,000 and 10,000 km². Main-channel
length follows Hack's relation and centroid channel length is half of the main
length. The observable rainfall-to-total-runoff peak lag must:

1. fall inside a broad factor-of-two envelope around the duration-corrected
   Snyder lag, with a half-day allowance for daily output; and
2. avoid a material reversal as geometry grows, with at least two days of lag
   growth across the full ladder.

The probe requires `pr`, `area_km2`, `main_channel_length_km` and
`centroid_channel_length_km`, and scores only the reported `mrro`. Snyder is
strictly an excess-rainfall unit-hydrograph relation; here it supplies an
explicit synthetic comparison scale for the observable gross-rainfall to
total-runoff lag. The probe therefore cannot isolate channel routing from
runoff generation and storage, a limit documented in the probe README.

All nine currently evaluated model adapters are N/A (INCOMPATIBLE), because
none declares consumption of both channel-length fields. Some do read area,
but area alone does not expose the travel path changed by this experiment. In
particular, the current Google Flood Forecast adapter reads `area_km2` only
after neural-network inference to convert runoff depth to discharge, so it
cannot use that value to change peak timing. The adapters are not run for
these N/A rows.

## Discrimination

| --- | --- | --- |
| `reference_snyder_router` | PASS | n/a; both criteria pass |
| `reference_instant_router` | FAIL | `lag_time_bounds` |
| `reference_inverse_router` | FAIL | `scaling_monotonicity` |

Across the three deterministic gate seeds, the positive reference measures
lags of 0, 1, 2 and 3 days against duration-corrected Snyder expectations of
0.631, 1.123, 2.250 and 3.335 days. The instantaneous control measures zero at
every scale. The inverse control uses individually in-bounds lags of 1, 2, 1
and 3 days, so only the scale reversal rejects it.

## Validation

- `ht validate`: passed; 24 probes and 44 model manifests discovered.
- `ht gate --probe momentum/routing-lag-consistency`: passed on all three gate
  seeds.
- Focused probe, reporting, documentation and line-ending tests: 226 passed.
- Full repository test suite: 845 passed, 8 skipped.
- `models/result.csv`: nine generated N/A (INCOMPATIBLE) rows; no Docker model
  was started.
- `docs/writing-a-probe.md`: documents the existing domain-specific-reference
  pattern for probes whose required inputs cannot be consumed by the standard
  physical references.

## Checklist

- [x] There is an `accepted` proposal issue and this PR closes it.
- [x] `authors` in `probe.yaml` carries the paper-ready name, affiliation and
  ORCID and is synchronized with `CONTRIBUTORS.md` and `CITATION.cff`.
- [x] `ht validate` passes.
- [x] `ht gate --probe momentum/routing-lag-consistency` passes.
- [x] The generator is deterministic given a seed and commits no generated
  data.
- [x] The reference gate runs in under a minute.
- [x] The tolerance bands, response threshold and daily discretization
  allowance are justified in the probe README and manifest.
- [x] Conservative deliberately wrong controls are caught by the named
  criteria.


chrimerss (changes requested):
Thanks for the probe in momentum balance and it is much needed. One note on your probe, can you enable at least one physical model for testing this and testify the physical model can pass? Feel free to include a hydrodynamic model here.

cehw:
Thanks for the clear geometry experiment. I reproduced the `Ct = 2` failure at `5d14cab` on all three gate seeds and found a timing detail worth fixing before changing the tolerances.

[Snyder lag is measured from the excess-rainfall centroid to the runoff peak](https://www.hec.usace.army.mil/confluence/hmsdocs/hmstrm/transform/snyder-unit-hydrograph-model). In `reference_snyder_router`, the fixed runoff fraction preserves that centroid, but the triangle's peak is placed `corrected_lag` after rainfall starts. The criterion measures from the rainfall centroid instead. For this one-day storm, those origins differ by 12 hours.

Keeping `Ct = 2` and the criteria unchanged, I set the triangle's peak to `corrected_lag + 12 hours` from rainfall onset. The daily lags changed from `[0, 0, 1, 1]` to `[0, 1, 1, 2]` days, and both criteria passed on all three seeds.

Could we align the reference timing and add this regression before reassessing the tolerance bands? These were synthetic-reference runs; they do not replace the physical-model validation Zhi requested.

binbinlan:
@cehw @chrimerss Thank you for the detailed timing clarification and for requesting a physical-model validation.

I have incorporated the rainfall-centroid correction. Snyder lag is now consistently interpreted from the excess-rainfall centroid to the runoff peak. In the reference router, the triangular hydrograph peak is therefore placed at the duration-corrected Snyder lag plus the event-centroid offset. For the one-day design storm, this offset is 12 hours. A new `Ct = 2` regression produces daily lags `[0, 1, 1, 2]` on all three gate seeds and passes the existing criteria.

I have also enabled the registered physical model `flex_lumped` as a `must_pass` control. FLEX retains its interception, nonlinear soil partitioning, fast reservoir and slow reservoir processes. Its native triangular routing lag is configured independently from the Snyder criterion using

`channel travel time = Lc / c0`

with `c0 = 1.0 m s-1`, followed by

`Tlag = 2 × (channel travel time + dt/2)`.

Here, the half-step aligns the generated-runoff interval centroid with FLEX's row-indexed routing kernel, and the factor of two converts the desired triangular mode into FLEX's full triangle-base parameter. No Snyder coefficient, length exponent or duration correction is shared with the physical model.

The celerity is documented as a first-order synthetic prior rather than a site-calibrated constant. The value is supported by Beven (2020), which distinguishes mean water velocity from kinematic-wave celerity and uses `1.0 m s-1` for an upland-channel example. The README also records the limitations of extrapolating this value across the synthetic catchment ladder.

Across all three registered gate seeds, the complete FLEX model produces daily lags `[0, 1, 2, 3]` days for the small-to-xlarge variants. The corresponding Snyder expectations are `[0.631, 1.123, 2.250, 3.335]` days. FLEX passes both `lag_time_bounds` and `scaling_monotonicity`, with a three-day end-to-end span above the unchanged two-day minimum.

Validation completed successfully:

- routing-lag gate passed for both `reference_snyder_router` and `flex_lumped`;
- FLEX adapter contract passed;
- full suite: 858 passed, 8 skipped;
- final focused and documentation checks: 203 passed.

The Snyder tolerance ratios and the two-day scaling threshold were not relaxed.

@chrimerss Could you please review the updated implementation when convenient?

PR #14, opened 2026-09-09
[PROBE: mass] Detect multi-decadal storage drift under repeated forcing

hiter-joe:
Adds `mass/multi-decadal-drift`. Closes #13.

Fifty scored years of a repeated five-year weather block, after five years of
spinup: 20,075 daily rows from a seeded generator, with no committed dataset.

## What it asserts

A budget can close exactly while a model hides a persistent runoff reporting
error in invented storage. Two criteria catch that:

- `state_bounds`, which holds soil and canopy inside the case's capacities;
- `total_storage_drift`, new here. The total of every reported store, `gw` and
  `channel` included, may change by no more than
  `2e-4 * P_final_block + 1e-6 mm` across the final repeated block. That
  allows slow equilibration and catches a steady bias above about
  0.0004 mm/day in any store.

## Controls

- `reference_slow_drift` hides a 0.012 mm/day runoff deficit in soil storage.
  It fails `state_bounds` and `total_storage_drift`.
- `reference_gw_slow_drift` parks a 0.001 mm/day deficit in groundwater, where
  no capacity is imposed. It fails `total_storage_drift` alone.

## Results

- The four physical baselines pass every criterion on all five gate seeds.
- `wflow_sbm`, `cwatm` and `lisflood` PASS. `dhbv2` fails `state_bounds` only,
  on the learned field capacity it already fails elsewhere. `summa` fails on
  negative ET and a zero runoff ratio, and passes the drift check at
  -0.449 mm against a 0.658 mm allowance.
- Archive rows, standings, the model cards, the site in three languages,
  ROADMAP, CITATION and CONTRIBUTORS are updated with it.

## Validation

- `ht validate`: 28 probes, 60 models.
- `ht gate`: passed over all 28 probes.
- `pytest`: 1018 passed.
- CI on 870976a: gate and container green.

PR #83, opened 2026-09-15
momentum/routing-conservation: land the dry-spell generator and parameterise the absolute allowance

### LINKED ISSUE #71: [PROBE] momentum/channel-routing-mass (recession-limb)
# Proposal: `momentum/channel-routing-mass`

**Authors**
```
Yuanhang Liu, Independent Researcher, Changchun, China, ORCID 0009-0005-9633-6004
```

**Conservation law**
momentum

**Residual equation**

For a single reach, the channel is a control volume. We test conservation on the *recession limb* of each storm, i.e. windows where external inflow into the channel is effectively zero. During those windows:

```
outflow + Δ(channel) = 0
```

where
- `outflow` is the discharge leaving the reach, reported as `mrro` (mm/day),
- `channel` is the reach storage, reported as the `channel` state (mm),
- `Δ(channel)` is the storage change over the step (mm/day when divided by the step length).

Equivalently, integrated over any contiguous dry window where inflow is zero:

```
Σ(mrro) = channel[start] − channel[end]
```

Dry windows are selected by requiring `mrro` to be above the dry threshold for at least one day before the window, and then requiring the next N days to stay below a small runoff fraction of the preceding peak (e.g. < 5 %). Storms with a baseflow recession are more discriminating than purely zero-flow periods because a non-zero `mrro` must be balanced by a falling `channel` store.

**Tolerance and its denominator**

Use the suite’s 5 % engineering rule. The denominator is the total outflow leaving the reach over the scoring window. Because the window is chosen on the recession limb, `Σ(mrro)` is not zero, but for very short dry spells we add an absolute floor of 0.1 mm over the window:

```
|Σ(mrro) − (channel[start] − channel[end])| ≤ 0.05 × Σ(mrro) + 0.1 mm
```

If the scoring window is shorter than 3 days or Σ(mrro) is below 1 mm the window is degenerate and the probe moves to the next seed.

**How would an unphysical model fail this?**

`routing_conservation` already catches a channel store that is negative or larger than its hydrograph can hold. `channel_routing_mass` catches the complementary failure: a store whose *change* is inconsistent with what left the reach when no new water was entering. A model can pass the bounds test while quietly leaking or gaining water between the channel store and `mrro`.

The discriminating reference model is `reference_leaky_router`: it routes runoff through the same lag kernel as the honest `reference_bucket`-with-router, but loses a fixed fraction (e.g. 5 %) of the water at every routing step. Its `channel` store stays within the bounds set by `routing_conservation`, yet on recession limbs the drop in `channel` does not match `Σ(mrro)`. The leak is invisible to whole-catchment closure because the water is subtracted from the channel store, not from any catchment store.

**How is the case generated?**

Synthetic daily weather over three years with a spinup year and a cluster of large storms in the second year, reusing the generator shape from `momentum/routing-conservation`. Submitted models receive the same `pr/tas/pet` forcing and must report `channel` and `mrro`. No additional `q_in` diagnostic or forcing column is introduced; the test only inspects the relationship between two variables the baselines already emit.

**Will you build it?**

Yes — assign it to me.

kawh1111:
This PR no longer adds a probe. Following the review's third round it changes `momentum/routing-conservation` instead: the generator from the dropped probe is now that probe's weather, and the criterion's constant absolute term is a parameter.

## The evidence the review rests on, reproduced

The criterion's floor is `rise_fraction` (1e-4) **of the store's own maximum**, so it reads as a fraction of the reach:

| Store | Floor | A bounded inflow of 6e-5 mm/step clears it? |
| --- | --- | --- |
| `reference_unreported_inflow` (holds nothing but the fault, max 0.05 mm) | 5e-6 mm | yes |
| a router that holds routed water (max 3.85 mm on this case) | 3.9e-4 mm | no |

That is the review's point stated as arithmetic, and it is why the criterion was the wrong shape: with the store otherwise empty it separates the two criteria, and with routed water in it, it does not.

## What the generator buys, measured

`routing_conservation` allows `max_lag_days` times the largest runoff of the preceding month, read over a rolling 30-day window. On the old weather the window always held a storm. On this one the peak decays out of it during the drains:

| Kernel | Residue (`0.001 x the runoff so far`) | Storm-time ceiling | Old weather | This weather |
| sums to 0.999 | 0.80 mm | 433–635 mm | passes | **fails, 3/3 gate seeds** |

0.80 mm is five hundred times inside what a peak-reading bound allows. It fails here because the ceiling comes down to the runoff behind it — which is exactly the case the new weather exists for, and `tests/test_routing_conservation_generator.py` pins it.

## The allowance, chosen by measurement

`min_allowance_mm` defaults to 1e-6 mm, which is what the fixed term was, so every other probe keeps its behaviour. This probe sets **0.05 mm**:

- `wflow_sbm`, the model the term exists for, needs **0.0026 mm** through the droughts — nineteen times of room. (Your estimate of 0.0016–0.0101 mm was in the right range; the exact need is below it.)
- the models that route sit at **0.01 to 0.58 of the whole allowance** on the gate seeds — `summa` 0.01, `dhbv2` 0.04, `flex_topo` 0.11, `lisflood` 0.12, `wflow_sbm` and `sacsma_snow17` 0.16, `cwatm` 0.47, `flex_lumped` 0.58. The tightest leaves 1.7x.
- the smallest fault this probe must catch, the 0.999 kernel, reaches **4.13x to 4.40x**
  the whole allowance on the gate seeds. It escapes the three gate seeds at 0.615, 0.627
  and 0.651 mm, and a must-fail counts as caught while any seed still fails, so the gate
  turns red once `min_allowance_mm` passes **0.651 mm**, 13.0x the shipped value.

## Dropped, added, moved

- Dropped: `recession_drainage`, `reference_unreported_inflow`, `probes/momentum/channel-routing-mass`, `tests/test_drainage.py`, `tests/test_channel_routing_generator.py`.
- The ROADMAP entry for `momentum/channel-routing-mass` stays **unclaimed**, and now carries the note on why the residual cannot be formed against today's contract. `Closes #71` becomes **`Refs #71`**.
- Added: `reference_leaky_router`, the 0.999 kernel as a model, pinned in `must_fail` so
  the fault the allowance is calibrated against is behind the acceptance gate rather than
  only behind a unit test, and `tests/test_routing_conservation_generator.py` (the weather's
  properties, and the 0.999 kernel failing on it).
- The probe is `version: 2`, the description and README are rewritten for the weather with a Limitations section, and Yuanhang Liu is added to `authors`, `CONTRIBUTORS.md` and `CITATION.cff`.
- `models/result.csv` keeps main's order and **appends** the re-archived rows: every model that had a row on this probe was re-run on the new case — `reference_bucket`, `flex_lumped`, `flex_topo`, `sacsma_snow17`, `dhbv2`, `cwatm`, `summa`, `lisflood`, `wflow_sbm` all pass, `google_flood_forecast` and the three reference models stay N/A. No verdict changed.

## Verification

```
ht validate                   29 probes, 64 models
ht gate --probe momentum/routing-conservation
  must_pass  reference_bucket / flex_lumped / flex_topo / sacsma_snow17
  must_fail  reference_stuck_router  -> tripped ['routing_conservation']
  must_fail  reference_leaky_router  -> tripped ['routing_conservation']
full suite                    1085 passed, 8 skipped
tests/test_docs_in_sync.py    203 passed
tests/test_routing_conservation_generator.py  9 passed
```

The head branch is still named `probe/momentum-channel-routing-mass`, which no longer describes it; rename it on merge, or say the word and I will push to a differently named branch.

Refs #71. The probe that issue proposed is withdrawn — its residual cannot be formed against today's contract, which is what the third review showed — and what lands instead is the weather the issue's generator produced, inside an existing probe. #71 stays open for you to dispose of; nothing in this branch closes it.


kawh1111:
Thanks — the generator finding is the important one, and it was right.

**The generator now does what the probe says it does.** `_weather()` was still drawing wet days at `p_wet` ≈ 0.25 underneath the storms and `_recessions()` was dead code, so the spells were chance runs of four dry days: 330–413 scored steps out of 1095, and `1095 × 0.75⁴ ≈ 346` is exactly what chance gives. `generate()` now applies the mask and zeroes the rain over every post-storm spell; the scored recessions go from about 350 steps to about 850. `tests/test_channel_routing_generator.py` pins the property rather than the storm table — the spells carry no rain, the longest rainless run is longer than the weather can produce, every storm still arrives at full depth, and at least 500 steps are in recession on every gate seed — which is what you asked for and which would have caught the original defect.

**`reference_leaky_router` is gone.** You were right that it was `reference_stuck_router` with different kernel weights, and I verified the consequence rather than assuming it: `reference_stuck_router` does trip `recession_drainage` on the gate seeds. Its `spec.py` allowlist entry and both doc rows went with it, and `must_fail` now names `reference_stuck_router` — one baseline answering both questions, the way `reference_leaky` serves four mass probes on three criteria.

**`requires: fluxes: [mrro]`** — taken, matching `momentum/routing-conservation`. No archived row moves, as you predicted.

**On the lagged inflow:** you were right that it is real, and right that the old generator was hiding it. With the long spells in place the honest side does move: `sacsma_snow17` rises on 8.7% of its recession steps (it was 5.7% on the chance dryness) and `summa` on 1.5%. Both are well inside `max_rising_fraction` of 0.25, so I kept `settle_days` at 3 — raising it would trade scored sample for a tail the threshold already absorbs — and did the other thing you offered instead: the probe README now explains that a slow hillslope pathway keeps delivering water after the rain stops and that the threshold is what absorbs it, and both the README and the criterion docstring now quote the submitted-model shares, so the margin is read off the evidence in this PR rather than off the four reference baselines.

**`settle_days` is a duration now**, converted through the window's `dt_days`; on this daily probe the number is unchanged, and an hourly probe reusing the criterion would get three days rather than three hours. `settle_steps` stays a step count and its comment says so.

**The stale `FAIL,ERROR` summa row is gone.** The rebuild was needed anyway — the case changed, so every row for this probe was stale — and the file is 445 rows with exactly one row per (model, probe) for this probe, so the standing no longer depends on row order.

Gate: 4 must_pass + 1 must_fail separated. `ht validate`: 26 probes, 52 models (`reference_leaky_router` was the 53rd). Full suite: **886 passed, 8 skipped**.

chrimerss (changes requested):
Thanks, @kawh1111. 7b2f11c answers all six points from the automated review, and the mechanics hold up. I checked it locally:

- `ht validate` (26 probes, 52 models), all 894 tests and the probe's gate pass.
- All nine scored archive rows reproduce on their recorded seeds, the five Docker rows included.
- On a trial merge with main at 466ff7b, the gate still passes. `flex_lumped` 1.1.0 gives the same three messages.
- Over 203 seeds, the honest models stay well inside the limit. The bucket and both FLEX models sit at 0, and `sacsma_snow17` at 0.02 to 0.12.

The problem is what the criterion adds. I could not find a routing fault that `recession_drainage` catches and `routing_conservation` misses.

**1. The probe needs a fault only it can catch.**

Each router below is `reference_stuck_router` with only the routing changed. I ran each on 203 seeds (the three gate seeds plus 200 random ones), in-process through `generate()` and `evaluate_criteria`. `recession_drainage` was scored on this probe, and `routing_conservation` on `momentum/routing-conservation`.

| Routing fault | `recession_drainage` | `routing_conservation` |
| --- | --- | --- |
| kernel sums to 0.9, remainder held in `channel` (`reference_stuck_router`) | fails 203/203, share 0.57 to 0.67 | fails 203/203 |
| kernel sums to 0.97 | fails 203/203, lowest share 0.30 | fails 203/203 |
| kernel sums to 0.99 | **passes 203/203**, highest share 0.14 | fails 203/203 |
| kernel sums to 0.999 | passes 203/203 | fails 6/203 (203/203 on this probe's weather) |
| kernel sums to 1.1, so the reach invents water | **passes 203/203** | fails 203/203 |
| 10% of surface runoff held, baseflow routed exactly | **passes 203/203** | fails 200/203 |
| kernel sums to 0.5 on every fifth day, 1.0 otherwise | **passes 203/203**, share 0.16 to 0.17 | fails 203/203 |
| kernel sums to 0.9, remainder destroyed rather than held | passes 203/203 | passes 203/203 |

A leak on baseflow alone, or only while runoff is under 1 mm/day, fails both. The destroyed-water router also passes `closure`. At `mass/catchment-closure`'s settings on this probe's weather, it sits at 3.3% to 3.5% of the rain against a 5% limit (two seeds).

On every fault I built, `routing_conservation` is at least as strict. `recession_drainage` needs a leak of about 3% to fail. `routing_conservation` fails a 1% leak on its own weather, and a 0.1% leak on this probe's weather.

Please add a `must_fail` that passes `routing_conservation` on both generators and fails `recession_drainage`. If no such fault exists, I suggest a different shape for this work. The long-drought generator is the valuable part. It could become a second case of `momentum/routing-conservation`, where the existing bound catches a 0.1% leak on every seed.

**2. Claims to correct, whichever way item 1 goes.**

- A router that "loses or invents water" rises on recession steps (`drainage.py:27-29`, `probe.yaml:20-22`, ROADMAP, the site in three languages). Routers that invent water or destroy it both pass.
- The error "shows up long before any bound is reached" (ROADMAP), and a leaking router "stays inside those bounds for a long time" (probe README:16-18). The table shows the bound is the more sensitive test.
- `reference_stuck_router` "sits above 0.85" (probe README:39, `drainage.py:74-77`). On the new generator it sits at 0.57 to 0.67. "About three times below the failing side" (README:44-45) is stale as well.
- The probe README (lines 61-69) explains SAC-SMA's rises as a slow hillslope still delivering water. That does not fit. On seeds 548877356 and 1205417207, its rising steps fall a median of 140 and 160 days into a dry spell. Of those steps, 93% and 98% land on a day evaporative demand drops, against 50% of all recession steps. That points to riparian evaporation drawn from channel inflow. So "no water enters on a rainless step" does not hold for an honest model either, and the README should say what actually moves the store.

**3. State the departure from #71.**

The accepted proposal was a residual: runoff leaving the reach on a recession against the fall in `channel`, within 5%. This PR ships a count of how often `channel` rises. It also rewrites the ROADMAP line "Inflow minus outflow minus the change in channel storage, per reach".

I understand why. The contract has no inflow to the reach, and `dis` is `mrro` times area. Please state the departure and its reason in the PR description and the probe README. Keep the original ROADMAP line until we agree on what this probe claims.

**4. Rebase onto main.**

#82 merged as 466ff7b. This branch now conflicts in `README.md`, `ROADMAP.md`, `docs/writing-a-probe.md`, `models/result.csv` and `site/index.html`. After the rebase:

- README:71 "Twenty-seven: seventeen under mass, six under energy and four under momentum."
- README:79 "Twelve of the twenty-seven require no model output beyond runoff."
- README:84 "the seven more we want"
- README:377 "Twenty-seven probes, seventeen mass, six energy, four momentum"
- `flex_lumped` 22 of 22, and `flex_topo` and `sacsma_snow17` 21 of 21. These go in the README table, the site's three languages and the SAC-SMA card.
- Re-archive the `flex_lumped` row at 1.1.0. Only the version changes.
- The site flowchart needs both new momentum dots.
- The PR description still mentions `reference_leaky_router`, 882 tests and 53 models.

Item 1 decides whether this probe merges in its current form. Items 2 to 4 are needed either way.

kawh1111:
Thanks — this is the review that mattered, and you were right that my earlier answer was not enough. **There is a fault of that shape, and it is now a shipped baseline.**

## 1. A fault only this criterion catches: `reference_unreported_inflow`

Your table is right that every fault in it is caught at least as strictly by the bound. What none of them has is a fault that does not **accumulate**.

`reference_unreported_inflow` routes exactly — what is generated in a step is released in that step, so the whole-catchment budget closes to the floating point — and on each rainless step it adds a small fixed amount to the channel store out of a reservoir it never reports:

```
channel += min(0.05 mm - channel, 6e-5 mm)      on a step it did not rain
```

Over 23 seeds (the three gate seeds plus 20 random ones), in-process:

| | `recession_drainage` | `routing_conservation`, this probe's weather | `routing_conservation`, its own weather |
| --- | --- | --- | --- |
| verdict | fails **23/23** | passes **0/23** failures | passes **0/23** failures |
| share / ratio to the bound | share 0.48–0.52 | ratio **0.42–0.62** | ratio 0.08–0.56 |

The bound is `max_lag_days` times the largest runoff of the preceding window, and on this probe's long dry spells that peak decays with the store's own recession, so the ceiling shrinks as the spell goes on — the store is measured against a bound that the weather is collapsing. That is why the inflow must be small: at 0.5 mm the bound does catch it (ratio 4.2–6.2), at 0.05 mm it does not, and the direction test does not care either way, because it reads the sign of the change rather than its size.

So the probe now declares both baselines, and the gate separates all six models:

```
momentum/channel-routing-mass  (seeds [1205417207, 1712369321, 71837788])
  ✅  must_pass  reference_bucket             recession_drainage=pass; non_degenerate=pass
  ✅  must_pass  flex_lumped                  recession_drainage=pass; non_degenerate=pass
  ✅  must_pass  flex_topo                    recession_drainage=pass; non_degenerate=pass
  ✅  must_pass  sacsma_snow17                recession_drainage=pass; non_degenerate=pass
  ✅  must_fail  reference_stuck_router       expected `recession_drainage`, tripped ['recession_drainage']
  ✅  must_fail  reference_unreported_inflow  expected `recession_drainage`, tripped ['recession_drainage']
```

And the separation the review asked for, run through the CLI rather than in-process:

```
$ ht run --model reference_unreported_inflow --probe momentum/routing-conservation
  ✅ PASS  momentum/routing-conservation
      routing_conservation: the channel store stays non-negative and within 0.02 of
      what a 15-day hydrograph can hold
```

If you would rather have the fault be one a router could plausibly commit by accident, the shape you suggested — a kernel summing to 1.1 — is *caught by the bound* (it makes the store walk away from its hydrograph, which is exactly what the bound is for); I could not get it past `routing_conservation` on either generator. The bounded inflow is what "the reach receives water from a store the model never reports" looks like in a real adapter, and I think it is a defect rather than a contrivance — but if you disagree, say so and I will take the other path you offered.

## 2. The claims are corrected

All four, in every file you named.

- "loses or invents water" → the criterion docstring, `probe.yaml`, ROADMAP and the site in three languages now say what is actually tested: a reach that receives water it never generated, of any size, applied where nothing enters to mask it. A router that *destroys* water passes, as you found, and the text no longer implies otherwise.
- "shows up long before any bound is reached" (ROADMAP) and "stays inside those bounds for a long time" (probe README) are gone. Both are replaced by the measured statement: a leak that accumulates is caught by both criteria, and an inflow that does not accumulate is caught only here.
- "sits above 0.85" is replaced by the measured 0.57 to 0.67 over the gate seeds, in the probe README and in the criterion's comment, with the margin now stated as "about three times the honest share and about a third of the failing one".
- The SAC-SMA explanation was wrong and I have taken yours. The probe README and the criterion comment now say that 93% to 98% of its rising steps land on a day the evaporative demand drops, against half of all recession steps, and that this points to riparian evaporation drawn from channel inflow — attributed to your analysis, since I took it from your comment rather than re-deriving it.

## 3. The departure is stated

In the probe README under "Where this departs from the accepted proposal", and in this PR's description: #71 asked for a residual against the fall in `channel`, the contract has no variable for water arriving at the reach, and `dis` is `mrro` times area — so such a residual would compare a model against itself and pass by construction. What ships is the one-sided version of the same physics, which is weaker as a statement and stronger as a test.

**The ROADMAP entry keeps its original line** — "Inflow minus outflow minus the change in channel storage, per reach" — with the departure noted beneath it rather than by rewriting what the probe claims.

## 4. Rebased onto main

Merged `main` at `e5c87e2` (which is past the `466ff7b` you tested); conflicts resolved in `README.md`, `ROADMAP.md`, `docs/writing-a-probe.md`, `models/result.csv` and `site/index.html`. With the counts rebased:

- 28 probes: seventeen mass, seven energy, four momentum.
- `flex_lumped` **22 of 22**, `flex_topo` and `sacsma_snow17` **21 of 21**, in the README table, the site in three languages and the SAC-SMA card.
- The `flex_lumped` archive row is re-run at 1.1.0; only the version changed, and the superseded 1.0.0 row for this probe is gone.

`ht validate`: 28 probes, 59 models. Gate: six models separated. Full suite: **1005 passed, 8 skipped**.

kawh1111:
Brought the branch up to date with `main` at `57d1d41` — `#14` (multi-decadal drift) merged after your review, which took the PR to a conflicting state.

Both sides had rewritten the same count lines, each describing a suite missing the other's probe:

- `models/result.csv` takes the **union**: both sides' rows are true of the merged suite, and the archive is what the docs are checked against.
- `README.md` and `site/index.html` keep this branch's wording and are re-counted from the merged archive, in all three languages. Every model scored on either new probe gains one from its denominator: `dhbv2` **12 of 21**, `wflow_sbm` **18 of 21**, `summa` **9 of 24**, `cwatm` **17 of 21**, `lisflood` **18 of 21**, `flex_lumped` **23 of 23**, `flex_topo` and `sacsma_snow17` **22 of 22**. The suite total is **twenty-nine probes: eighteen mass, seven energy, four momentum**, in the three places the README states it.
- The model cards are re-counted with them, including SUMMA's narrative chain (`momentum/channel-routing-mass` adds one scored PASS, bringing the current count to 9 of 24), which keeps the older counts as history rather than overwriting them.

Verified on the merge: `ht validate` **29 probes, 61 models**; `ht gate --probe momentum/channel-routing-mass` separates all six models (`must_pass` `reference_bucket`, `flex_lumped`, `flex_topo`, `sacsma_snow17`; `must_fail` `reference_stuck_router`, `reference_unreported_inflow`); full suite **1031 passed, 8 skipped**.

Nothing else changed — the separation baseline and the corrected claims from my previous message are untouched by the merge.

chrimerss (dismissed):
Thanks, @kawh1111. cc4f01d brings the branch up to date with 57d1d41, and the merge checks out:

- `ht validate` (29 probes, 61 models), all 1039 tests and the probe's gate pass. CI is green.
- The recount matches main in the README, the site and the model cards.

I have now tested `reference_unreported_inflow` the same way as the first table. My recommendation is to move your generator into `momentum/routing-conservation`, rather than merge `recession_drainage` as a probe of its own.

**1. The new baseline separates the two criteria only because its channel is otherwise empty.**

Its `channel` holds nothing but the fault. It starts at zero and gains 6e-5 mm on each rainless step, up to 0.05 mm. A rise has to clear 1e-4 of the store's maximum, so the floor here is 5e-6 mm. I added the same fault to models that actually route, over 203 seeds:

| Same fault added to | `recession_drainage` | `routing_conservation`, this weather / its own |
| --- | --- | --- |
| the bucket, as shipped | fails 203/203 | passes / passes |
| the bucket, at twice the rate | **passes 203/203** | passes / passes |
| an exact 3-day kernel (weights 0.5, 0.3, 0.2) | **passes 203/203** | passes / passes |
| `flex_lumped` | **passes 203/203** | passes / passes |
| `sacsma_snow17` | **passes 203/203** | fails 181/203 / fails 19/203 |

In a channel that holds routed water, the floor is 1e-3 to 1e-2 mm. The store also drains faster than the fault fills it, so the rise is never counted. At 100 times the rate (6e-3 mm per step, capped at 5 mm), `recession_drainage` catches the fault on 38/203 seeds for the kernel, 171/203 for SAC-SMA and 0/203 for `flex_lumped`. `routing_conservation` fails all three on every seed of this weather. At twice the rate, the shipped baseline reaches its cap early and then stays flat, so it passes.

So what `recession_drainage` adds is a slow creep into a channel that holds no routed water. I don't think that is enough to carry a probe of its own.

**2. Your generator is the valuable part, because it sharpens the bound.**

Through your long dry spells, the bound shrinks with the recession, so a channel that keeps water is caught early. Over the same 203 seeds, `routing_conservation` fails:

| Router | routing-conservation's current weather | your weather |
| --- | --- | --- |
| kernel sums to 0.99 | 203/203 | 203/203 |
| kernel sums to 0.999 | 6/203 | **203/203** |

The four physical baselines pass on your weather on all 203 seeds. I also gated `momentum/routing-conservation` with only its generator swapped for yours, from a private probe root. The gate passes. On its gate seeds, `cwatm`, `dhbv2`, `lisflood` and `summa` pass, and `google_flood_forecast` stays N/A.

`wflow_sbm` fails, and that needs one change before the swap. Through the droughts its river keeps 0.0016 to 0.0101 mm while its runoff falls to 1e-5 mm/day. That is up to seven times what 15 days of that runoff could hold. The criterion's absolute allowance is fixed at 1e-6 mm. With 0.01 mm, `wflow_sbm` passes on all three gate seeds, though with little margin. Every leaking router I tested still fails on the same seeds: kernels summing to 0.9, 0.99, 0.999 and 1.1, the surface-runoff leak and the every-fifth-day leak. Please make the allowance a parameter with 1e-6 mm as its default, and pick a value with some margin.

**What I suggest**

- Replace `probes/momentum/routing-conservation/generate.py` with yours, set the absolute allowance, and bump the probe's `version`.
- Point `tests/test_channel_routing_generator.py` at routing-conservation.
- Add a test that a kernel summing to 0.999 fails on the new weather. It is the case the new generator exists to catch.
- Drop `recession_drainage`, `reference_unreported_inflow` and `probes/momentum/channel-routing-mass`.
- Keep the ROADMAP entry for `momentum/channel-routing-mass` unclaimed, with your note on why the residual cannot be formed today. Change "Closes #71" to "Refs #71".
- Re-archive `momentum/routing-conservation` for every model, since its case changes.
- Update routing-conservation's README and `description` for the new weather, and add yourself to its `authors`.
- Keep `models/result.csv` in main's order and append the new rows. cc4f01d sorts the whole file by date and model, which turns a 10-row addition into a 223/213-line diff. The harness appends, and the docs tests read the last row per model and probe.

If you would rather keep `recession_drainage` as its own probe, it needs a must-fail that routes water and still passes the bound. Its claims would also need to say how narrow the check is. The current text says it sees an error "of any size".

xinlan-technology (changes requested):
Thanks @kawh1111, and thanks @chrimerss for the two rounds. I agree with the round-two recommendation to move the generator into `momentum/routing-conservation`; what follows adds to it.

Your two findings reproduce here. The shipped creep added to an exact 0.5/0.3/0.2 kernel, or to `flex_lumped`, gives a rising share of 0.0000 on every seed I tried. On the gate seeds (1205417207, 1712369321, 71837788) the baseline fails at 47.7%, 50.1% and 49.1%; with only `INFLOW_MM_PER_DAY` doubled it passes at 5.4%, 5.0% and 7.4%, and at 1.42 times the shipped rate (8.5e-5) all three already pass.

What I am asking for is the round-two route, plus the corrections in point 5 for whatever text survives. The six points below are not six separate requests. Points 1-4 have not been raised yet and apply only if `recession_drainage` is kept in some form; if the round-two route is taken they fall away with the criterion. Point 1 is the one I would call a defect; 2-4 are limits worth stating or testing. Point 6 asks for nothing in this PR.

1. **One sample controls the floor.** A single `inf` or `-inf` in a post-spinup row of `channel` passes `protocol.read_result` (NaN is rejected there, inf is not), makes `store_max` and the floor infinite, and `recession_drainage` returns PASS with a share of 0.0. `non_degenerate` never reads `channel`, so `run_probe` gives PASS/OK for both must-fail baselines on all three gate seeds. `routing_conservation` still fails those runs. A finiteness check alone would not close this: a finite 1 mm spike on the first scored row, which `settle_steps` excludes from scoring, also turns `reference_unreported_inflow` into a PASS on all three seeds, and `routing_conservation` passes it too. I would check the store and the computed floor for finiteness, take the scale from something a single sample cannot set, and add a regression test that goes through the real result-reading path.

   Minor: `rise_fraction: .nan` or `.inf` and `max_rising_fraction: .inf` load through the schema and pass both baselines. Only the probe author can write them and the gate would then report the probe as broken, but a `ValueError` as in `radiative_identity` or `event_water_closure` would be cheap.

2. **The must-fail verdict depends on how much of the scored window falls before the cap.** Each share is the fraction of recession steps before the 0.05 mm cap: it is reached at scored step 612, 645 and 627 of 1095, and the store never rises on a recession step after it. About a third of the fill (0.0174 mm on the first seed) already happens in the 365-day spinup. With the first spinup year repeated in front of the unchanged record (730-day spinup, identical scored weather) the unchanged baseline passes at 0.154, 0.180 and 0.174. So the control's separation rests on the spinup length, the rate and the cap, and does not yet show that the criterion detects this kind of fault in general.

3. **The floor does not scale with the time step.** This does not affect the current gate, which runs at PT1D only; it matters if the criterion is reused on an hourly case. The baseline declares PT1H and scales its inflow by dt. Handing it the same weather at PT1H (daily rates repeated 24 times, a case built by hand and scored in-process) it passes on every seed with a share of 0.0000: the per-step rise of 2.5e-6 mm is under the 5e-6 mm floor. Dividing `rise_fraction` by 24 restores the failure. A sentence in the docstring, or a test, would be enough.

4. **`tests/test_drainage.py` pins less than its docstring says.** Six mutants I tried survive its ten tests, the gate and the full suite: an absolute floor, a floor relative to the mean store, no dt conversion of `settle_days`, scoring the spinup rows instead of excluding them, a one-step misalignment between storage changes and the recession mask, and a default threshold of 0.45. This shows the tests do not fix these behaviours, not that the implementation has them wrong. The tests never exercise non-default spinup or timestep values.

5. **Wording.** The probe README (lines 23-29) says the baseline "closes its budget to the floating point", that "No whole-catchment budget can see this", and that `closure` closes because the error lives inside the channel; "to the floating point" is repeated in `ht_adapter.py`, `model.yaml`, the `drainage.py` docstring and the PR description. `closure` sums every reported store, `channel` included, so it does register this fault: called with default parameters on the gate seeds it reports -0.0326, -0.0332 and -0.0330 mm (about 2e-5 of the rain, 6e-5 mm on roughly 550 of 1095 steps), exactly the water added to the channel in the scored window, while `reference_bucket` closes to 4e-13 mm through the same runner. The fault is capped at 0.05 mm. Its scored-window residual is about 0.033 mm, so `closure` passes on these gate seeds because the residual is far below the 5 percent tolerance, not because it is invisible to the budget. README line 28 also says the water is "subtracted from the reach"; in this baseline it is added.

   Separately, README lines 67-75 now explain that rainless does not mean nothing moves the store, but the `probe.yaml` description (lines 17-19), the criterion docstring (lines 12-18) and the FAIL message (lines 187-190) still say no water enters. And the largest honest share on the gate seeds is 10.9% (`sacsma_snow17`), so 0.25 is 2.3 times it rather than about three times.

6. **A possible follow-up, not a change to this PR.** I agree the ROADMAP id should stay unclaimed with "Refs #71". #71's residual is taken on inflow-free recession windows with the outflow as denominator, and the contract indeed gives no way to know the reach inflow there. A different experiment is within reach of what main already has: `closure` accepts `denominator: sum_inflow`, which reads a forcing column `q_in` (as #71's proposal review noted), and since #39 a probe can require models to declare that they consume it through `requires.forcing`. For a reach-only control volume that gives `[sum(q_in) - sum(mrro) - d(channel)] / sum(q_in)`. On 10 seeds with a reach-only router it is complementary to the bound: destroying the last tenth fails it 10/10 (+0.1000) and passes the bound; holding it back does the reverse.

   The conditions are real, though. `closure` sums every reported store, not only `channel`, so a full-catchment model needs the control volume set up explicitly. The denominator is zero on a truly inflow-free window, which `closure` reports as degenerate, so this is a prescribed-inflow test rather than #71's recession test. No generator supplies `q_in` and no manifest declares it, so every model would be N/A today. And an adapter that reports `channel` as cumulative inflow minus outflow closes by construction. If useful, the ROADMAP note could name this as the route a later probe might take.

kawh1111:
Done — the generator has moved into `momentum/routing-conservation` and the probe is gone. Your evidence holds, including the part I had been arguing against.

**Your point, as arithmetic.** `recession_drainage`'s floor is `rise_fraction` (1e-4) *of the store's own maximum*. On `reference_unreported_inflow` the store holds nothing but the fault (max 0.05 mm), so the floor is 5e-6 mm; on a router that holds routed water (3.85 mm on this case) it is 3.9e-4 mm, and the store's own drainage is orders of magnitude larger than any inflow small enough to stay inside the bound. So the criterion separated the two only while the channel it watched was empty. I could not construct the alternative you left open — a must-fail that routes water, trips the direction test, and still passes the bound — and I no longer think one exists for a store whose decline is its own physics.

**What the generator buys, measured.** A kernel summing to 0.999 leaves a residue of `0.001 x the runoff so far`: **0.80 mm** on the gate seeds against a storm-time ceiling of **433 to 635 mm**. It is five hundred times inside the bound wherever the peak is read from a storm, and outside it once the peak has decayed out of the 30-day window. It fails **3/3 gate seeds** here; on the old weather it passed. That is now pinned by `tests/test_routing_conservation_generator.py`, along with the weather's properties (spell length, the storms, determinism).

**The allowance.** `min_allowance_mm` is a parameter with 1e-6 as its default, so nothing else in the suite moves. This probe sets **0.05 mm**, and I measured rather than guessed:

- `wflow_sbm` needs **0.0026 mm** through the droughts — nineteen times of room. Your 0.0016–0.0101 mm range bracketed it; the exact need sits below it, so 0.01 mm would have worked but with less margin than your estimate suggested.
- the models that route sit at **0.01 to 0.58 of the whole allowance** over the gate seeds (`summa` 0.01, `dhbv2` 0.04, `flex_topo` 0.11, `lisflood` 0.12, `wflow_sbm` and `sacsma_snow17` 0.16, `cwatm` 0.47, `flex_lumped` 0.58) — tightest 1.7x of room;
- every leaking router you named still fails: kernels 0.9, 0.99, 0.999 and 1.1, the surface-runoff leak, and the every-fifth-day leak. The smallest is caught at 4.4x the allowance.

**Also done:** probe `version: 2` with the description and README rewritten for the new weather (and a Limitations section stating what the bound cannot see — the peak having to decay, timing not being read, a small loss not being read directly); `Yuanhang Liu` added to `authors`/`CONTRIBUTORS.md`/`CITATION.cff`; `recession_drainage`, `reference_unreported_inflow`, `probes/momentum/channel-routing-mass` and their tests dropped; the ROADMAP entry left **unclaimed** with the note on why the residual cannot be formed; `Closes #71` -> `Refs #71`; `models/result.csv` **keeps main's order and appends** (my earlier sort is gone — you were right that it turned a ten-row addition into a two-hundred-line diff). Every model that had a row on this probe was re-run on the new case and **no verdict changed**.

```
ht validate      28 probes, 60 models
ht gate --probe momentum/routing-conservation   separates all five baselines
full suite       green
```

One loose end: the head branch is still named `probe/momentum-channel-routing-mass`, which no longer describes what it does. Rename it on merge, or tell me the name you want and I will push a fresh branch.

kawh1111:
Thanks @xinlan-technology — this is a better review than the change deserved, and it landed on a head that no longer exists. I took the round-two route, so most of what you found went with the criterion it was about. Here is what falls away, what I carried across, and the one general thing your point 1 turned up.

## Points 1-4: gone with `recession_drainage`

The criterion, its baseline `reference_unreported_inflow`, the probe and their tests are all deleted, so the `inf` floor, the cap-before-window dependency, the per-step floor, and the mutant-surviving test file are all out of the tree. I did not try to salvage any of it: your point 4 was the fair one — a test file that six mutants walk through is not doing what its docstring says, and rather than defend it I withdrew it.

**Point 3's lesson did carry.** The surviving bound takes its scale from the recent peak runoff, not from the store, and both terms are millimetres: the peak is a rate, `lookback_days` is converted through `w.dt_days`. So it is step-independent, and the probe README now says so explicitly, precisely because a per-step floor was the defect you found.

**Point 1's parameter half carried too.** `min_allowance_mm: .inf` would pass every model and `.nan` compares false against everything, so both would read as satisfied — same shape as `rise_fraction` and `max_rising_fraction`. The criterion now raises `ValueError` on a non-finite or out-of-range parameter, the way `radiative_identity` and `event_water_closure` do.

**Your point 1 generalises further than this PR.** A non-finite value in the spinup is outside *every* criterion's window, because `make_window` drops the spinup before any criterion sees the frame. I hit it while writing the regression test for the surviving bound: an `inf` placed at row 0 passed, and the same `inf` placed inside the scored window failed. That is not specific to a store-relative floor, and it may be worth its own issue if you want it chased.

## Point 5: the text is gone, and I accept the correction

The README paragraphs, the adapter comment, `model.yaml`, the `drainage.py` docstring and the old PR description all went with the probe, so nothing in the tree still says "closes to the floating point", "no whole-catchment budget can see this", or that the water is subtracted from the reach.

And your correction is the one that settles the whole episode: `closure` **does** register that fault (−0.0326 mm, −0.0332 mm, −0.0330 mm on the gate seeds), and it passed only because that residual sits far inside its 5 % tolerance. So the probe's claim that no budget can see it was wrong, and with the bound catching the same fault at least as early there was no gap for it to fill. Round two's recommendation was right and my earlier defence of the probe was wrong on both counts.

## Point 6: taken, and credited

The ROADMAP entry for `momentum/channel-routing-mass` stays **unclaimed** with `Refs #71`, and now names your route — `closure` with `denominator: sum_inflow` over a `q_in` forcing column, reach-only control volume — with the conditions stated alongside it: the denominator is zero on a truly inflow-free window so it is a prescribed-inflow test, `closure` sums every reported store, no generator produces `q_in` and no manifest declares it, and an adapter reporting `channel` as cumulative inflow minus outflow closes by construction. I attributed it to this review rather than presenting it as my own.

## The head you reviewed is superseded

The probe is gone and its generator now runs `momentum/routing-conservation` as `version: 2` (the kernel summing to 0.999 fails on all three gate seeds there; it passed on the old weather). The criterion's constant `1e-6` is now the `min_allowance_mm` parameter, defaulted to what it was, with this probe setting 0.05 mm by measurement: `wflow_sbm` needs 0.0026 mm through the droughts, the models that route sit at 0.01 to 0.58 of the allowance over the gate seeds, and the smallest fault the probe must catch is caught at 4.4x. The archive was re-run for every model that had a row and no verdict changed.

Since the merge now needs two reviewers, could I ask you to look at the new head rather than the old one? `cc4f01d` is not what would merge.

xinlan-technology:
Thanks @kawh1111 — I reviewed `1d1d08a`, not the old head. The route you took is the right one, and the probe earns its change: with an explicit release queue I independently confirm that a consistently constructed 0.999 kernel fails on all three gate seeds of the new weather.

**The three items in the first list are what I still need. Once they are addressed and verified on the updated head, I expect to approve. Nothing in the second list blocks me.**

## Three places where the tree says something that is not so

1. The criterion is not deleted, only unregistered. `src/hydroturing/criteria/drainage.py` is still at the head, still carries `@criterion("recession_drainage")` at line 98, and `git diff cc4f01d HEAD -- src/hydroturing/criteria/drainage.py` is empty — it is byte-identical to the version I reviewed. What `de742da` removed is the `drainage,` line from `criteria/__init__.py`. The runtime is safe: `criteria.get("recession_drainage")` raises `KeyError` and `spec.py` rejects a `probe.yaml` that names it. But the file would merge into main unreferenced and now untested, since `tests/test_drainage.py` went in the same commit. It also still carries the text this PR corrected everywhere else: lines 36-38 say the baseline has "a budget that closes to the floating point", and line 7 says the fault is invisible to a whole-catchment budget. Those are the two claims we agreed were wrong, surviving verbatim in `src/` and pointing at a model this PR deletes. Please delete the file.

2. `tests/test_routing_conservation_generator.py:172-173` says the row is placed inside the scored window "since `make_window` drops the spinup and no criterion sees a row in it". `criteria/base.py:134-139` sets `prior = max(start - 1, 0)` and hands `run.table.iloc[prior]` to every `Window` as `state0`, so the last spinup row does reach every criterion that reads it — an `inf` in that row can turn a passing `closure` into a failure. Handing the prior row over is the right design; only the sentence needs correcting, and it is the same generalisation you made in your reply. If you would rather not carry the harness question here, I am happy to open an issue for it.

3. The 0.999 test does not build a 0.999 kernel. Lines 127-129 add `0.001 * mrro.cumsum()` to `channel` and leave `mrro` untouched, so what is scored is a store that gains water from nowhere rather than a router that retains a tenth of a percent. The conclusion stands — I independently confirmed that a consistently constructed 0.999 kernel fails on all three gate seeds — but this test is the evidence for the whole change, so I would rather it ran the fault it names: build the routed outflow and the retained store together, and cover all three gate seeds instead of `gate_seeds(PROBE, 3)[0]` alone.

## Six smaller things, none of them blocking

4. "caught at 4.4x" (`README.md:69`, `probe.yaml:64`) is not the binding seed. I measure 4.445, 4.218 and 4.128 on the three gate seeds, and a second construction gives a range bottoming at 4.13. Every seed has to pass, so the figure that governs is about 4.13x. Worth saying which statistic the number is.

5. `probe.yaml:87-89` says this probe's case is not the one the rows dated before 2026-09-17 were run on and that "those rows are history". `de742da` adds eight `routing-conservation` rows dated 2026-09-16, including the `reference_bucket`, `flex_lumped`, `flex_topo` and `sacsma_snow17` rows the README's own ratio table cites. A date is not a clean version boundary here; the note and the rows should agree.

6. The storms are pinned by nothing. `test_the_spells_follow_storms_rather_than_dry_weather` asserts only `forcing["pr"].max() >= 30.0` over the whole record, which the background weather clears on its own — on gate seed 924646966 the wettest day is 79.81 mm both with the storms and without them. Deleting the storm injection outright survives the new tests, the gate and the full suite, because the dry mask is applied whether or not the storm rain was added. Check that each intended storm supplies substantial rainfall and is followed by the intended dry interval. `p_wet` 0.25 -> 0.50 and doubling the peak window in `routing_conservation` also survive all three.

7. The 0.0026 mm measurement is documented, in the README and in `probe.yaml`, so my earlier worry that it lived only in a comment was misplaced. Please also give the seeds, command and calculation used to obtain it — `wflow_sbm` is a container model I cannot run here, so I am taking the figure on trust. A regression test that fails with the default allowance and passes with a specified `min_allowance_mm` would pin the parameter's effect, which nothing currently does.

8. `routing_conservation` does not reject unknown parameter names the way `event_water_closure` does, so `min_allowance` without the `_mm` is accepted and silently replaced by the default. Optional, but cheap.

9. Two model cards still quote v1 numbers for this probe: `models/cwatm/README.md:74` says 0.50 where you now measure 0.47, and `models/wflow_sbm/README.md:293` says 0.17 where you now measure 0.16.

kawh1111:
All three are done, and the six smaller ones with them.

## The three

**1. `drainage.py` is deleted.** You were right that unregistering is not deleting: `criteria/__init__.py` stopped importing it, and the file stayed in the tree byte-identical to the version you read, with `@criterion("recession_drainage")` still in it and the two claims we agreed were wrong still in its prose. It is gone, along with its last reference.

**2. The 0.999 test now runs a 0.999 kernel.** It was adding `0.001 * cumsum(mrro)` to the store and leaving the outflow alone, so it scored a store that invents water. The released series and the retained store are now derived together from the bucket's own runoff, over **all three gate seeds**, and the exact kernel is asserted to pass on the same record. The ratios come out 4.40, 4.17 and **4.13**, so the README and `probe.yaml` now quote the range and say that the smallest governs, since every seed has to pass. That also settles the figure you measured: your 4.128 and my 4.128 are the same seed and the same step.

**3. The spinup sentence is corrected.** You are right about `criteria/base.py:134-139`: the last spinup row is handed to every criterion as `state0`, so a non-finite value there is not invisible — it reaches the criteria that read it, and an `inf` there can fail a passing `closure`. The test docstring now says that instead of the generalisation I made. And yes please, open the issue: the asymmetry is worth its own record — the scored rows of the spinup are dropped while its final row is handed over as an initial state, which is a reasonable design for a state that has to start somewhere and an unreasonable one to reason about from the outside.

## The six

**4.** Fixed as above: 4.13x to 4.40x, and the statistic named.

**5.** The note no longer uses a date as the boundary. It now says the boundary is the probe version, and gives the dates that hold — the rows up to 2026-09-15 are the old weather, those dated 2026-09-16 and 2026-09-17, including the four baselines my ratio table cites, are the new one.

**6.** You were right that the storms were pinned by nothing: the record-wide maximum is cleared by the background weather on its own. The test now checks **each event**: the rainfall over its own window against what the table adds, and the spell that follows it rainless all the way to the next storm. On your evidence I also bounded the record's annual rainfall, since a doubled wet-day frequency survived the same way. Mutation-checked, restoring the generator each time and comparing the file's hash:

```
M1 no storm injection            2 failed  (the storm test, and the 0.999 test)
M2 wet-day frequency 0.25 -> 0.50 2 failed  (the annual-rainfall test, and 0.999)
M3 no dry spells                 4 failed  (spells, storms, rainfall, 0.999)
```

The peak window you mention (`lookback_days` 30 -> 60) still passes every test. I have not pinned it: it is a probe-level choice, and unlike `min_allowance_mm` nothing measured justifies one value over another. If you think it needs pinning too, say so and I will.

**7.** The figure is documented with its seeds, its command and the calculation, in the probe README: `ht run --model wflow_sbm --probe momentum/routing-conservation --gate-seeds --workdir`, gate seeds 417694852, 924646966, 1431599080, largest channel 16.93 / 20.19 / 15.90 mm, `need = max(channel - max_lag_days x peak runoff x 1.05)` = **2.5821e-3 mm** on each seed. And the parameter's effect is now pinned by a test: a store sitting 0.0026 mm above the proportional bound fails with the 1e-6 mm default and passes with this probe's 0.05 mm. That is exactly the regression you asked for.

**8.** `routing_conservation` refuses unknown parameter names now, the way `event_water_closure` does, so `min_allowance` without the `_mm` is reported rather than silently dropped.

**9.** `models/cwatm` reads 0.47 and `models/wflow_sbm` 0.16. The `wflow_sbm` comparison table also carries a note: its four rows are a comparison between settings made on the case before this change, the shipped row measures 0.16 on each of the three seeds on the current one, and the other three settings were not re-run — so the table is not read as current numbers for them.

## Also

I merged `main`, so the branch carries the new two-reviewer rule in `GOVERNANCE.md` and is current; nothing else in that commit touches this PR. It names the momentum pool as Zhi Li and Siddik Barbhuiya, so if the gate needs a pool reviewer who has not been on this PR yet, point me at them and I will ask.

`ht validate` 28 probes / 60 models; `ht gate --probe momentum/routing-conservation` separates all five baselines; full suite green.

kawh1111:
Brought the branch up to date with `main` at `e7edcfd` (`#79`, the exchange probe). Only `models/result.csv` conflicted — both sides appended at the end — and it is resolved as the union, main's rows first and mine after, which is the order the file is written in.

Nothing else moved. `ht validate` 29 probes / 63 models, `ht gate --probe momentum/routing-conservation` separates as declared, full suite 1085 passed / 8 skipped, and the docs tests pass unchanged at 242. The three blocking points from your last review and the six smaller ones are all in `c3d5e2f`, which this merge does not touch.

Barbhuiya12 (commented):
Ran the sweeps on `pr83`. One finding I think is blocking, one question for @chrimerss that is above this PR, and one thing that is right and worth recording.

## The gate does not protect the calibration this PR is about

`must_fail` pins `reference_stuck_router`, and over 103 seeds it is caught at a `max_ratio` of **348** at its weakest, against an honest worst of 0.577. It is three orders of magnitude clear of the allowance, so it is insensitive to `min_allowance_mm` — which is the one number this PR exists to choose.

The fault that actually governs that choice is the 0.999 kernel, and it lives only in `tests/test_routing_conservation_generator.py`. Holding everything else fixed and varying the allowance on the three gate seeds:

```
  min_allowance_mm   reference_stuck_router   0.999 kernel
             0.05            FAIL                 FAIL
             0.50            FAIL                 FAIL
             1.00            FAIL                 pass
             2.00            FAIL                 pass
             5.00            FAIL                 pass
```

At 1.0 mm — twenty times the value you chose, and argued for at length — `ht gate` is still green while the leak the new weather exists to catch walks through. `docs/writing-a-probe.md` puts it as "`must_fail` pins which criterion does the catching, so a probe cannot appear to work while catching things for the wrong reason", and that is the state here: the gate proves the criterion can catch *a* router fault, not that the allowance is where the PR says it is.

The fix looks like the one the guide already prescribes — "if your probe needs a broken model that does not exist yet, add it under `models/`". A `reference_leaky_router` carrying the 0.999 kernel, pinned in `must_fail` on `routing_conservation`, would put the governing case behind the gate rather than behind a unit test. You are already dropping `reference_unreported_inflow`, so the model count does not grow.

A unit test is not a substitute here for the reason the gate exists: it passes on the author's machine, whereas the gate runs on every probe pull request, and the calibration is exactly the thing a later change would move.

## Changing a merged probe: which version records it

GOVERNANCE says a change to a merged probe's tolerance "requires the probe's authors and the steering committee, because it silently changes every recorded result. Any such change bumps the suite minor version." This is more than a tolerance — the case is different weather — and the probe goes to `version: 2`, but the suite stays `0.1.0`.

The concrete consequence is in the archive. `flex_lumped` now has three rows on this probe, all at `suite_version 0.1.0`, all on the same three gate seeds:

```
2026-09-07 ... 0.1.0 ... momentum/routing-conservation,PASS,OK ... all criteria pass
2026-09-15 ... 0.1.0 ... momentum/routing-conservation,PASS,OK ... routing_conservation: t...
2026-09-16 ... 0.1.0 ... momentum/routing-conservation,PASS,OK ... routing_conservation: t...
```

The first is the old case and the others the new one, and nothing in the row says so — there is no probe-version column, so `run_date` is the only discriminator and it is not a version. "Scores are only comparable within a suite version" is the sentence that the archive is now at odds with.

I do not think that is yours to resolve. The steering committee is still being formed, so the process GOVERNANCE prescribes cannot be followed as written, and whether this lands as 0.1.0 or 0.2.0 is @chrimerss's call rather than something to solve inside this branch. Flagging it because the PR changes a merged probe and that is the rule that covers it.

## The calibration is seed-stable, which I checked because I got this wrong elsewhere

`min_allowance_mm: 0.05` is argued from three gate seeds, and I wanted to know whether a wider draw moves it. Over seeds 0-99 plus the gate seeds:

```
  model             gate worst   0-99 worst   at seed
  flex_lumped            0.575        0.577        51
  flex_topo              0.112        0.115        36
  sacsma_snow17          0.158        0.159        56
  reference_bucket       0.000        0.000         -
```

No failures, and the binding model moves by two thousandths across a hundred draws. That is a real property of this case rather than luck: the quantity is set by the drought structure the generator builds every time, not by a rare draw.

I checked it because I made the opposite mistake on #73 last week — I calibrated a design change on three gate seeds, reported it clean, and @chrimerss found a `must_pass` model failing at 60% on seed 171 of a 200-seed sweep. Three seeds cannot see a one-percent failure rate. Yours holds up, and the 19x headroom on `wflow_sbm` is measured against the model the term exists for rather than assumed, which is the right way round.

## Where that leaves me

Not approving while the gate does not cover the governing fault, since that is the part of this PR a later change would most easily undo. If the 0.999 kernel goes in as a reference model and stays pinned, I will re-run both sweeps and the allowance table above against the final head.

xinlan-technology (dismissed):
@kawh1111 Thanks — the three blocking points from my review are resolved at `154da97`, and the additional fixes address my comments. I also checked the new `reference_leaky_router` baseline: it is caught for the right reason (the store stays positive throughout, so this is the excess branch rather than a non-negativity trip), `ht validate` and the gate are clean, and nothing else in the suite moves. Approving from my side.

Three non-blocking follow-ups:

- Please use the same time interval for the rainfall numerator and denominator in `tests/test_routing_conservation_generator.py`, and distinguish the background weather's draw from the record the generator delivers.
- The allowance sweep in the probe README jumps 0.50 to 1.00; the boundary is at about 0.615 mm, so a row there would show the real margin.
- Please align the PR description's 4.4x with the README's 4.13-4.40x range.

kawh1111:
Thanks @xinlan-technology — all three are done, in `23902be`. One github mechanic to flag first, and it is my doing rather than yours: pushing that commit **dismissed your approval** (this repo drops stale approvals when a new commit lands), so it will need re-giving. Sorry about that.

**The rainfall interval.** You were right that the two sides were not the same quantity. The test summed the whole record and divided by the scored interval, and the 880 mm/yr the description claims belongs to the draw `_weather` makes, not to the record `generate` returns. The two are far apart on this case — **828, 901 and 1017 mm/yr drawn against 575, 564 and 598 mm/yr delivered** — because the spells take out more than the storms add, which is the mechanism rather than a detail. The test now replays the draw, rebuilds the delivered record from it and asserts the reconstruction, so a reordering of `generate` fails loudly instead of quietly comparing two different draws. It then checks the two properties that hold: the draw is temperate at 700 to 1200 mm/yr, and the record is drier than the draw. A doubled wet-day frequency draws 1600 to 1900 mm/yr, so the mutant the docstring names is still caught.

**The sweep's boundary.** I bisected the allowance on the stored runs rather than guessing where it sat: `reference_leaky_router` stops being caught at **0.615 mm** on the tightest gate seed, and at 0.627 and 0.651 on the other two — which is your 0.615 mm. The table now carries rows at 0.60 and 0.65 that bracket it, and states that the shipped 0.05 mm sits **12.3x** under the point at which the gate turns red. For contrast, `reference_stuck_router` is caught until 75.7 mm, some 1500x further out, which is what makes the pair worth pinning separately.

**The ratio.** The PR description now reads **4.13x to 4.40x**, measured on the three gate seeds, the smallest governing.

@Barbhuiya12 — this head is the one to re-run against. The 0.999 kernel is now `reference_leaky_router`, pinned in `must_fail` next to the 0.9 one, reporting the store it actually leaves behind rather than a store that gains water. Your boundary figure reproduces independently, and the allowance table in the probe README now brackets it, so the number the two sweeps support is the number the README states.

Both of you: since a push clears approvals here, I will not touch this branch again unless a review asks for something. If either of you has anything else, it is cheaper to raise it now than after an approval has been given and dropped.

@chrimerss — your two reviews are against `cc4f01d`, which the branch left behind some commits ago, and the PR is still held by that state. It needs either a fresh look at the current head or a dismissal. No rush.

chrimerss:
@Barbhuiya12 @xinlan-technology Please reevaluate and let me know any remaining issues. Otherwise, please approve to move forward.

xinlan-technology (approved):
Rechecked a5869b2. All three follow-ups from my approval are addressed, and the focused tests and routing-conservation gate pass. Re-approving.

Barbhuiya12 (approved):
Re-ran everything against `a5869b2`, as I said I would once the 0.999 kernel was
pinned. On this machine: `ht validate` (29 probes, 64 models), `pytest -q`
(1093 passed), the full `ht gate`, and the probe's own gate:

```
momentum/routing-conservation  (seeds [417694852, 924646966, 1431599080])
  must_pass  reference_bucket        routing_conservation=pass; non_degenerate=pass
  must_pass  flex_lumped             routing_conservation=pass; non_degenerate=pass
  must_pass  flex_topo               routing_conservation=pass; non_degenerate=pass
  must_pass  sacsma_snow17           routing_conservation=pass; non_degenerate=pass
  must_fail  reference_stuck_router  expected `routing_conservation`, tripped ['routing_conservation']
  must_fail  reference_leaky_router  expected `routing_conservation`, tripped ['routing_conservation']
  GATE PASSED
```

## The allowance table, re-derived rather than read

I bisected `min_allowance_mm` independently on each gate seed — to 1e-9 mm,
through the harness rather than by reimplementing the bound — and took the
figures out of the README afterwards to compare. Every one reproduces:

| quantity | README | this run |
| --- | --- | --- |
| `reference_leaky_router` store, as a multiple of the whole allowance | 4.13x to 4.40x | 4.128, 4.171, 4.396 |
| allowance at which the leaky kernel escapes each gate seed | 0.615, 0.627, 0.651 mm | 0.6152, 0.6268, 0.6509 |
| allowance up to which the 0.9 kernel is still caught | 77.5 mm | 77.474 (max of 77.474, 75.654, 75.569) |
| shipped 0.05 mm, under the point the gate turns red | 13.0x | 0.651 / 0.05 = 13.02 |
| the 0.9 kernel's headroom | about 1550x | 77.47 / 0.05 = 1549 |

The two "caught until" figures both take the largest seed rather than the
smallest, which is the right way round and applied consistently to both kernels:
a `must_fail` counts as caught while any one seed still fails, so the gate is
green until the last seed lets go. Worth stating because the natural mistake is
to quote the tightest seed — 0.615 would have given 12.3x rather than 13.0x, and
the README has the correct one.

## The allowance demand, over a wider draw

`min_allowance_mm: 0.05` is argued from three gate seeds, so I re-ran the wider
draw I ran last time — seeds 0-99 plus the gate seeds, scoring
`max(channel / allowed)` for every `must_pass` model. Numbers below 1.0 pass:

```
model                 gate worst   0-99 worst   at seed   failures
reference_bucket           0.000        0.000        -           0
flex_lumped                0.575        0.577       51           0
flex_topo                  0.112        0.115       36           0
sacsma_snow17              0.158        0.159       56           0
```

No failures in 400 runs, and these are the same figures to three decimals as
before the leaky router was pinned — so adding it to `must_fail` did not move
the `must_pass` margins, which is what should happen and is worth having
measured rather than assumed. The binding model is still `flex_lumped` at 0.577,
leaving 1.73x, and it moves by two thousandths across a hundred draws. That is
the drought structure the generator builds every time rather than a lucky seed.

I ran this because I got it wrong on #73 — I calibrated a design change on three
seeds, reported it clean, and @chrimerss found a `must_pass` model failing at
60% on seed 171 of a 200-seed sweep. Three seeds cannot see a one-percent
failure rate. This one holds.

## The three points from my last review

**The 0.999 kernel.** This was the one that kept me from approving: the
allowance was calibrated against a fault that nothing in the gate held in place,
so a later change could have moved `min_allowance_mm` out from under the
calibration and left the gate green. `reference_leaky_router` now sits in
`must_fail`, and the table above is the demonstration — at 0.66 mm the gate goes
red instead of quietly passing. That is the property I wanted and it is now
enforced rather than documented.

**The rainfall interval.** Fixed properly rather than patched. The test now
rebuilds the delivered record from the draw by replaying the generator's own
storms and recessions and asserts `np.array_equal(rebuilt, pr)` before comparing
anything, so a reordering of `generate` fails loudly instead of silently
comparing two different draws — which was the actual defect, not the numbers.
Both figures are then taken over the same scored interval. The 828/901/1017
drawn against 575/564/598 delivered is the mechanism showing through, and the
doubled-wet-day mutant is still caught at 1600-1900 mm/yr.

**The boundary.** Bracketed at 0.60 and 0.65 with the red row at 0.66, and the
per-seed escape points stated rather than a single number.

## Two things you did that I had not asked for

The 203-seed result — that the leaky kernel escapes some seeds from **0.47 mm**,
so an allowance between 0.47 and 0.651 would keep the gate green while letting
the leak through on some seeds — is the honest version of the number and it
weakens the headline margin rather than flattering it. Volunteering that is the
right call, and it is the figure a reader should actually use.

The criterion also gained the guard I would otherwise have raised:
`min_allowance_mm: .inf` passes every model and `.nan` compares false against
everything, so either would have read as satisfied rather than as a mistake. I
hit exactly this on my own probe — an overflow made a demand, a gap and an
allowance all infinite and `inf <= inf` returned PASS. Rejecting unknown
parameter names alongside it closes the other half, where a misspelled
`min_allowance` changes nothing and says nothing.

## The archive question is still open, and is not yours

Repeating this only so it is not lost: `flex_lumped` now carries three rows for
this probe at `suite_version 0.1.0`, the first against the old case and the
others against the new one, with nothing in the row distinguishing them. The
probe goes to `version: 2` but the archive has no probe-version column, so
`run_date` is the only discriminator and it is not a version. GOVERNANCE's
"scores are only comparable within a suite version" is what that sits against.
@chrimerss — this is your call on whether the suite version moves, not something
to solve inside this branch.

## Approving

Everything I asked for is in, every number in the README reproduces here
independently, and the wider sweep is clean. Approving.

Issue #41, opened 2026-09-11
[PROBE] Groundwater–surface water exchange response to river-stage perturbation


yaji-wang:
### Conservation law

mass

### Residual equation

Two otherwise identical simulations are compared: a control case and a perturbed case in which surface-water stage is increased.

The changes in directional groundwater–surface water exchange are

```math
\Delta Q_{\mathrm{SW\rightarrow GW}}
=
Q_{\mathrm{SW\rightarrow GW}}^{\mathrm{high\ stage}}
-
Q_{\mathrm{SW\rightarrow GW}}^{\mathrm{control}}
```

and

```math
\Delta Q_{\mathrm{GW\rightarrow SW}}
=
Q_{\mathrm{GW\rightarrow SW}}^{\mathrm{high\ stage}}
-
Q_{\mathrm{GW\rightarrow SW}}^{\mathrm{control}}.
```

For an actively connected groundwater–surface water system, increasing surface-water stage should shift the exchange toward groundwater. Therefore,

```math
\Delta Q_{\mathrm{SW\rightarrow GW}} \ge 0
```

and

```math
\Delta Q_{\mathrm{GW\rightarrow SW}} \le 0.
```

Equivalently, the net groundwater contribution to surface water should not increase:

```math
\Delta Q_{\mathrm{GW,net}} \le 0.
```

All exchange fluxes have units of water volume per unit time ($L^3T^{-1}$), or equivalent depth per unit time ($LT^{-1}$).

### Tolerance and its denominator

The primary criterion is the direction of the exchange response. Increasing surface-water stage must not produce a substantial increase in groundwater discharge to surface water or a substantial decrease in surface-water leakage to groundwater.

The tolerance is scaled by the gross groundwater–surface water exchange in the control simulation:

```math
D =
Q_{\mathrm{GW\rightarrow SW}}^{\mathrm{control}}
+
Q_{\mathrm{SW\rightarrow GW}}^{\mathrm{control}}.
```

Changes smaller than 5% of this gross exchange are treated as numerical tolerance:

```math
T = 0.05D.
```

A change larger than this tolerance in the physically wrong direction fails the probe. The synthetic perturbation is chosen to produce an exchange response substantially larger than this tolerance in the reference model.

If the control gross exchange approaches zero, an absolute flux tolerance scaled to the synthetic case is used instead.

### How would an unphysical model fail this?

[...]

PR #94, opened 2026-09-15
[PROBE: mass] exchange-response: a declared head-driven exchange must respond to its external head

### LINKED ISSUE #76: [PROBE] mass/exchange-response: a declared head-driven exchange must respond to its external head
### Authors

Songkun Yan, School of Civil Engineering and Environmental Science, University of Oklahoma, Norman, OK, USA, 0009-0008-3820-104X

### Conservation law

mass

### Residual equation

The budget the suite already checks, written without the exchange:

```
R0 = pr - evspsbl - mrro - d(mrso + snw + canopy + gw + channel)/dt   [mm/day]
R  = R0 + gwex
```

`closure` requires `|Σ R Δt| <= 0.05 Σ pr Δt`. A model that closes has `R ≈ 0` and therefore `gwex ≈ −R0` identically — measured on a CREST implementation over ten years, `max |R|` per step is 2.3e-13 mm/day. So the declared flux and the residual it closes are the same number twice, and neither the magnitude of `gwex` nor its timing on its own can separate an honest exchange from an invented one. A discriminator has to use information that is not algebraically fixed by closure.

This probe uses a paired counterfactual on a prescribed driver. The case supplies an external hydraulic head `gwh` (m) as a visible forcing column and runs the same weather three times — as given, raised by 0.5 m, lowered by 0.5 m — with every other column byte-identical. With `G = Σ_t gwex_t Δt` over the scored record, for a model that declares it consumes `gwh`:

```
G(raised)  − G(control)  >  +ε
G(lowered) − G(control)  <  −ε
```

`gwex` is positive into the catchment: raising the external head must bring more water in, lowering it less.

`ε = 1e-8 · max(G_gross, 1 mm)`, with `G_gross = Σ_t |gwex_t| Δt` on the control run. **This is a floating-point tolerance, not a physical threshold.** No physical minimum response is asked for: a general-head boundary with a very small conductance responds to a 0.5 m shift by a very small amount and is entirely physical, so what the gate excludes is a response that is absent or wrongly signed, not one that is small. The denominator cannot approach zero in a way that matters: `G_gross` is floored at 1 mm inside the tolerance, and a model whose gross exchange is under `1e-4` of the record's precipitation is returned `PASS (negligible exchange)` before any response is computed, never an unscored state.

Measured across the five gate seeds: the honest control responds by +3650 mm and −3197 to −3385 mm against ε ≈ 1.7e-5 mm; the adversarial control responds by exactly 0 both ways. There is no calibrated threshold to report because there is no physical threshold.

`gwex` is the contract's confession channel: *declared, it is a source in the budget and the budget can close; hidden, it is the residual* (`AGENTS.md`). Nothing yet says what a declared exchange may look like, so a model can solve for the exchange as whatever balances its budget exactly as `reference_cheater` solves for storage.

**`reference_noise_sink`** — the exact bucket, conserving water internally, which misreports its evaporation by a seeded ±30% and declares the difference as a groundwater exchange. It declares that it consumes `gwh` and never reads it. Its reported budget closes to 0.0000% of precipitation; it passes `closure`, `state_bounds`, `non_degenerate`, `forcing_fidelity`, and every one of the 23 probes currently in the suite. Because the harness holds the model seed fixed across variants and nothing it computes reads `gwh`, its exchange is identical with the head raised and lowered: a response of exactly 0 both ways. Caught by `exchange_response`.

**Why not gate on reversal.** Two earlier drafts of this proposal did, first on its own inside rainless windows, then relative to the prescribed head's turning points. A native MODFLOW 6.7.0 run supplied on this issue showed the first to produce false positives on a physical, mass-conserving aquifer behind an oscillating general-head boundary. The second rested on a claim that between two turning points of the external head `h_ext − h_catchment` crosses zero at most once; that is false whenever the internal head moves (`h_catchment = t + 0.1 sin(10t)` against `h_ext = t`), and in a real aquifer it does. Reversal statistics are therefore reported and never gated.

**Why declaring `gwh` is the condition, and what it means.** `needs_forcing` and `uses_forcing` establish only that a model reads the column, so the contract text makes the declaration a semantic opt-in: `gwh` is the prescribed external hydraulic head associated with the model's declared `gwex`, and a model declaring it asserts that its external exchange responds monotonically to this potential — the semantic of a GHB, `Q = C(H_ext − h)`. Whatever else drives such an exchange, raising the external head cannot make less water come in. A model that does not declare `gwh` is not judged on its response; the response is reported.

**Positive control.** `reference_driven_exchange` follows the prescribed head against a fixed catchment reference head, reverses 360 times over the record and 15 times inside the rainless windows, and answers +0.5 m with exactly `2 mm/m/day × 0.5 m × 3650 d`. It must pass, and does on every seed.

**What it does not catch.** A model that reads the head, ignores it for its budget and adds a token head-proportional term passes, as it would on every counterfactual probe in the suite; and a residual sink that declines to declare `gwh` passes with its reversals reported, because the case cannot know what drives that model's boundary.

Synthetic, generated at run time from the recorded seed and variant. Ten scored years of daily weather behind 365 days of spinup, on the closure probe's climate: 843 mm/yr precipitation, 759 mm/yr potential ET, a real seasonal snowpack. Three variants, `control`, `raised`, `lowered`, differ only in `gwh`: a 10 m datum, a 0.15 m annual cycle and a 0.35 m cycle of 20 days, shifted by 0, +0.5 and −0.5 m. Two rainless stretches of 90 and 75 days, labelled `_regime = dry` (stripped before the model sees the forcing), carry the reversal diagnostics. The whole record is scored (`min_window_days` 3650).

### Will you build it?

Yes — assign it to me. Built; the acceptance gate passes on all five gate seeds against `reference_bucket`, `flex_lumped`, `flex_topo`, `sacsma_snow17` and `reference_driven_exchange`, and fails `reference_noise_sink` on `exchange_response`. Contract additions: the optional forcing column `gwh` (documented in `AGENTS.md` beside `abstr`) and an optional `model: ModelManifest | None` on `RunResult` so a criterion can read what the model declared.

Skyan1002:
Closes #76

## What this probe asserts

`gwex` is the contract's confession channel: a model whose catchment gains or loses water across its boundary declares that flux and its budget closes over a term it admits to. Write the budget without it, `R0 = pr − evspsbl − mrro − dS/dt`, so the suite checks `R = R0 + gwex`; a model that closes has `R ≈ 0` and therefore `gwex ≈ −R0` identically (2.3e-13 mm/day on a CREST implementation), so neither the magnitude nor the timing of `gwex` on its own can separate an honest exchange from an invented one. Only a counterfactual can.

The probe prescribes an external hydraulic head `gwh` (m) as a visible forcing column and runs the same weather three times — as given, raised by 0.5 m, lowered by 0.5 m — **from the first scored step on, with the spinup byte-identical across the three.** The probe requires `gwh`; a model that does not declare it in `needs_forcing`/`uses_forcing` is INCOMPATIBLE (N/A), the footing `energy/radiation-consistency` stands on. For a model that declares it, with `G = Σ_t gwex_t Δt` and `G_gross = Σ_t |gwex_t| Δt` on the control:

```
G     = Σ_t gwex_t Δt                  TV(g) = Σ_t |gwex_t − gwex_{t−1}| Δt  (control run)
G(raised)  − G(control)  >=  +max(s · TV(g), g · G_gross, ε)
G(lowered) − G(control)  <=  −max(s · TV(g), g · G_gross, ε)
s = 3e-4,   g = 1e-5,   ε = 1e-8 · max(G_gross, 1 mm)
```

`gwex` is positive into the catchment: raising the external head must bring more water in, lowering it less, **by at least a small share of how much the exchange the model itself declared moves from day to day, and by a much smaller share of how much it moves in total.** A share, not an absolute minimum, so conductance cancels. Of the variation, not the gross: a boundary that also drains recharge has a gross of the order of the throughput, independent of storativity, while its response is `S·Δh`, so a share of the gross failed an exact losing catchment across the whole confined range (2.4e-4 at S = 1 mm/m). A steady throughput adds little to how much the exchange moves, so the variation does not have that defect: the same catchment sits at 6.3e-3 at S = 1 mm/m and 2.6e-3 at S = 0.3. The floor is set three times under the value at S = 0.1 mm/m and three hundred times above the token cheat; the class it still excludes — very small storativity carrying a strongly varying throughput, under about S = 0.03 mm/m here — is named in the README. Reversal statistics are reported and never gated.

## Discrimination

`ht gate --probe mass/exchange-response`, five gate seeds:

| --- | --- | --- |
| `reference_driven_exchange` | PASS | n/a — constant internal head; answers +0.5 m with exactly `2 mm/m/day × 0.5 m × 3650 d = +3650 mm`, resp/TV 6.5 |
| `reference_evolving_exchange` | PASS | n/a — internal head `S dh/dt = C(H − h)`, τ = 5 d; in equilibrium at scoring start, answers with `S·Δh = 5 mm` and little more, resp/TV 1.1e-2. **Fails if the shift is applied during spinup** |
| `reference_recharge_exchange` | PASS | n/a — the evolving boundary on a losing catchment, S = 1 mm/m, half the runoff draining through it: `S·Δh = 0.5 mm` against ~1700 mm of throughput, resp/TV 5.4e-3 to 8.6e-3 on the gate seeds, 3.95e-3 to 1.36e-2 over seeds 0–99 (minimum 13× the floor). **Fails under a share of the gross (2.4e-4)**, and pins the normalisation |
| `reference_noise_sink` | FAIL | `exchange_response` — declares `gwh`, never reads it, budget closes to 0.0000%, response exactly 0 both ways |
| `reference_token_exchange` | FAIL | `exchange_response` — the noise sink plus a head trickle at C = 1e-6: right sign, +0.0018 mm, **1.0e-6 to 1.2e-6 of how much its exchange moves** (floor 242× above) |
| `reference_steady_sink` | FAIL | `exchange_response` — the same trickle over a *steady* sink: its variation is ~0 so its share of it is 7.1, and only its share of the gross, **1.7e-6**, shows it |
| `reference_bucket`, `flex_lumped`, `flex_topo`, `sacsma_snow17` | N/A | do not declare `gwh` → INCOMPATIBLE, as on the radiation probe |

`reference_leaky`, `reference_cheater` and `reference_degenerate` declare no `gwex` and are N/A by design.

## What changed since the third review (at d76e83c)

- **A steady sink defeated the variation floor (@chrimerss).** A share of the variation sets no bar on a *steady* unrelated part: a constant sink adds nothing to the variation, so a token head term of any size clears a share of its own movement. Reproduced exactly — a model overstating evaporation by a constant 0.3 mm/day, 13% of the rain, declared as steady inflow, passed 25 of 25 seeds while answering the head with 1.7e-6 of its gross, the same share as `reference_token_exchange`. The bar is now the larger of two shares, the second of the gross at `1e-5`: two orders under the variation share, it binds only where that one has already given up, changes no honest control's verdict anywhere in the storativity sweep, and fails the steady sink on every seed. `reference_steady_sink` is a third `must_fail` control so a later change of denominator cannot reopen it with the gate still green.
- The variation now prints with three significant figures rather than rounding to `0 mm`; `spec.py`'s trusted-model list gives each probe's controls their own comment; the generator's docstring no longer calls the dry windows the scored window and gives the case's real climate; `reference_token_exchange` is described as caught on its variation; `AGENTS.md` carves `gwh` out of the "must be consumed whenever supplied" rule; the README's excluded class is measured over 25 seeds (`S = 0.05` loses one seed, so the band starts a little higher than the gate seeds alone showed).

## What changed since the second review (at b06771f)

- **Normalisation (@cehw's losing-catchment check).** The share floor is now taken against the control exchange's total variation, not its gross. The gross derivation assumed the boundary was the aquifer's only source; on a losing catchment that drains recharge through the same boundary the gross is throughput and the share fell below 1e-3 across the whole confined range. `reference_recharge_exchange` — the evolving boundary plus half the runoff as recharge, S = 1 mm/m — is the new positive control that pins this: 2.4e-4 of the gross, 6.3e-3 of the variation. A storativity sweep from 10 to 0.1 mm/m is in the README, with the class the floor still excludes named.
- **Floor 3e-4**, three times under the sweep's value at S = 0.1 mm/m; over a 105-seed draw per control the tightest honest margin is 13× and the cheat sits 242× under, with nothing on the wrong side on any seed.
- Top-level README counts (29 / 19); the probe README now says what `reference_noise_sink` actually does across the suite — every budget probe passed, failing only this probe and where its evaporation noise shows as noise (`mass/steady-state`, and on some seeds `energy/pet-consistency`); the ROADMAP block sits after `mass/human-abstraction`'s contributor line; `docs/writing-a-probe.md`'s reference table has the five controls.

## What changed since the first review (at 6ccb404)

- **Timing (@cehw's MODFLOW check).** The head shift now begins with the scored record; the spinup is identical across variants and the criterion refuses a case where it is not. A responsive aquifer given the shift during spinup had finished answering before scoring began and failed for responding correctly; with this change the native model gives +5/−5 mm and passes.
- **Second positive control.** `reference_evolving_exchange`, whose internal head moves — the control a constant-head one cannot be. It is the one that would fail under the old timing, and the one the floor is calibrated against.
- **Share floor (the token-term cheat).** A conductance-invariant share, first of the gross at `1e-3`, superseded by the variation floor at `3e-4` in the section above; `reference_token_exchange` is the cheat from the review, kept as a negative control. Disclosed limit: a `gwex` that mixes a small head-driven part with a large unrelated one is indistinguishable from the cheat; the floor is set low to tolerate a mixture of that order.
- **N/A rather than PASS for models that do not declare `gwh`** (`requires.forcing: [gwh]`), following the radiation and snowmelt precedent. This probe scores no model in the current field; all nine standings are unchanged by it, and the six Docker rows are N/A rows the harness writes without running the model.
- **Unit tests** — `tests/test_exchange_response.py`, including one that asserts the gate fires on a hand-built `RunResult` with no manifest, since `RunResult.model` defaults to `None`.
- **Checklist items.** Reference-model rows are in the reference-model table (all four controls; the misplaced row is gone). ROADMAP entry moved beside `mass/human-abstraction`, where it belongs in kind, so "These four probes" counts four again; blank line restored. Calibration re-measured on the probe's actual gate seeds (they changed when the probe was renamed): lowered response −3333 to −3447 mm.

## What this adds beyond the probe directory

- `src/hydroturing/criteria/exchange.py` — the paired `exchange_response` criterion, registered in `criteria/__init__.py`; `tests/test_exchange_response.py`.
- `gwh` — a new optional forcing column, added the way `mass/human-abstraction` added `abstr`; documented in `AGENTS.md` beside it as a semantic opt-in. The `gwex` output contract is unchanged.
- `RunResult.model: ModelManifest | None` (optional, default `None`), set by `Runner.run`. Every existing criterion ignores it; a hand-built `RunResult` is judged, not excused.
- Six reference models, added to `TRUSTED_SUBPROCESS_MODELS`.
- The merge checklist from `AGENTS.md`: README rows, CONTRIBUTORS.md, CITATION.cff, ROADMAP.md, `site/index.html` (three languages and the flowchart), `tests/test_report_detail.py` HEADLINES, and archive rows in `models/result.csv` for all nine evaluated models.

Relation to `mass/gw-sw-exchange-consistency` (#79, merged meanwhile): complementary, not overlapping. #79 checks a single run's exchange bookkeeping — components sum to the net, recharge plus exchange equals Δ`gw`, signs — on the river–aquifer flux *inside* the catchment; its own README calls those checks "bookkeeping, not physics". This probe checks a paired counterfactual on the boundary-crossing `gwex`: whether the exchange answers the driver the model declares. `reference_noise_sink` would pass every #79 criterion, since its accounts are exact; `reference_exchange_sign_error` is N/A here. The earlier #41 proposal had a stage counterfactual; the merged #79 perturbs nothing, so the mechanisms no longer meet. They also fit together on one model: a GHB reported as `gw_boundary` is held to the aquifer balance by #79, and reported as `gwex` with `gwh` declared it is held to its driver here.

## Checklist

- [x] There is an `accepted` proposal issue and this PR closes it
- [x] `authors` in `probe.yaml` names every author with `name`, `affiliation` and `orcid`, matching the proposal issue; CONTRIBUTORS.md, CITATION.cff and the paper's author list are built from it
- [x] `ht validate` passes
- [x] `ht gate --probe mass/exchange-response` passes
- [x] The generator is deterministic given a seed and a variant and commits no data
- [x] The probe runs in under a minute on a two-core runner (reference models take well under a second per variant; `max_runtime_s: 120` per adapter)
- [x] Tolerance and denominator are justified in `probe.yaml` — a share of the exchange's own variation, calibrated by a storativity sweep, and a floating-point ε
- [x] I constructed a model that passes closure without doing physics, and a criterion catches it — three of them; and three exact models that must pass, one of which the previous floor failed
- [x] Archive rows for every evaluated model

https://claude.ai/code/session_01N5ymxQy7bPrGb1GCYk8qLn


chrimerss (changes requested):
Reviewed at `6ccb404`. The physics of this revision is right, and the parts that were argued hardest are the parts that hold up: `gwex ≈ −R0` under closure, so only a counterfactual can discriminate; the reversal gate correctly abandoned after the MODFLOW counter-example; `closure` kept as a precondition because the declared number has to be the water that moved. The generator, the byte-identity check between variants, the availability bound and the flowchart re-spacing all check out.

What I verified locally, on a clean worktree at the head commit:

- `pytest -q` — **999 passed**
- `ht validate` — 28 probes, 60 models
- `ht gate --probe mass/exchange-response` — passes: all five `must_pass` models, and `reference_noise_sink` trips `exchange_response` as declared
- the three subprocess archive rows reproduce exactly, including sacsma_snow17's detail string
- CI green at `6ccb404`; CLEAN against `main` at `e5c87e2`

I am not ready to merge it, for one substantive reason and three checklist ones.

---

## 1. The negative control is defeated by a five-line change

`reference_noise_sink` fails only because its response is *exactly* zero. Keep every bit of the cheat and add a token head term, and it passes:

```python
CONDUCTANCE = 1.0e-6          # mm/day per metre of head
head    = step.get("gwh", 10.0)
trickle = CONDUCTANCE * (head - 10.0) * dt_days
soil   += trickle             # so the budget still closes exactly
...
rows[-1]["gwex"] = (evspsbl - true_et) - removed / dt_days + trickle / dt_days
```

All five gate seeds:

```
✅ PASS  mass/exchange-response
   ✅  closure            cumulative residual 0.0000% of sum_pr
   ✅  exchange_response  the declared exchange answers the prescribed head:
                          +0.00183 mm raised +0.5 m, -0.00183 mm lowered
   ✅  state_bounds, non_degenerate, forcing_fidelity
```

Diagnostics: `gross_mm = 1206.3` (14.3% of the record's precipitation), response `±0.00183 mm`. **1.5e-6 of the declared exchange answers the head; the other 99.9999% is still the day's accounting error.**

The README anticipates this — *"a model that reads the head, ignores it for its budget and adds a token head-proportional term would pass — which is also why no physical minimum is asked for: a minimum would not stop the token term and would fail a genuine boundary of small conductance."*

That is true of an **absolute** minimum and false of a **share** one, and the distinction is the whole fix. For a general-head boundary `Q = C(H − h)`:

```
response = C · Δh · T                      gross = C · Σ|H − h| dt
response / gross = Δh · T / Σ|H − h| dt     — independent of C
```

Both halves scale with conductance, so the ratio does not. A boundary a million times weaker than `reference_driven_exchange` has a response a million times smaller *and a gross exchange a million times smaller*, and its ratio is unchanged. Measured on this probe:

| model | Δ raised (mm) | gross (mm) | ratio |
| --- | --- | --- | --- |
| `reference_driven_exchange` | +3650 | 1702 | **2.1** |
| the token-term cheat above | +0.00183 | 1206 | **1.5e-6** |

Six orders of magnitude apart. A floor at, say, `1e-3 · gross` passes the positive control with a 2000× margin, fails the token term by 700×, and still lets a genuinely weak boundary through untouched. As it stands the gate reads "does the model reference the column at all", not "is the declared exchange head-driven", and the model it was built to catch needs one line to escape.

I am open to being argued out of this — it is your probe and the limitation is disclosed rather than hidden — but the disclosure rests on an argument about absolute minima that does not carry over to the conductance-invariant ratio, so I would like the ratio considered on its merits before this lands.

## 2. Merge checklist item 1: the reference-model rows (`README.md`)

`AGENTS.md:293` asks for *"a row in the reference-models table for every reference model the probe adds"*. At `6ccb404`:

- `reference_driven_exchange`'s row is in the **probes** table (`README.md:117`), immediately before `## Models`
- `reference_noise_sink` has **no row at all**

No test catches either — `test_docs_in_sync.py` checks probe rows and standings counts, not reference-model rows. (Both were raised in the automated review at `24d466d` and are unchanged.)

## 3. `ROADMAP.md:137` — "These four probes" now introduces five

The Generalisation section holds exchange-response, extreme-event-closure, ungauged-basin-closure, precipitation-counterfactual and time-origin-invariance. Checklist item 4 asks specifically to *"rewrite the paragraph above it if it counted the entries"*. Same entry: the `*Discriminates:*` line runs straight into the next `###` with no blank line.

Worth reconsidering the placement while you are there — this probe does not move the question to another place or time, it asks whether a declared term answers a prescribed driver, which is closer in kind to `mass/human-abstraction` one section up. Moved there, the count needs no change.

## 4. Stale calibration numbers

The calibration table (`probes/mass/exchange-response/README.md:200`) and the PR body both give `reference_driven_exchange`'s lowered response as **−3197 to −3385 mm**. Across the same five gate seeds I measure **−3331 to −3447**:

```
1541944022  -3377      915316717  -3331
2048896136  -3340     1422268831  -3447
 408364603  -3339
```

`+3650` raised is exact on every seed and matches.

---

## Not blocking, but worth doing

**No unit tests for `criteria/exchange.py`.** 315 lines, and every comparable criterion ships one — `test_rating.py`, `test_routing_lag_consistency.py`, `test_melt_energy.py`, `test_event_water_closure.py`, `test_human_abstraction.py`. It matters more here than usual because `RunResult.model` defaults to `None`, and a hand-built `RunResult` — the repo's normal way to write a criterion test — silently sets `consumes = False` and turns the gate off. Only `Runner.run` supplies the manifest. A test that asserts the gate actually *fires* is worth having precisely because the default is "do not judge".

**The opt-in leaves nothing in the current field judged, and it moves published standings.** No submitted model declares `gwh`, so none is gated. `dhbv2` — the one submitted model whose `gwex` is the closure residual, budget to 1e-8 — escapes the head test by not declaring the driver, and its row fails on the pre-existing `state_bounds` issue instead. Meanwhile a model that does not declare `gwh` gets **PASS**, not N/A, which adds one to both numbers for six models: flex_lumped 21→22, flex_topo 20→21, sacsma_snow17 20→21, wflow_sbm 16→17, cwatm 15→16, lisflood 16→17 — on a probe whose headline criterion did not judge any of them. The precedent from the radiation and snowmelt probes is the other way: *"model does not declare that it consumes forcing rn"* gives N/A. I am not asking for a change here, only for the choice to be deliberate, since it moves numbers on the site and in the README.

**Docker rows.** I did not reproduce the six rows produced by the fork-only workflow. The two N/A rows are consistent with the manifests (`summa` emits no `gwex`; `google_flood_forecast` emits none of `pr, evspsbl, gwex, mrso, snw, canopy`), and `dhbv2`'s `state_bounds` failure matches its behaviour on the other budget probes.

cehw (dismissed):
Thanks for working this through, Songkun. The prescribed-head experiment gives the exchange check a physical driver to test, and the reversing positive control carries the earlier feedback into the implementation. I think this paired design is worth keeping, with one timing adjustment.

The head offset starts during spinup, but the response is measured only afterwards. A responsive aquifer can finish adjusting before scoring begins, so this needs resolving before merge.

I checked this at `6ccb404` with a native MODFLOW 6.7.0 confined aquifer using the probe's head series, alongside the unchanged surface bucket. The raised case admitted 5 mm of extra water during spinup, raising its internal head by 0.5 m. The subsequent exchange difference was nearly zero, so only `exchange_response` failed on all five seeds, even though the aquifer had responded correctly.

Could we keep spinup identical across variants and apply the head offset when scoring begins? With that timing change alone, the native control gave +5/-5 mm and passed all five seeds. Your existing positive and negative controls also retained their expected results. Adding a positive control whose internal head can evolve would help preserve this distinction. I'd settle this before calibrating a response/gross floor, since a moving internal head changes that response too.

Validation at this head: 999 tests and the original probe gate passed. I did not rerun the Docker-only models.

Skyan1002:
@chrimerss Thank you for the review, and for the five-line cheat — it is now `reference_token_exchange` and it is caught. Everything below is at `b06771f` (main merged in after the drift probe landed).

**1. The share.** Taken as you proposed: `min_response_share: 1e-3` of the control run's gross exchange, floored by a floating-point `ε = 1e-8 · max(G_gross, 1 mm)`. Your algebra is right and I had argued the wrong case: an absolute minimum fails a weak boundary, a share does not, because for `Q = C(H − h)` both the response and the gross scale with `C`. I settled the timing first (per the other review) and it changed the numbers in a way worth recording. Against a constant internal head the ratio is `Δh·T / Σ|H − h| dt ≈ 2`, as you measured. Against an internal head that moves, `S dh/dt = C(H − h)`, a shift is answered by `S·Δh` and in the fast limit the gross is `S·TV(H)`, so the ratio tends to `Δh / TV(H) ≈ 2e-3` for this head series — independent of both `C` and `S`. That is the physical minimum for any head-driven boundary, and `1e-3` sits under it rather than being calibrated to a control. The token cheat measures 1.5e-6 on every gate seed, six orders under. What the share does not close is disclosed in the README: a `gwex` that mixes a small head-driven part with a large unrelated one is indistinguishable from the cheat from outside, and the floor is set low to tolerate a mixture of that order.

**2. Reference-model rows.** All four controls are in the reference-model table; the misplaced `reference_driven_exchange` row is out of the probes table. You are right that no test catches this; I did not add one, since the checklist is prose and the drift probe's rows landed the same way.

**3. ROADMAP.** Moved beside `mass/human-abstraction`, as you suggested — it is the same kind of question, a declared term answering a prescribed driver — so "These four probes" counts four again. Blank line restored.

**4. Calibration numbers.** Found the cause: the gate seeds are derived from the probe id, and the id changed when the probe was renamed to match the issue title; the README still carried the numbers from the old seeds. Re-measured on the probe's own seeds: lowered response −3333 to −3447 mm, matching yours; +3650 raised is exact on every seed.

**Tests.** `tests/test_exchange_response.py`, fourteen of them, including one that strips the manifest from a hand-built `RunResult` for both controls and asserts the cheat still fails and the control still passes. Your reading of the default was correct and I have inverted it: with no manifest the criterion judges, and a manifest that reaches it without declaring the driver is a contract error rather than an excuse.

**N/A.** Changed, following the radiation and snowmelt precedent: `requires.forcing: [gwh]`, so a model that does not declare the head is INCOMPATIBLE and its standing is untouched. The probe scores no model in the current field; the seven standings that had moved are back to what the archive says, and `dhbv2` is N/A here and continues to fail `state_bounds` elsewhere. One consequence you may want to know: every one of the nine archive rows is now an N/A row the harness writes without running the model, so the fork-only Docker workflow is no longer part of the story.

Validation at `b06771f`: `pytest` 1039 passed, `ht validate` 29 probes / 64 models, `ht gate` passes on two `must_pass` and two `must_fail`, CI green.

Skyan1002:
@cehw Thank you — this was the one my controls could not have found. A constant internal head is never absorbed, so `reference_driven_exchange` answered with a sustained flux under either timing and the spinup problem was invisible from inside the repository.

Done as you proposed, at `b06771f`. The head shift now begins with the first scored step; the spinup is byte-identical across the three variants, and the criterion checks that and refuses a case where it is not. Your MODFLOW result is reproduced by a new positive control, `reference_evolving_exchange`, `S dh/dt = C(H − h)` with `S = 10 mm/m`, `C = 2 mm/m/day`, a five-day time constant: in equilibrium with the prescribed head when scoring begins, it answers +0.5 m with **+5 mm** and −0.5 m with **−5 mm** on all five gate seeds — `S·Δh`, the number your native run gave — and it fails under the old timing exactly as your aquifer did. The existing positive and negative controls keep their results.

I also followed your ordering on the floor. With the timing settled, the response/gross ratio of the evolving control is 3.4e-3, not the 2 of the constant-head control, and in the fast limit the ratio tends to `Δh / TV(H) ≈ 2e-3` for this head series — both the response (`S·Δh`) and the gross (`S·TV(H)`) scale with storativity, neither with conductance. The share floor is set at `1e-3`, under that minimum, so no head-driven boundary can fall beneath it; the token-term cheat from the other review sits at 1.5e-6.

Validation at `b06771f`: 1039 tests, the gate on two `must_pass` and two `must_fail`, CI green. The Docker-only models are all N/A rows now that the probe requires `gwh`, so nothing there needed re-running.

cehw (commented):
Thanks, Songkun. The evolving-head control is the one I most wanted to see. A model whose internal head never moves answers a shift forever, so it could never have shown the spinup problem, and this one reproduces the MODFLOW number from inside the repository. That settles the timing issue I raised at `6ccb404`.

I then checked the new share floor with the same native setup. Your fast-limit argument holds: at a 0.1-day time constant the gross came to 2555.0 mm against `S·TV(H)` of 2554.9 mm, and the share to 1.957e-3, so conductance drops out.

Storativity does not, once the boundary also carries throughput. I let half the bucket's runoff recharge the aquifer and drain out through the same boundary. The response is still exactly `S·Δh`, but the gross is now the throughput, about 1700 mm, so the share falls with storativity. At 1e-3, the top of the usual confined range, it is 2.8e-4 and fails on all five seeds, while every other criterion passes and the response stays monotone in the head.

The README discloses a `gwex` that mixes a head-driven part with an unrelated one. This is different: the whole gross is the same boundary, large because the aquifer drains recharge through it.

Could the README say which models the floor excludes? The alternative would be to normalise on the variability of the control exchange rather than its gross. Either way a recharge-fed control would pin it, your evolving-exchange model with a runoff recharge term at 1 mm/m.

Three smaller things. The top-level README still says twenty-eight probes. The probe README says `reference_noise_sink` passes every other probe, but it fails `mass/steady-state`. And the ROADMAP block landed above `mass/human-abstraction`'s contributor line.

Validation at this head: `ht validate` 29 probes, 1039 tests and the probe gate passed. I did not rebuild the Docker images.

Skyan1002:
@cehw Thank you — the losing-catchment case is exactly right, and I had assumed it away. The gross derivation took the boundary to be the aquifer's only source; with recharge draining through the same boundary the gross is throughput, independent of storativity, while the response stays `S·Δh`, so the share falls with `S` and the whole confined range went under the floor. Reproduced in the repository with the control you specified: `reference_recharge_exchange`, the evolving boundary plus half the runoff as recharge at `S = 1 mm/m`, measures 2.4e-4 of its gross on the gate seeds and failed on all five, with closure at 0.0000% and every other criterion passing.

I took your second option, with one change of variable. Normalising on `Σ|gwex − mean(gwex)|` did not rescue it — on this catchment the recharge pulses with the rain, so that measure is still 60% of the gross and the share only rose to 4.9e-4. The total variation `TV(g) = Σ|gwex_t − gwex_{t−1}|` does: a steady throughput adds little to how much the exchange moves day to day, and what it adds is the throughput's own variation, which is bounded. On the recharge control the share of the variation is 6.3e-3 at `S = 1 mm/m`, 2.6e-3 at 0.3 and 9.2e-4 at 0.1 — a sweep at fixed ten-day time constant is in the README — against 6.5 for the constant-head control, 1.1e-2 for the evolving one and 1.1e-6 for the token cheat. The floor is now `3e-4`: three times under the value at `S = 0.1 mm/m`, three hundred times above the cheat.

It is not a physical bound, and I have said so where you suggested. The variation has a floor of its own set by how the throughput varies while the response falls linearly with storativity, so a boundary of very small storativity that also carries a strongly varying throughput does go under `3e-4` eventually — about `S = 0.03 mm/m` on this catchment, a thirtieth of the confined top. That is a named class of false FAIL under "What the floor excludes", separate from the cheat paragraph, and far narrower than the class the gross floor excluded. Your table's shape-based separator (a constant paired difference against a decaying pulse) I tried too; it does not separate the constant-head control from the token term, since a constant internal head also gives a constant difference, so I did not use it.

The three smaller things: top-level README says twenty-nine and nineteen; the probe README now says what `reference_noise_sink` does across the suite — every budget probe passed; it fails this probe, `mass/steady-state`, and on some seeds `energy/pet-consistency`, where its evaporation noise shows as noise — and keeps the narrower claim, that no budget probe catches it; the ROADMAP block sits after the contributor line; `docs/writing-a-probe.md`'s reference table has the five controls.

Gate on all five seeds: three `must_pass` (`driven`, `evolving`, `recharge`), two `must_fail` (`noise_sink`, `token`). `pytest` all passing with the new recharge test, which asserts that the same run fails a share of the gross and passes a share of the variation. On the pool point, understood — this settles the physics, and I will wait for the mass pool.

Barbhuiya12 (dismissed):
Reviewed at `83d4ec7`, on a clean worktree. `ht validate` (30 probes, 69 models),
`pytest -q` (1116 passed), the full `ht gate`, and the probe's own gate all pass
here, with the three positive controls and both cheats behaving as declared.

The argument the probe rests on is right, and it is the part worth saying
plainly: under closure `gwex ≈ −R0` identically, so the declared exchange
carries no information about whether it was driven by anything. A counterfactual
is the only instrument that can ask, and prescribing the head as a visible
forcing column is the way to build one. Requiring `gwh` so that a model which
never declared it is N/A rather than judged follows the radiation precedent and
is the right footing.

## I re-derived the calibration rather than reading it

The floor is the whole argument here, so I swept `STORATIVITY_MM_PER_M` on
`reference_recharge_exchange` with conductance held at `S/10` to keep the
ten-day time constant, and took the README's table out afterwards to compare.
Median over the five gate seeds:

| S (mm/m) | resp/gross (README) | mine | **resp/TV** (README) | **mine** | seeds passing |
| --- | --- | --- | --- | --- | --- |
| 10 | 1.5e-3 | 1.5e-3 | 1.0e-2 | 1.0e-2 | 5/5 |
| 3 | 6.3e-4 | 6.0e-4 | 1.0e-2 | 1.0e-2 | 5/5 |
| 1 | 2.4e-4 | 2.4e-4 | 6.3e-3 | 6.9e-3 | 5/5 |
| 0.3 | 8.2e-5 | 8.6e-5 | 2.6e-3 | 3.0e-3 | 5/5 |
| 0.1 | 2.8e-5 | 2.9e-5 | 9.2e-4 | 1.0e-3 | 5/5 |

@cehw's losing-catchment objection reproduces exactly: at `S = 1 mm/m` an exact,
monotone, budget-closing boundary measures **2.4e-4** of its gross and would have
failed a `1e-3` gross floor, while measuring 6.9e-3 of its variation. `closure`
passes at every storativity in the sweep, so those were false FAILs of a correct
model rather than anything the probe should have been catching — which is what
makes the change of denominator a fix rather than a loosening.

I also ran the sweep past where the README stops, because a disclosed limit is
worth checking is where it is said to be:

```
S = 0.05 mm/m   resp/TV 5.1e-4   5/5 seeds pass
S = 0.03 mm/m   resp/TV 3.0e-4   3/5 seeds pass     <- the floor, exactly
S = 0.02 mm/m   resp/TV 2.0e-4   0/5 seeds pass
```

"About `S = 0.03 mm/m`" is accurate to the seed. The class the floor excludes is
where the README says it is, and it is named rather than left for a reader to
find.

## Seed stability of the margins

The floor is argued from five gate seeds, so I ran all five controls over seeds
0-99 plus the gate seeds — 525 paired runs, 1575 model executions — scoring
`response_share_of_variation`, the quantity the floor gates:

```
model                        role         n    wrong   response_share_of_variation
reference_driven_exchange    must_pass   105       0   min 6.154e+00  med 6.550e+00  max 6.869e+00
reference_evolving_exchange  must_pass   105       0   min 1.076e-02  med 1.076e-02  max 1.076e-02
reference_recharge_exchange  must_pass   105       0   min 3.954e-03  med 7.047e-03  max 1.358e-02
reference_noise_sink         must_fail   105       0   0 on every seed
reference_token_exchange     must_fail   105       0   min 1.019e-06  med 1.120e-06  max 1.238e-06
```

Nothing lands on the wrong side of the floor on any seed. `reference_evolving_exchange`
is identical to four figures across all 105, which is right rather than suspicious:
its exchange answers only the head series, and the weather does not enter it.

**One number to correct.** The PR description gives the recharge control as
"resp/TV 5.4e-3 to 8.6e-3" on the gate seeds. Over the wider draw the range is
**3.954e-3 to 1.358e-2** — the minimum is about 27% below the tightest gate seed,
so the binding margin is **13.2x**, not the ~18x the gate seeds suggest. The
token cheat likewise reaches 1.238e-6 rather than 1.1e-6, putting the floor 242x
above it rather than three hundred. Both still comfortable, and neither changes a
conclusion; they are just the figures a wider draw gives, and the README's are
the ones a reader will quote.

That leaves a factor of 3193 between the tightest honest control and the worst
cheat, with the floor 13.2x under one and 242x over the other. Worth saying that
this is deliberate and I think correct: the geometric centre of that gap is about
7.0e-5, and `3e-4` sits a factor of four above it, spending margin against the
cheat to buy room for honest low-storativity boundaries. For a conservation
benchmark that is the right way round — a false FAIL against a model that closes
its budget exactly is the more damaging error, and the storativity sweep is what
makes the choice legible rather than a guess.

I ran this because I got it wrong on #73 — I calibrated a design change on three
gate seeds, reported it clean, and @chrimerss found a `must_pass` model failing at
60% on seed 171 of a 200-seed sweep. Five seeds cannot see a one-percent failure
rate. This one holds.

## One thing I chased and dropped

`response_share` is set to `float("inf")` when `variation == 0`, which reads like
a bypass: a model reporting a constant `gwex` that never answers the head would
have an infinite share and sail through. I built that case — a real run with
`gwex` overwritten by a constant in all three variants — and it **fails**, because
the requirement is `max(s·TV, ε)` and with the variation gone it falls back to ε,
which a zero response cannot clear. The structure already covers it. Recording
the check because the `inf` is the kind of thing a later reader will wonder about
too.

The only residue is cosmetic: `value` is then `inf`, and Python's `json` writes
that as a bare `Infinity`, which is not valid JSON and which a strict parser
(`JSON.parse`, among others) rejects. No model in the field can reach it, so this
is a note rather than a request.

## Coverage, which is the thing I would want settled for the paper

All ten packaged models are N/A on this probe, and the PR says so plainly, which
I would rather have than a number that flatters. But the path to a first real
answer is narrower than it looks from the outside, and worth writing down:

| | |
| --- | --- |
| emit `gwex`, do not declare `gwh` | `cwatm`, `dhbv2`, `flex_lumped`, `flex_topo`, `lisflood`, `sacsma_snow17`, `wflow_sbm` |
| declares neither | `google_flood_forecast`, `summa` |
| right physics, blocked on `closure` | `modflow6` |

The seven in the first row are not one column away, as a model was on
`mass/snowpack-mass-closure`. Consuming an external head is a structural
property — a lumped bucket has no boundary for a head to act on — so `gwh` is
not something their adapters can simply start declaring honestly.

`modflow6` is the one that has the physics. A general-head boundary is this
experiment. It is N/A only because it reports no surface budget, so the `closure`
precondition cannot run on it.

Which is worth noticing: @cehw has now twice built exactly the missing model —
"a native MODFLOW 6.7.0 confined aquifer using the probe's head series, alongside
the unchanged surface bucket" — to check this PR, first for the spinup timing and
then for the losing-catchment normalisation. That pairing is a packageable model
and it would give this probe its first real standing. It is a follow-up rather
than anything owed by this branch, but it seems a shame for it to exist only in
two review comments.

## Smaller

**The PR description contradicts the probe.** Under *What changed since the first
review*, it still reads `min_response_share: 1e-3`. `probe.yaml` ships `3.0e-4`,
and the second section says so. The older section is superseded, but the PR body
is what a reader reads first, and the two numbers are the before and after of the
argument this whole revision turns on.

**The variation is not timestep-invariant.** `variation` is taken on the per-step
depths, `Σ|g₀ₜ − g₀ₜ₋₁|` with `g₀ = gwex·Δt`, so for a smooth exchange it scales
like `Δt`: the differences fall with `Δt` and the step weight falls again, while
the count only rises as `1/Δt`. `G` and the response are both `Δt`-independent, so
the bar `s·TV` drops roughly 24-fold going from `PT1D` to `PT1H` while the
response does not move. Nothing is affected today — the case pins `PT1D` and
there are no per-variant timesteps — and the direction is lenient rather than the
false-FAIL kind, so this is not a defect. Worth a line in the criterion's
docstring, since `exchange_response` is registered in the shared library and the
calibration above is a `PT1D` number.

## What I checked and found correct

- Only `gwh` differs between variants, and only from the first scored step; the
  criterion refuses a case where the spinup differs, which is the failure @cehw
  found at `6ccb404` and it is now enforced rather than described.
- The criterion refuses a model that reached it without declaring the driver,
  and judges a hand-built `RunResult` with no manifest rather than excusing it.
- `closure` kept as a precondition: the declared number has to be the water that
  actually moved before asking what drove it.
- Reversal statistics reported and never gated, after the MODFLOW counter-example.
- `reference_token_exchange` is caught on its size rather than its sign, which is
  the only thing that can catch it, and the mixed-exchange escape it leaves is
  disclosed in its own paragraph rather than folded into the physics.

## Approving

The physics is right, the denominator change is a fix rather than a loosening and
the sweep shows why, the calibration reproduces independently including the limit
it discloses, and 525 runs put nothing on the wrong side of the floor. The two
numbers above are worth updating in the README and the PR body, and the coverage
question is worth an issue, but neither is a reason to hold this.

Approving.

chrimerss (changes requested):
Re-reviewed at `d76e83c`. Every point from my review at `6ccb404` is addressed: the share floor, the reference-model rows, the ROADMAP count, the calibration figures, the unit tests (including the no-manifest case), and N/A rather than PASS for models that do not declare `gwh`. cehw's two findings — the spinup timing and the losing-catchment denominator — are fixed too, and each is pinned by a control.

Verified locally on a clean worktree: `pytest` **1116 passed**, `ht validate` 30 probes / 69 models, the full `ht gate`, and `ht gate --probe mass/exchange-response` (three `must_pass`, two `must_fail`). The three subprocess archive rows reproduce as N/A (INCOMPATIBLE). CI is green at this head.

I'm requesting changes for one reason: the variation denominator has no bar at all for a **steady** sink, so the token cheat comes back in its simplest form.

---

## 1. A steady sink plus a token head term passes (`src/hydroturing/criteria/exchange.py:321`)

The bar is `required = max(3e-4 · TV(g₀), ε · max(G_gross, 1 mm))`, where `TV(g₀) = Σ|gwex_t − gwex_{t−1}|` is taken on the control run. `reference_token_exchange` is caught only because its sink is white noise: the ±30% random ET error gives it a large variation, and the token term cannot clear 3e-4 of that. If the sink is steady, it adds nothing to the variation. The only thing that then moves `gwex` is the token term, so the token clears a share of its own variation whatever its size.

**Reproduction.** Take `models/reference_token_exchange/ht_adapter.py` and change only the reported evaporation (line 134). Keep the same `1e-6` token:

```python
# was: "evspsbl": max(0.0, (canopy_evap + soil_evap)
#                  * (1.0 + ET_BIAS * rng.gauss(0.0, 1.0))) / dt_days,
"evspsbl": ((canopy_evap + soil_evap) / dt_days + 0.3),
```

The model now invents a constant 0.3 mm/day of regional inflow, 1095 mm over the record (13% of the rain), and declares it as `gwex` to close the budget. On the gate seeds:

```
✅ closure            cumulative residual 0.0000% of sum_pr
✅ exchange_response  the declared exchange answers the prescribed head: +0.001825 mm over the record
                      with it raised +0.5 m, -0.001825 mm with it lowered, 7.14 of the 0 mm its
                      exchange moves (required 0.0003); stays available
✅ forcing_fidelity, state_bounds, non_degenerate
```

It passes on **25 of 25** seeds (the five gate seeds plus seeds 0–19). Its head response is **1.7e-6 of its gross exchange — the same share as `reference_token_exchange` (1.5e-6), which this probe exists to catch.**

This is not the mixture the README discloses at line 123. That paragraph describes a *strongly varying* unrelated part, which inflates the variation and leaves an honest-but-small share. A steady unrelated part is the opposite case and the easier escape: it leaves the variation untouched, so the share is set by the token term alone and any conductance passes. A parameter-set regional-groundwater term has roughly this shape — HBV 2.0's `LF`, which the README describes as a deterministic function of drainage area and two parameters. (I did not run dhbv2; that is an inference from its shape.)

**Suggested fix.** Keep the variation floor and add a small share of the gross:

```python
required = max(share * variation, share_of_gross * gross, epsilon * max(gross, 1.0))
# min_response_share_of_gross: 1.0e-5
```

cehw's objection to the gross is that honest throughput inflates it. At `1e-5` rather than `1e-3`, that objection does not bite anywhere the variation floor has not already given up. Measured on 25 seeds (five gate seeds plus 0–19). The two recharge rows at lower storativity are local copies of `reference_recharge_exchange` with `S` changed and `C = S/10`, as in the README's sweep:

| model | current gate | with `1e-5 · gross` added | response / gross, min–max |
| --- | --- | --- | --- |
| `reference_driven_exchange` | 25/25 | 25/25 | 1.89 – 2.06 |
| `reference_evolving_exchange` | 25/25 | 25/25 | 3.39e-3 |
| `reference_recharge_exchange` (S = 1) | 25/25 | 25/25 | 1.99e-4 – 3.03e-4 |
| recharge, S = 0.1 mm/m | 25/25 | 25/25 | 2.46e-5 – 3.58e-5 |
| recharge, S = 0.05 mm/m | 24/25 | 24/25 | 1.23e-5 – 1.79e-5 |
| `reference_token_exchange` | 0/25 | 0/25 | 1.48e-6 – 1.72e-6 |
| **steady sink + 1e-6 token** | **25/25** | **0/25** | 1.67e-6 |
| 30% smooth ET bias + 1e-4 token | 25/25 | 25/25 | 1.33e-4 – 1.55e-4 |

The added term changes no honest control's result, and fails the steady sink on every seed. The last row still passes. It is the mixture the README already discloses (its token is sized to 3e-4 of the sink's variation), and no ratio floor can close that one. What the gross term restores is a *finite* bound on the steady part, which the variation floor alone sets at infinity.

Please also add the steady sink as a third `must_fail` control. Without one, a later change of denominator can reopen this with the gate still green. If you would rather keep the variation floor alone, then the README's "What the floor excludes" needs to say that a steady unrelated exchange of any size passes with any head term, and the steady sink should be documented as passing by design.

---

## Smaller

- **`exchange.py:352`, `:360`, `:366`, `:394` — the variation prints as "0 mm".** `{variation:.0f}` rounds anything under half a millimetre to zero, which produces the "7.14 of the 0 mm its exchange moves" above. Use `.3g`.
- **`src/hydroturing/spec.py:174`.** The comment `# Negative controls for mass/exchange-response.` now heads three positive controls, and after the merge from main also #79's `reference_exchange_sign_error` and `reference_exchange_exact`. This list is the audit trail for code allowed onto the host, so give each probe's controls their own comment.
- **`probes/mass/exchange-response/generate.py:22`, `:42`.** The docstring still says the dry windows are "the window the criterion is scored on" (the probe scores the full record and gates no reversal statistic), and still gives 830 / 700 mm.
- **`models/reference_token_exchange/model.yaml:9`** (and the adapter docstring). These still say the cheat "is caught because its response is a millionth of its own gross exchange". It is now caught on its variation.
- **`AGENTS.md:62` against `AGENTS.md:254`.** Line 62 calls declaring `gwh` "a semantic opt-in, not a statement that the column was read". Line 254 says `uses_forcing` inputs "must be consumed whenever supplied". A reader has to pick one. A sentence at 254 that carves out `gwh` would settle it.
- **README, "What the floor excludes".** Recharge at S = 0.05 mm/m failed on 1 of my 25 seeds (the README says 5/5 at 0.05, measured on the five gate seeds). The excluded class starts a little above where the README puts it.
- **PR body.** "What changed since the first review" still quotes `min_response_share: 1e-3`, and the recharge range quoted there (5.4e-3 to 8.6e-3) is the gate-seed range; the README now gives the wider one.

chrimerss (approved):
Approving at `bccf21b`. This supersedes my `CHANGES_REQUESTED` at `d76e83c`; everything it asked for is in.

The steady sink is caught. `023c965` takes the second term whole — `min_response_share_of_gross: 1e-5` beside the variation share, `reference_steady_sink` as a third `must_fail`, and a unit test that turns the gross term off and asserts the same run then passes, which is the test that keeps this from being reopened by a later change of denominator. The README section that argues it is the right length for what it settles, and it records the band I measured at `S = 0.05` rather than rounding it away.

## What I checked at this head

```
pytest -q                               1118 passed
ht validate                             30 probes, 70 models
ht gate --probe mass/exchange-response  3 must_pass, 3 must_fail
CI                                      gate + container green at bccf21b
```

I re-ran my 25-seed sweep (the five gate seeds plus 0–19) against the shipped criterion rather than a patched copy, on the three honest controls, two local low-storativity copies of `reference_recharge_exchange` with `C = S/10`, the three cheats, and the mixture case:

| model | seeds passing | response / gross, min–max |
| --- | --- | --- |
| `reference_driven_exchange` | 25/25 | 1.89 – 2.06 |
| `reference_evolving_exchange` | 25/25 | 3.39e-3 |
| `reference_recharge_exchange` (S = 1) | 25/25 | 1.99e-4 – 3.03e-4 |
| recharge, S = 0.1 mm/m | 25/25 | 2.46e-5 – 3.58e-5 |
| recharge, S = 0.05 mm/m | 24/25 | 1.23e-5 – 1.79e-5 |
| `reference_noise_sink` | **0/25** | 0 |
| `reference_token_exchange` | **0/25** | 1.48e-6 – 1.72e-6 |
| `reference_steady_sink` | **0/25** | 1.67e-6 |
| 30% smooth ET bias + a 1e-4 token | 25/25 | 1.33e-4 – 1.55e-4 |

No honest control lost a seed to the gross term, and the 24/25 at `S = 0.05` is the variation term, not the new one — which is what the README says. The last row is the mixture the README discloses in its own paragraph: a sink that varies, with a token sized above `3e-4` of that variation. No ratio floor closes it, and saying so plainly is better than a floor that pretends otherwise.

## What I pushed

`bccf21b`, on top of `023c965`, docs and dead code only:

- `reference_steady_sink` had no row in README.md's reference-model table or in `docs/writing-a-probe.md`'s, which lists the other five. That is checklist item 1 in AGENTS.md and no test catches it — worth knowing for the next probe that adds a control.
- `response_share_of_gross_required` was a second name for `required_share_of_gross` in the diagnostics; one is enough.
- The steady sink carried the noise sink's `ET_BIAS`, `import random` and `rng`, which it does not use, and the bucket's docstring line claiming it is bit for bit the bucket, which it is not.

1118 tests, `ht validate` and the probe gate pass on it, and CI is green.

## For the record, not for this branch

The probe scores no model in the current field, and the PR says so rather than dressing it up. `modflow6` is the one with the physics and is N/A only because it reports no surface budget, so `closure` cannot run on it. @cehw has now built the missing pairing twice — a native MODFLOW general-head boundary alongside the surface bucket — to check this branch. That is worth an issue: it would give the probe its first real standing, and it is the model the whole argument was written for.

Thanks for the three rounds, and for taking the counter-examples as specifications rather than as arguments to win. The probe that lands is a different and better one than the probe that was proposed: a prescribed driver, a paired counterfactual, three exact boundaries that must pass including one that only exists because a false FAIL was found, and three cheats that must not.

Barbhuiya12 (approved):
Re-reviewed at `bccf21b`, after my approval was dismissed by the new commits.
`ht validate` (30 probes, 70 models), `pytest -q` (1118 passed), the full gate and
the probe's own gate all pass here, with all three cheats tripping
`exchange_response` alone.

All three of my points are in: the wider-draw margins are quoted, the step
dependence is written into the module docstring, and `response_share` is `None`
rather than `float("inf")` so the report stays valid JSON. Nothing further on
those.

The reason to re-read this properly is the middle commit, which is new physics
rather than an answer to a review.

## The steady sink is a real hole and I missed it

I had this hypothesis and dismissed it too early. I tested a `gwex` with
*exactly* zero variation, found the `max(s·TV, ε)` structure caught it on the ε
term, and wrote that the structure already covered it. The actual exploit is
near-zero variation with a large gross, and it goes straight through:

```
reference_steady_sink, seed 5
  variation                0.0002555 mm        gross  1095 mm
  share of variation       7.143               floor  3e-4     clears, by four orders of magnitude
  share of gross           1.667e-06           floor  1e-5     BELOW
  required                 0.01095 mm          response +0.00183 mm      -> FAIL
```

A share of the variation is a bar against a *varying* unrelated part, because that
is what inflates a gross. It sets no bound at all on a steady one: a constant sink
adds nothing to the variation, so the only thing moving the exchange is the token
term itself, and a token of any size clears a share of its own movement. Hence a
share of variation of 7.143 on a model that answers the head with 1.7e-6 of what
it declares. Under the criterion I approved, this passes.

So the gross returning as a small second floor is right, and the way it is framed
— the bar being the larger of the two shares, with the gross share two orders of
magnitude smaller so it binds only where the variation term has given up — is the
correct division of labour between them.

## The new floor costs the honest range nothing

That is the thing a second floor can break, so I re-ran the storativity sweep on
`reference_recharge_exchange` at a fixed ten-day time constant, median over the
five gate seeds, now reading both floors:

```
S (mm/m)   gross (mm)   resp/TV    resp/gross    TV >= 3e-4   gross >= 1e-5   verdict
      10         1832   1.0e-02       1.5e-03           ok              ok    5/5 pass
       3         1733   1.0e-02       6.0e-04           ok              ok    5/5 pass
       1         1729   6.9e-03       2.4e-04           ok              ok    5/5 pass
     0.3         1729   3.0e-03       8.6e-05           ok              ok    5/5 pass
     0.1         1729   1.0e-03       2.9e-05           ok              ok    5/5 pass
    0.07         1729   7.1e-04       2.0e-05           ok              ok    5/5 pass
    0.05         1729   5.1e-04       1.4e-05           ok              ok    5/5 pass
    0.04         1729   4.1e-04       1.2e-05           ok              ok    5/5 pass
    0.03         1729   3.0e-04       8.7e-06           ok           BELOW    0/5 pass
```

"Across a storativity sweep from 10 to 0.05 mm/m it changes no honest control's
verdict" reproduces, with 0.04 in hand as well. The excluded class moves from
about `S = 0.03` to about `S = 0.035` — at 0.03 the variation share was already
sitting exactly on its own floor, so the boundary barely shifts and the README's
disclosed limit still describes it.

Over seeds 0-59 plus the gate seeds, nothing lands on the wrong side:

```
model                        role         n   wrong   response_share_of_variation
reference_driven_exchange    must_pass   65       0   6.189e+00 .. 6.869e+00
reference_evolving_exchange  must_pass   65       0   1.076e-02  (identical on every seed)
reference_recharge_exchange  must_pass   65       0   3.964e-03 .. 1.171e-02
reference_noise_sink         must_fail   65       0   0 on every seed
reference_token_exchange     must_fail   65       0   1.019e-06 .. 1.219e-06
reference_steady_sink        must_fail   65       0   caught on the gross floor, not this one
```

The binding honest control is still `reference_recharge_exchange` at 3.964e-3,
13.2x over the floor, matching the wider-draw figure now quoted in the README
rather than the gate-seed range that was there before.

## Approving again

The two cheats I could think of are pinned by controls, the one I could not think
of is now pinned by a control, the second floor is measured against the honest
range rather than asserted to be safe, and the limit it leaves is written down.

For the record, the coverage point from my first review is unchanged and still not
a blocker: all packaged models are N/A because consuming an external head is
structural rather than a column, and `modflow6` is N/A only for want of a surface
budget. @cehw's native MODFLOW-plus-bucket rig, built twice now to review this
branch, is the model that would give this probe its first real standing, and it
would be a shame for it to live only in review comments.

Issue #87, opened 2026-09-15
[PROBE] momentum/recession-friction-floor: on a force-free step, flow decays no faster than the declared friction time-scale and never accelerates while draining


mchen15ouedu:
### Authors

Mengye Chen, School of Civil Engineering and Environmental Science, University of Oklahoma, Norman OK, 0000-0003-2135-1670, @mchen15ouedu

### Conservation law

momentum

### Residual equation

`momentum/recession-friction-floor`. On a step with no rain and no external force, the momentum in the reach can only be lost to friction, and friction has a fastest rate. So the flow may not stop faster than the declared friction time-scale allows, and it may not speed up at all. No flow that suddenly becomes zero; no momentum from nothing.

**The momentum variable.** This is the one criterion of the four where velocity is the primary form and discharge the shadow. The dissipation term in the momentum equation is the friction slope, `S_f ∝ u²`, and it acts on the speed; the declared time-scale `tau_f = L / c` with `c = (5/3) u_n` is a statement about `u`. Per unit length of reach the momentum is `ρ A u = ρ Q`, so for a model that reports only `dis` the bounds are written on `dis`, and they hold there because on a prismatic reach `Q` and `u` are linked by the rating and fall together. A model that reports the section-averaged outlet velocity as a diagnostic `vel` [m/s] is scored on `vel` as well as on `dis`, and both must pass. Where `stage` is also reported, `vel` must be the velocity of the same water, `dis = vel · w · (stage − bed_elevation_m)` within 5 percent at every wet step.

The case declares `tau_f` [s] in `static.json`, the fastest admissible decay time-scale of the reach: the time the kinematic wave takes to clear the reach at bankfull, from the declared Manning geometry. Nothing the reach can do drains its momentum faster than that.

Let a step `i` be *force-free* when `pr_i = 0` and, where the case prescribes a boundary head, `hds_i` is at its baseline. Writing `m` for each momentum variable reported, two one-sided bounds over force-free steps:

**No sudden stop.**

```math
m_{i+1} \;\ge\; m_i\, e^{-\Delta t / \tau_f}
```

**No momentum from nothing.** On a force-free step once the wave has passed, meaning at or after the peak of `dis` within the current contiguous dry spell,

```math
i \ge \arg\max_{\text{spell}} Q \;\Rightarrow\; m_{i+1} \le m_i
```

[...]

Issue #86, opened 2026-09-15
[PROBE] momentum/routing-causality: rain-to-discharge lag is non-negative at onset and centroid, on dis, in a single run


mchen15ouedu:
### Authors

Mengye Chen, School of Civil Engineering and Environmental Science, University of Oklahoma, Norman OK, 0000-0003-2135-1670, @mchen15ouedu

### Conservation law

momentum

### Residual equation

`momentum/routing-causality`. If the water came from rain, the flow cannot respond before the rain fell. The lag from rain to routed flow is non-negative, at the onset and at the centroid, for every storm in a single run.

`mass/causality` already asks the paired version of this question, and asks it well: nothing may change before an added storm. But its variable list is `[mrro, evspsbl, mrso, snw, canopy]`. It does not look at `dis`. A model can generate runoff causally and then route it with a centred kernel, a bidirectional smoother, or a learned router that sees the whole runoff series it is handed, and `causality` never finds out, because the anticipation lives entirely between `mrro` and `dis`. That gap is the momentum law's to close, since routing is where momentum is conserved.

**The momentum variable.** Per unit length of reach the momentum is `ρ A u = ρ Q`, so a model that reports only `dis` is scored on `dis`. A model that reports the section-averaged outlet velocity as a diagnostic `vel` [m/s] is scored on `vel` as well, and both must pass: a hydrodynamic emulator can keep its discharge causal while its velocity field is smoothed in time, and the anticipation then lives between `dis` and `vel`. Where `stage` is also reported, `vel` must be the velocity of the same water, `dis = vel · w · (stage − bed_elevation_m)` within 5 percent at every wet step.

For each isolated storm `k`, with first rain at `t_0^k` and rain centroid `t_c^k`, and for each momentum variable `m` reported, take the flow above its pre-storm baseline `m' = m − m_base` and define:

```math
\tau_\mathrm{onset}^k = t_\mathrm{rise}^k - t_0^k, \qquad
\tau_\mathrm{cent}^k = \frac{\sum_i t_i\, m'_i}{\sum_i m'_i} - t_c^k \qquad [\mathrm{h}]
```

where `t_rise` is the first step at which `m'` exceeds 2 percent of the storm's eventual peak rise. Both lags must be non-negative for every storm and every reported momentum variable:

```math
\tau_\mathrm{onset}^k \ge -\delta, \qquad \tau_\mathrm{cent}^k \ge 0
```

[...]

Issue #85, opened 2026-09-15
[PROBE] momentum/forced-flow-reversal: no backflow without a force, and no more reversed flow than the force could have moved


mchen15ouedu:
### Authors

Mengye Chen, School of Civil Engineering and Environmental Science, University of Oklahoma, Norman OK, 0000-0003-2135-1670, @mchen15ouedu

### Conservation law

momentum

### Residual equation

`momentum/forced-flow-reversal`. Water flows from high to low unless something pushes it the other way. When something does — a storm surge, a tidal boundary, a flash-flood bore rebounding off a valley wall — the flow may reverse, but only by as much as the push could have moved. Both halves are conservation statements: no reversal without a force, and no more reversal than the force supplied.

The case prescribes a downstream boundary head as a new forcing column `hds` [m], the water level imposed at the reach outlet, and runs paired variants: `control`, where `hds` stays at its baseline, and `surge`, identical except for a raised `hds` over a surge window `S`. `dis` is signed, positive downstream; a negative `dis` is a reversal.

**The momentum variable.** Per unit length of reach the momentum is `ρ A u = ρ Q`, so a model that reports only `dis` is scored on `dis`, and the reversal cap is then written on volume, the mass shadow of the impulse. A model that reports the section-averaged outlet velocity as a diagnostic `vel` [m/s], signed positive downstream, is scored on `vel` as well, and for it the cap can be written on momentum directly against the impulse the surge supplied, which is the bound the physics actually states. Both must pass. Where `stage` is also reported, `vel` must be the velocity of the same water, `dis = vel · w · (stage − bed_elevation_m)` within 5 percent at every wet step.

Three checks:

**Unforced sign.** Wherever the boundary applies no adverse head, the flow does not reverse, in both variants:

```math
Q_i \ge -\varepsilon_Q, \qquad u_i \ge -\varepsilon_u \qquad \text{for all } i \text{ with } \mathrm{hds}_i \le \mathrm{hds}_\mathrm{base}
```

**Bounded reversal.** On discharge, over the surge window, the reversed volume cannot exceed what the surge head could have backed up into the reach:

```math
V_\mathrm{rev} = \sum_{i \in S} \max(-Q_i,\,0)\,\Delta t \;\le\; V_\mathrm{cap} = w\,L\,\max_{i \in S}\left(\mathrm{hds}_i - \mathrm{hds}_\mathrm{base}\right) \qquad [\mathrm{m^3}]
```

[...]

Issue #84, opened 2026-09-15
[PROBE] momentum/event-impulse-balance (withdrawn)


mchen15ouedu:
> **Withdrawn 2026-09-16.** Tested against the four required baselines, the residual does not hold up; see the comments. Kept for the record.

### Authors

Mengye Chen, School of Civil Engineering and Environmental Science, University of Oklahoma, Norman OK, 0000-0003-2135-1670, @mchen15ouedu

### Conservation law

momentum

### Residual equation

`momentum/event-impulse-balance`. Over an event, the accelerations a routing store puts into the flow must be paid back by decelerations before the event is over. A router that only ever speeds the flow up, and never lets it slow back down, is adding momentum the inflow never supplied.

**The momentum variable.** The momentum of the water in a reach, per unit length, is `ρ A u = ρ Q`, so `dis` [m3/s] is the momentum density itself and not merely a proxy for it. A model that reports only `dis` is scored on `dis`. But a change in `ρ A u` can be a change of section or a change of speed, and only the velocity tells them apart. A model that reports the section-averaged velocity at the reach outlet, as 1-D and 2-D hydrodynamic models do, declares it as a diagnostic `vel` [m/s], signed positive downstream, and is then scored on `vel` as well as on `dis`; both must pass. Where `stage` is also reported, `vel` has to be the velocity of the same water, `dis = vel · w · (stage − bed_elevation_m)` within 5 percent at every wet step, so a decorative velocity cannot pass while the discharge does something else.

*Revised 2026-09-16 after running the four required baselines; the original window, closing on the inflow's return to its pre-event level, failed honest models on a third of events and is recorded in the comments below with the numbers.*

Writing `m` for whichever momentum variable is being scored and `I` for the reach's inflow, the generated runoff `mrro` in the same unit (m3/s through `area_km2`, or `I / A` against `vel`), the residual over an event window `W` is the outflow's change less the inflow's change:

```math
r = \left(m_\mathrm{close} - m_\mathrm{start}\right) - \left(I_\mathrm{close} - I_\mathrm{start}\right)
```

[...]

PR #95, opened 2026-09-16
Contract: signed dis, optional vel diagnostic, hds boundary-head forcing

mchen15ouedu:
The `/io` contract additions the four accepted momentum proposals rest on (#84, #85, #86, #87). No probe uses them yet, so nothing in the suite changes behaviour; this lands the definitions first so the probes can be built against `main` rather than each carrying its own.

**`vel`**, a new optional diagnostic in `DIAG_VARS`, m s⁻¹: the section-averaged velocity of the water leaving the reach, positive downstream, the velocity of the same water `dis` counts. Momentum per unit length of reach is ρ·A·u = ρ·Q, so `dis` is the momentum density and a model that reports only `dis` is scored on it; one that also reports `vel` is scored on both, because only the velocity separates a change of speed from a change of section, and friction acts on the speed. Hydrodynamic models report the section mean at the outlet; models without a velocity field report nothing.

**`dis` sign convention**, stated in its AGENTS.md row: positive downstream, negative only while water actually moves upstream under a prescribed surge or backwater. A model that clips at zero says so in `run.json`.

**`hds`**, a forcing column a probe may prescribe, m: the water level imposed at the reach outlet by a surge, a tide or a backwater, documented beside `abstr`. Consumption is declared the way `rlds` is: a model that declares neither `needs_forcing` nor `uses_forcing` for it is N/A (INCOMPATIBLE) on such a probe, and a model that declares it and reports a hydrograph the head never touched has failed. Forcing columns are not whitelisted in `spec.py`, so this is documentation only.

Checks: `ht validate` passes on all 27 probes and 58 models; `tests/test_docs_in_sync.py::test_agents_doc_defines_every_variable` passes with `vel` in the table; full `pytest`: 985 passed, 8 skipped in 394.76s (0:06:34).


licm13 (commented):
Hi @mchen15ouedu, thanks for this.
Landing the contract before the four probes makes sense to me, and I would much rather review these shared definitions once here than see the same questions reappear independently in #84–#87. I checked the change on `49cf555` against the harness as it stands, mainly from the energy side, where two of these patterns already have fairly close precedents.

One thing seems to connect most of my comments: **the three new rules are currently written in a way that sounds as though the harness already implements them, while in practice none of the three is enforced yet.** I do not think that argues against landing the contract first — having the contract lead the implementation is perfectly reasonable — but it may help to make that distinction a little more explicit for whoever picks up #84–#87.

## 1. "Optional, scored where reported" does not have a first-class mechanism today, although there is a useful precedent

`required_vars` is `requires_fluxes + requires_states + requires_diagnostics` (`spec.py:234`). So if a probe wants to read `vel` through `requires.diagnostics: [vel]`, that is currently all-or-nothing: every model without a velocity field becomes `N/A (INCOMPLETE)`. `momentum/stage-discharge-monotonic` uses `diagnostics: [stage]` in exactly that way, so there is not yet an example of an optional diagnostic expressed through `requires`.

There is, however, a very close precedent outside `requires`. `flux_identity` reads `sbl` opportunistically:

`reported_split = bool(sublimation_var) and sublimation_var in w.table.columns` (`coherence.py:127`)

and scores the split where a model reports it while still scoring the rest of the criterion where it does not. `melt_energy` follows the same pattern at `coherence.py:804`. That seems very close to the intended meaning of "optional, scored where reported" for `vel`.

The main caveat is the one the automated review pointed out, and I think it is worth preserving because we have already encountered it on the energy side. Reading a column outside `required_vars` bypasses the finiteness validation in `protocol.read_result`, so `melt_energy` carries its own explicit check at `coherence.py:808-811`. Without that, a NaN in one `sbl` row can propagate into the calculation and produce a misleading diagnostic such as `melt of nan mm demands nan W/m2`.

A momentum criterion that reads `vel` opportunistically will probably want the same guard.

My preference would therefore be to keep `vel` in `DIAG_VARS` and `UNITS` as this PR already does, but phrase the AGENTS.md rule more in terms of criterion behaviour than `requires`, for example: a criterion may read `vel` where a model reports it, and should validate that optional field itself. A pointer to `sbl` as the worked example might make the intended implementation especially clear.

## 2. `hds`: I read this closer to the `abstr` case than the `rlds` case

I think the distinction here is less about the variable itself and more about **whether the criterion remains meaningful when the model did not consume the input**.

For `abstr`, it does. The water was removed from the catchment by the case. A model that never reads the abstraction column still reports a water budget, and that budget can still be assessed against the imposed abstraction. AGENTS.md already gives essentially this rationale: the model reports everything the criterion needs.

For `rlds` and `eps` in `energy/radiation-consistency`, the situation is different. The identity

`rlus = eps*sigma*ts^4 + (1-eps)*rlds`

is specifically about the case-provided sky forcing and emissivity. If a model uses its own values instead, it may be internally consistent while simply answering a different forcing problem. Treating that as INCOMPATIBLE therefore makes sense to me.

By that test, I think `hds` behaves more like `abstr`. A downstream surge is a physical boundary condition imposed on the reach. If a model routes water and reports `dis`, but does not allow the downstream head to influence the solution, then a reversal criterion can still make a meaningful statement about that behaviour.

The model that seems genuinely outside the scope is a lumped rainfall-runoff model with no routed reach at all. In practice, though, that model would not report `dis`, so `requires.fluxes: [dis]` should already make it `N/A (INCOMPLETE)`.

For that reason, I would lean toward letting eligibility fall out of `requires.fluxes` rather than adding a separate INCOMPATIBLE rule for silence on `hds`.

If we do want an explicit escape hatch for a routing model that genuinely cannot represent backwater, the existing idea in the PR — having the model declare that it clips at zero in `run.json` — seems cleaner to me than inferring that limitation from silence. That leads into the next point.

## 3. The proposed `run.json` declaration is not read yet

At the moment, `protocol.py` parses `run.json` for `status` and `n_steps` only (`protocol.py:256-263`). I could not find a schema for the additional field under `schemas/`, or another place in the tree that consumes it.

So the sentence saying that a model which clips at zero declares this in `run.json` describes a useful design, but not yet an enforced one. A model can currently add, omit, or misstate that field without changing harness behaviour.

I would keep the sentence, because the declaration mechanism seems like the right direction. It may just be worth either adding a reader alongside the contract or marking it as not yet enforced, so adapter authors do not assume that the declaration already affects eligibility or scoring.

## 4. The forcing-name asymmetry is reproducible

I also reproduced the automated review's point on `49cf555`. Starting from a merged probe and changing only `requires`:

```text
forcing: [h_ds, total_nonsense]         ACCEPTED
requires_forcing=('h_ds', 'total_nonsense')

static: [made_up_attribute]             ACCEPTED

diagnostics: [stage, total_nonsense]   refused
unknown variables in requires: ['total_nonsense']
```

`load_probe` validates `fluxes`, `states`, and `diagnostics` against their vocabularies, but passes `forcing` and `static` through as free strings (`spec.py:437-458`). The existing comment explains why the vocabulary check matters: a typo in a required output variable would otherwise make every model look incomplete.

I could not find an equivalent check that every `requires_forcing` entry was actually written by the case generator. `check_declared_inputs` compares the required names against the model declaration (`harness.py:346-355`), and the visibility check covers `needs_forcing`, but that still leaves the case-generator side unchecked.

The failure mode seems particularly relevant to `hds`: a probe could declare `requires.forcing: [hds]` while the generator accidentally writes `h_ds`; `ht validate` would still pass, a model declaring `uses_forcing: [hds]` could be considered compatible, and the model would then be scored on a backwater condition that was never actually present in the generated forcing.

This is pre-existing, so I would not necessarily make it a blocker for this PR. But with four momentum probes about to use `hds`, it may be a good opportunity to add the inexpensive half of the guard: in `build_case`, assert that every `requires_forcing` column exists in the generated forcing. That would catch the dangerous direction without requiring a global forcing vocabulary.

## Smaller point

The `vel` row includes `vel * A = dis` conditioned on the model also reporting `stage`, but `A` is not itself part of the contract and there is no generic mapping from `stage` to area. As written, I read that identity more as explanatory than directly checkable.

That is completely fine if it is intended as a definition or motivation. It may just be worth saying that explicitly, or indicating which momentum probe will supply the geometry needed to close the relation.

Overall, I am still comfortable with the contract-first approach. Most of these comments are really about making the boundary between the intended contract and the currently enforced harness behaviour explicit before the four probe implementations start.

If useful, I would also be happy to take another look at whichever of #84–#87 ends up reading `vel`, since the optional-column finiteness guard is probably the easiest detail to miss when translating the `sbl` pattern.

Barbhuiya12 (commented):
Came to this from #94, which lands a prescribed boundary head of its own, so the
two are worth reading together. On `49cf555`: `ht validate` passes (27 probes, 58
models), `ht gate` passes, `test_docs_in_sync.py` passes with `vel` in the table.
Landing the definitions before the four probes that need them is the right order,
and the `vel` justification — momentum per unit length is ρ·A·u = ρ·Q, so only the
velocity separates a change of speed from a change of section — is the part that
makes `vel` worth having rather than a convenience column.

One thing I would settle before this merges, and one coordination point.

## The signed `dis` is on a collision course with `rating_monotonic`

"No probe uses them yet, so nothing in the suite changes behaviour" is true today,
and it is exactly what makes this easy to miss. `rating_monotonic` is already
merged, and it asserts that stage rises monotonically with discharge. It bins by
discharge and accumulates every fall in stage as a deficit:

```python
running_max = np.maximum.accumulate(hb)
deficits = np.maximum(0.0, running_max - hb)
```

A prescribed surge is the one situation where that assertion is false in the
right-hand direction. Under a backwater the outlet level is *held up* while the
flow reverses, so the highest stages sit at the most negative discharges — and
once `dis` is signed, those become the lowest bins. I ran it on
`momentum/stage-discharge-monotonic` with `reference_rating`, taking its own
output and reversing the flow through one surge episode while the backwater holds
the level up:

```
as-run (ordinary one-way flow):   pass   deficit  0.0004 m
with a prescribed surge:          FAIL   deficit 13.0583 m   (tolerance 0.1033 m)
  "stage falls as discharge rises: 13.0583 m of running-maximum deficit
   over 12 bins, source dis"
```

That is a hydrodynamic model being scored as violating a rating law for reporting
exactly the reversal this PR's `dis` row tells it to report. The model that clips
at zero — which this PR asks to declare the clip in `run.json` — passes instead.
As it stands the contract would reward the clip.

Nothing needs fixing in the suite today, because no probe prescribes `hds` yet.
But this PR is where the convention is decided, so it is the right place to record
what the convention obliges. Three ways out, and I do not think it is mine to pick:

- `rating_monotonic` reads the rating over forward-flow steps only, and says so;
- a probe that prescribes `hds` may not also apply `rating_monotonic`, stated in
  the criterion's docstring rather than left to each probe author;
- or the momentum probes carry a reversal-aware rating criterion of their own.

Whichever, the first surge probe should not be the place this is discovered.
`rating_loop` deserves the same look — it classifies limbs on `dq > 0`
(`rating.py:376`), and "rising limb" has no settled meaning across a reversal.

## `hds` and `gwh` are the same contract pattern under two naming schemes

#94 (`mass/exchange-response`, which I reviewed and approved today) adds `gwh`:
a prescribed external head in metres, supplied as a visible forcing column,
consumption declared under `needs_forcing`/`uses_forcing`, N/A (INCOMPATIBLE) for
a model that declares neither, and documented beside `abstr`. That is this PR's
`hds`, clause for clause, for a different boundary.

The physics is genuinely different — `hds` is a surface level at the reach outlet,
`gwh` an aquifer head — so two columns is right and I am not suggesting they merge.
The naming is the issue: `hds` puts the `h` first and the place second, `gwh` the
other way round. A reader meeting both learns no rule. `hds`/`hgw`, or `swh`/`gwh`,
would at least be a scheme.

Neither is in `main` yet and both PRs are open, so this is cheap now and permanent
later. It is also already concrete rather than hypothetical: both PRs insert their
paragraph into the same stretch of AGENTS.md beside `abstr`, and this branch
already conflicts there.

## Rebase

`AGENTS.md` conflicts with `main` (`spec.py` auto-merges). The branch is at a base
of 27 probes and 58 models, which is where the PR's stated check numbers come
from; `main` has moved well past that, and #102's `snm` and #94's `gwh` both touch
the same variable table. Worth rebasing and re-running the numbers before a
maintainer reads them.

## Smaller

`vel · A = dis` in the `vel` row is conditioned on the model "also reporting
`stage`", but `A` is not a contract quantity and stage alone does not give it —
it needs the section, and the section comes from probe-level statics (`width_m`,
`bankfull_depth_m` in `momentum/stage-discharge-monotonic`'s generator) plus the
assumption that the bed is the rectangle those describe. True on the momentum
probes, which supply exactly that; the row reads as though it were general. Worth
saying the identity holds where the probe pins the section.

## Checked and correct

- `vel` added to `DIAG_VARS` and `UNITS`, declared under `emits.diagnostics`,
  excluded from water-storage sums — the same footing as `stage` and `ts`.
- Optional in the right way: a model without a velocity field reports nothing and
  is scored on `dis` alone, rather than being made N/A for lacking a diagnostic.
- `hds` consumption declared as `rlds` does it, so a model that never declared the
  head is N/A rather than judged — the same footing #94 and the radiation probe use.
- Forcing columns are not whitelisted in `spec.py`, so `hds` really is
  documentation, and the PR says so instead of implying a code change.

PR #102, opened 2026-09-18
[PROBE: mass] snowpack-mass-closure: internal snowpack water-balance closure

### LINKED ISSUE #17: [PROBE] mass/snowpack-mass-closure: internal snowpack water-balance closure
### Conservation law

mass

### Residual equation

$$
dS_{snow} / dt = P_{snow} - Q_{melt} - E_{sbl} [mm/day]
$$

where $S_{snow}$ is the total water stored within the model's snowpack control volume [mm]; $P_{snow}$ is snowfall entering the snowpack [mm/day]; $Q_{melt}$ is meltwater leaving the snowpack and entering the underlying hydrologic system [mm/day]; and $E_{sbl}$ is sublimation from the snowpack [mm/day].

The cumulative snowpack mass-balance residual is defined as: 

$$
R_{snow} = P_{snow,cum} - Q_{melt,cum} - E_{sbl,cum} - (S_{snow,end} - S_{snow,start})
$$

where `cum` denotes the cumulative flux over the full evaluation period; $S_{snow, start}$ and $S_{snow, end}$ represent the total snowpack water storage at the beginning and end of the evaluation period, respectively.

For a perfectly mass-conserving snowpack, $R_{snow} = 0$.

Allowing for numerical, temporal-discretization, and diagnostic-output errors, I propose to evaluate the normalized closure error as

$$
\epsilon_{snow} = |R_{snow}| /  P_{snow,cum} \le 0.05
$$

The denominator represents the total snowfall input to the snowpack during the evaluation period. The synthetic forcing will be designed to ensure substantial cumulative snowpack input, so that the denominator remains safely away from zero.

Thus, a model passes when the unexplained cumulative snowpack mass imbalance is no greater than 5% of the total water supplied to the snowpack.

An unphysical model would fail  this probe if water is lost or created inside the snowpack without being reported as snowpack outflow, sublimation, or a change in snow storage. 

For example, a model may store only part of the snowfall and silently lose the rest. This would produce a non-zero $R_{snow}$. The model would fail if the normalized closure error is greater than 5%.

A trivial model could also satisfy the mass balance by immediately routing all snowfall out of the snowpack while maintaining zero snow storage. Therefore, the closure test should be paired with a simple non-degeneracy check to ensure that the model develops a non-zero snowpack under the prescribed cold-season forcing.

The case will contain two complete synthetic snow cycles. Each cycle will include three stages: snow accumulation, cold storage, and snowmelt.

During the accumulation stage, air temperature will remain below freezing. Several precipitation events with different timing and intensity will be generated. These conditions will produce substantial snowfall and build a clear snow storage.

During the cold storage stage, temperature will remain below freezing and precipitation will be zero. The accumulated snow should remain in the snowpack, except for explicitly simulated losses such as sublimation. This stage test whether the model can maintain snow mass without unexplained losses.

During the melt stage, temperature will rise above freezing. Precipitation will be zero so that snowpack outflow mainly comes from melting snow rather than rainfall. The snow storage should decrease clearly during this period.

Precipitation timing, precipitation amount, and temperature variability will change with the random seed。This avoids relying on a single fixed test case and prevents the model from memorising specific examples.

Each stage is provisionally set to 60 days. The three stages will be repeated twice, giving a total simulation period of 360 days. Every generated case will contain sufficient snow input, a clear accumulation period, a stable cold-storage period, and a clear melt period. This ensures that the main terms in the snowpack mass balance are exercised.

### Will you build it?

Yes — assign it to me

jinlong-hu:
## Summary

Adds the accepted `mass/snowpack-mass-closure` probe from #17.

The probe tests internal snowpack water-balance closure using precipitation,
snow-module liquid outflow, sublimation, and snow-water storage. Closure is
evaluated separately across contiguous accumulation, storage, and melt
regimes so that errors in different snow regimes cannot cancel.

## Implementation

- adds `snm`, the liquid water delivered from the snow module to the ground,
  to the model contract
- emits `snm` from `reference_bucket`, `sacsma_snow17`, and
  `reference_sublimating`
- extends `closure` to support selected states, optional sinks, and
  regime-wise closure
- adds the snow-specific `snowpack_response` criterion
- adds `reference_snow_bypass` to test internal snowpack leakage while
  preserving the whole-catchment water balance
- adds `reference_snowless` to test the degenerate no-storage solution
- adds the synthetic `mass/snowpack-mass-closure` probe
- updates documentation, site metadata, and archived model results

## Validation

- `pytest -q`: 1084 passed, 8 skipped
- `ht validate`: validation passed: 30 probe(s), 65 model(s), suite 0.1.0
- `ht gate --probe mass/snowpack-mass-closure`: passed
- `ht gate`: passed

Closes #17


Barbhuiya12 (commented):
Ran this on `fd990fb`: `ht validate` (30 probes, 65 models), `pytest -q` (1092 passed),
the full `ht gate`, and a 200-seed sweep of the probe against both `must_pass`
baselines. The probe is a good one — the internal snowpack budget is a real gap
in the suite, the regime-wise absolute-residual sum is the right way to stop
accumulation and melt errors cancelling, and the `_regime` annotation and
`min_window_days` are used exactly as the harness intends (I confirmed the model
only ever receives `time,pr,tas,pet`).

Two things I would fix before this merges. Both are reproducible.

## 1. A `must_pass` baseline fails while conserving mass to machine precision

`sacsma_snow17` is declared `must_pass`. Over 200 seeds it fails on two of them,
seeds 32 and 179:

```
seed 32, sacsma_snow17:
  closure            pass   cumulative residual 4.12e-13 mm of 553.0 mm
  state_bounds       pass
  snowpack_response  FAIL   cycle 2 melt-end snow fraction 0.1164 > 0.1
  -> verdict FAIL / VIOLATION
```

The model closes its snowpack budget to 4e-13 mm and is reported as violating
mass conservation. The five gate seeds do not contain this case.

This is not an unlucky draw. It is the calendar. `generate.py` starts the record
at 2000-01-01 and lays two 180-day cycles back to back, so the two cycles sit at
opposite ends of the astronomical year:

```
accumulation  2000-01-01 -> 2000-02-29   doy   1-60    147.5 mm
storage       2000-03-01 -> 2000-04-29   doy  61-120     0.0 mm
melt          2000-04-30 -> 2000-06-28   doy 121-180     0.0 mm   <- summer solstice
accumulation  2000-06-29 -> 2000-08-27   doy 181-240   405.5 mm   <- snowfall at -10 C in July
storage       2000-08-28 -> 2000-10-26   doy 241-300     0.0 mm
melt          2000-10-27 -> 2000-12-25   doy 301-360     0.0 mm   <- winter solstice
```

Cycle 2 asks the model to remove 2.75x more snow than cycle 1, in the 60 days
centred on the winter solstice — and Snow-17's melt factor is a sinusoid of day
of year. `models/sacsma_snow17/sacsma_snow17.py:292`:

```python
mf = math.sin(dayn * 2.0 * 3.1416 / 366.0) * diff * 0.5 + (mfmax + mfmin) * 0.5
```

with `diff = mfmax - mfmin` and `dayn` counted from 21 March, so `mf` peaks at the
summer solstice and bottoms at the winter solstice (Anderson 2006, NOAA NWS,
*Snow Accumulation and Ablation Model — SNOW-17*). At the repo's own defaults
(`ht_adapter.py:67`, `mfmax` 1.2 and `mfmin` 0.4 per 6 h, i.e. 4.8 and
1.6 mm/degC/day at a daily step):

| melt window | day-of-year | mean `mf` | range |
| --- | --- | --- | --- |
| cycle 1, 30 Apr - 28 Jun | 121-180 | **4.62** mm/degC/day | 4.21 - 4.80 (at `MFMAX`) |
| cycle 2, 27 Oct - 25 Dec | 301-360 | **1.81** mm/degC/day | 1.60 - 2.25 (at `MFMIN`) |

Cycle 2 is melted with a coefficient 2.56x smaller while holding 2.75x more snow,
so it is about 7x harder than cycle 1. That is the whole effect. The sweep shows
it cleanly — cycle 1 never leaves any snow, cycle 2 is the only one that ever
approaches the limit:

```
sacsma_snow17, 200 seeds, as submitted        FAILED 2/200  (seeds 32, 179)
  melt_fraction cycle 1:  med 0.0000  p99 0.0000  max 0.0000     (limit 0.10)
  melt_fraction cycle 2:  med 0.0000  p99 0.1017  max 0.1164     (limit 0.10)
  peak_fraction cycle 1:  min 0.9117  med 0.9664                 (limit 0.50)
  peak_fraction cycle 2:  min 0.8135  med 0.9665                 (limit 0.50)

reference_bucket, 200 seeds, as submitted     FAILED 0/200
  melt_fraction cycles 1 and 2: 0.0000 everywhere
```

Cycle 1 leaves no snow on any of 200 seeds. Every failure, and every near miss,
is cycle 2.

`reference_bucket` never fails because it melts on the static
`degree_day_factor_mm_per_C_day` of 3.2, with no day-of-year term at all. So of
the two `must_pass` baselines, only Snow-17 can see this, and the bucket will go
on passing however the calendar is arranged — which is worth keeping in mind when
reading the gate as evidence.

A 1% failure rate is also close to invisible to five seeds: the chance of catching
it is 1 - 0.99^5, about 5%.

A seasonally varying melt factor is the norm rather than a Snow-17 quirk — it is
the standard temperature-index formulation (Hock 2003, *J. Hydrol.* 282, 104-115) —
and two models already in the repository carry it, as their own documentation says:

- `models/cwatm/ht_adapter.py`: "the melt coefficient follows a northern-hemisphere
  sine of the day of year, and extra ice melt acts on snow between days 166 and 259".
- `models/lisflood/README.md`, under the heading **"Snow is more than the degree-day
  factor"**: "a seasonal melt coefficient of ±0.5 mm/degC/day at `SnowSeasonAdj = 1`;
  a summer ice-melt term on any pack left between days 165 and 257", with the
  latitude setting "the hemisphere of LISFLOOD's seasonal snowmelt coefficient".

Both ice-melt windows — days 166-259 and 165-257 — sit entirely inside cycle 2's
*accumulation* stage (days 181-240), where the probe is asking for 405 mm of snow
to pile up at -10 C. `latitude_deg` is 44.0, so the northern-hemisphere branch is
the one that runs. Both models will meet this the moment they emit `snm`.

Snowfall at -10 C through July and August is incoherent forcing for any model that
reads day of year for radiation, albedo or melt factor, independently of whether
it happens to trip the criterion.

**Fix I tested.** Give each cycle a full year so both sit at the same phase, and
start the record in autumn so melt falls in spring: `STAGE_DAYS` 60 -> 120 and
`pd.date_range("1999-09-01", ...)`, with `period_days`/`min_window_days` 360 -> 720.

```
accumulation  1999-09-01 -> 1999-12-29   doy 244-363
storage       1999-12-30 -> 2000-04-27   doy 364-118
melt          2000-04-28 -> 2000-08-25   doy 119-238
accumulation  2000-08-26 -> 2000-12-23   doy 239-358
storage       2000-12-24 -> 2001-04-22   doy 359-112
melt          2001-04-23 -> 2001-08-20   doy 113-232
```

Seed 32 passes, and the gate still separates all four baselines with
`reference_snow_bypass` and `reference_snowless` tripping their pinned criteria:

```
sacsma_snow17, 200 seeds, year-aligned        FAILED 0/200
  melt_fraction cycle 1:  med 0.0000  p99 0.0000  max 0.0000     (limit 0.10)
  melt_fraction cycle 2:  med 0.0000  p99 0.0000  max 0.0000     (limit 0.10)
  peak_fraction cycle 1:  min 0.9242  med 0.9641                 (limit 0.50)
  peak_fraction cycle 2:  min 0.9084  med 0.9663                 (limit 0.50)

reference_bucket, 200 seeds, year-aligned     FAILED 0/200

ht gate --probe mass/snowpack-mass-closure
  must_pass  reference_bucket        closure=pass; state_bounds=pass; snowpack_response=pass
  must_pass  sacsma_snow17           closure=pass; state_bounds=pass; snowpack_response=pass
  must_fail  reference_snow_bypass   expected `closure`, tripped ['closure']
  must_fail  reference_snowless      expected `snowpack_response`, tripped ['snowpack_response']
  GATE PASSED
```

The two cycles also become statistically alike, which they are not at present:
`peak_fraction` minima go from 0.9117 / 0.8135 to 0.9242 / 0.9084. Two cycles
that are meant to be replicates of one another should not be separable.

Other routes exist — one cycle per run, or a looser `max_melt_fraction` — but
they either lose the cross-cycle cancellation check or weaken the criterion,
and neither fixes the midsummer snowfall.

## 2. `snowpack_response` ignores the timestep

```python
pr = np.asarray(w.forcing[precipitation], dtype=float)
...
total_pr = float(pr[accumulation_start:storage_end].sum())
```

`pr` is a rate in mm/day. Summing it treats one step as one day. `Window.volume()`
exists for exactly this conversion and `closure` uses it two files over, and
`make_window`'s own docstring says per-step depths "have to be integrated with
its own dt".

At `PT1D` dt is 1, so nothing is visibly wrong today. At any other step
`total_pr` is inflated by `1/dt` and `peak_fraction` collapses by the same
factor. Same physical snowpack, 100 mm accumulating, held, then fully melted:

```
PT1D   dt=1.00000 d   peak_fraction=1.0000   -> pass
PT1H   dt=0.04167 d   peak_fraction=0.0417   -> fail  (0.0417 < 0.5)
PT15M  dt=0.01042 d   peak_fraction=0.0104   -> fail  (0.0104 < 0.5)
```

The probe pins `PT1D`, so this does not fail anything in this PR. But
`snowpack_response` is registered in the shared criteria library and takes its
state, precipitation and labels as parameters, so it is plainly meant for reuse,
and the first sub-daily snow probe will silently mis-score every model.
`total_pr = float(w.volume(w.forcing[precipitation])[accumulation_start:storage_end].sum())`
is the whole fix. `max_melt_fraction` is a ratio of two storages and is unaffected.

## On the canopy point in the automated review

Worth supporting with the code, because it is the one issue here that cannot be
caught by adding seeds. The probe's control volume is entered by the forcing's
`pr`, but `pr` enters the *catchment*, not the snow module. In `reference_bucket`
the canopy sits downstream of the snowpack:

```python
snowfall = pr if tas < t_snow else 0.0
swe += snowfall                      # snowfall is never intercepted
...
water_in = rain + melt
intercepted = min(canopy_cap - canopy, water_in)   # canopy sees only what leaves the pack
```

So in both `must_pass` baselines the canopy can never hold intercepted snowfall,
and the residual in question is structurally unreachable — which is exactly why
the gate passes. A model that intercepts snow in the canopy *above* the snow
module reports a residual equal to that store's change, for correct bookkeeping.

`closure`'s new explicit `states` list makes the cheap fix available —
`states: [snw, canopy]` encloses the canopy in the control volume, and it is
sound for this case specifically because every drop of precipitation falls at
tas <= -3 C, so anything the canopy holds is intercepted snow. I have not
claimed it works, because the gate cannot demonstrate it either way for the
reason above; it would need a canopy-intercepting model to confirm.

## Smaller

**No unit tests.** This adds a criterion to the shared library, three new
parameters to `closure` (`optional_sinks`, explicit `states`, `segment_column`)
and two reference models, and the only change under `tests/` is one line in
`test_report_detail.py`. The repo has 36 test modules and every recent probe
brought at least one. A single test of `snowpack_response` at `PT1H` would have
caught point 2 on the first run, and the per-segment storage indexing in
`closure` (`storage[start-1]` vs `storage_initial` at `start == 0`) is worth
pinning too — it is correct, but it is the kind of off-by-one the module's own
docstring warns about.

**`snm` is not marked as an internal flux.** The AGENTS.md row says what `snm`
is but not what may count it. Every other ambiguous entry in that table is
explicit: `sbl` is "a component of `evspsbl`, never an addition"; `gw_boundary`
names which budget credits it. `snm` is water crossing an *internal* boundary and
must never be added to a catchment outflow budget beside `mrro`, or meltwater is
counted twice. Nothing in the code does that today — `closure` defaults to
`["evspsbl", "mrro"]` — but the table is what a model author reads.

**`reference_snow_bypass`'s comments describe a different model.**

```python
BYPASS_FRACTION = 0.30  # fraction of snowfall that bypasses the catchment and vanishes
...
# Make some SWEs disappear
bypass = BYPASS_FRACTION * snowfall
...
# The disappearing water enters the soil directly.
soil += throughfall + bypass
```

Nothing vanishes and nothing leaves the catchment — the bypass lands in the soil,
which is the entire point of the model and what the README and the PR description
correctly say. That distinction is what separates it from `reference_leaky`. For a
reference model the comment is the specification, so it is worth making it say
that the catchment balance is preserved and only the snowpack's is broken.

Also `COLUMNS` lists `"snm"` twice, which writes the column twice per row.

**`spec.py` is mostly reformatting.** The functional change is three lines: `snm`
in `FLUX_VARS`, its unit, and the two trusted models. The other ~50 lines re-wrap
`required_vars`, `spinup_steps_for`, `n_steps_for`, `load_probe`, `_check_variants`
and `load_model` without changing behaviour. `spec.py` is open in #73 and #83 too,
and this will conflict with both for no gain. (`closure.py` line 22 also has
`reported_states,segments,` with the space missing.)

## Coverage

Nine of the ten archived rows are `N/A INCOMPLETE — does not report snm`. The
only model that can answer the probe is `sacsma_snow17`, and only because this PR
added the column to it. That is the normal cost of a new contract variable and not
a reason to hold the probe — but it is worth a follow-up issue, because the gap is
smaller than it looks. Eight of the ten packaged models already report both `snw`
and `sbl`:

| model | `snw` | `sbl` | `snm` | meltwater already available? |
| cwatm | yes | yes | — | degree-day melt with `SnowMeltCoef` |
| dhbv2 | yes | yes | — | tracks `MELTWATER` explicitly; `snw = SNOWPACK + MELTWATER` |
| flex_lumped | yes | yes | — | |
| flex_topo | yes | yes | — | |
| lisflood | yes | yes | — | degree-day melt with `SnowMeltCoef` |
| summa | yes | yes | — | reads `scalarSfcMeltPond`; `iLayerLiqFluxSnow` at the pack base |
| wflow_sbm | yes | yes | — | HBV-style snow routine |
| sacsma_snow17 | yes | yes | **yes** | |
| google_flood_forecast | — | — | — | no snow state; out of scope |
| modflow6 | — | — | — | groundwater only; out of scope |

Adding `snm` to those seven takes the probe from one answerable model to eight.

One thing that would help the write-up: the discrimination is currently
demonstrated only at the extremes. Both `must_pass` models close to ~1e-13, and
the two failures leak 30% and 40% against a 5% threshold — six to eight times
over. Nothing exercises the criterion near its limit, so the probe does not yet
report what size of leak it can actually resolve.

## Checked and correct

- `_regime` is withheld from the model; the delivered forcing is `time,pr,tas,pet`.
- `min_window_days: 360` is a floor, and the harness already errors if a window
  drops a labelled stretch, so a cycle can never be cut.
- Segment storage indexing (`storage[start-1]`, `storage_initial` at `start == 0`)
  is right, and the absolute-residual sum does prevent inter-stage cancellation.
- Treating a missing `sbl` as zero, so unreported sublimation shows up as a
  residual rather than being inferred, is the right call and is documented.
- `min_peak_fraction: 0.5` has comfortable margin: observed peak fractions were
  0.91-1.00 across 200 seeds for Snow-17 and 1.00 for the bucket.
- The probe independently catches `reference_sublimating` at `closure` = 0.4000,
  exactly the 40% of snowfall that model is built to lose without reporting. It
  is not one of the declared baselines, so that is a genuine check that the
  criterion measures what it says it measures.

jinlong-hu:
@Barbhuiya12  Thank you for the detailed review and for running the 200-seed sweep.

I’ll address the two reproducible issues first:
1. align the synthetic snow cycles with the seasonal calendar;
2. make `snowpack_response` timestep-aware using integrated precipitation.

I’ll also clean up the duplicate `snm` entries, add targeted unit tests, and update the missing documentation noted in the review.

I’ll push the revision to this PR and follow up here with the updated validation results.

jinlong-hu:
@Barbhuiya12 Thanks again for the detailed review. I have pushed the revision and merged the latest `upstream/main` into this branch.

The main changes are:

- aligned the synthetic case seasonally using two 360-day cycles with 120-day stages. The generator begins on 1999-08-31 with one cold, dry spinup day, so the first scored accumulation stage begins on 1999-09-01;
- made `snowpack_response` timestep-aware using integrated precipitation via `Window.volume()`;
- reused `storage_at()` for segmented storage boundaries and exposed per-segment residuals in diagnostics;
- added targeted regression tests for segmented closure, spinup/storage boundaries, optional sinks, explicit states, the 5% closure threshold, PT1D/PT1H invariance, and malformed snow cycles;
- removed duplicate `snm` entries and corrected the `reference_snow_bypass` comments;
- clarified `snm` as an internal snow-module flux;
- updated the probe/reference-model documentation and archived results;
- removed the unrelated formatting changes from `spec.py`.

On the canopy/control-volume point, I kept the closure storage explicitly scoped to `states: [snw]` rather than changing it to `[snw, canopy]`. In the current `reference_bucket`, canopy interception is downstream of the snow-module `snm` boundary: snowfall enters `snw`, melt/rain leaves the snow module as `snm`, and only then can that liquid water enter canopy storage. Including `canopy` while still treating `snm` as an external sink would therefore mix control volumes and count storage downstream of a boundary that has already been crossed. Instead, the synthetic case sets `canopy_capacity_mm = 0`, and the documentation now makes the control-volume assumption explicit for models with interception upstream of the snow module.

I also went through the automated review comments on the previous revision. The concrete issues raised there are covered by this update: the duplicate `snm` output was removed, per-segment residuals are now exposed in diagnostics, the first-row storage issue is avoided with an explicit spinup day, `snowpack_response` now integrates precipitation with `Window.volume()`, segmented storage uses `storage_at()`, and the missing regression tests and documentation entries were added.

Validation after merging the latest `upstream/main`:

- `pytest -q tests/test_snowpack_mass_closure.py`: passed
- `ht validate`: 30 probe(s), 66 model(s), suite 0.1.0
- `ht gate --probe mass/snowpack-mass-closure`: passed
- full `pytest -q`: 1105 passed, 8 skipped
- full `ht gate`: passed
- `sacsma_snow17` seeds 32 and 179: both pass
- 200-seed sweep:
  - `sacsma_snow17`: 0/200 failures
  - `reference_bucket`: 0/200 failures

I also added a regression test around the 5% closure threshold so that near-threshold discrimination is now explicitly pinned.

For the broader `snm` coverage, I agree that extending the probe to additional packaged snow models would be useful. I kept those changes outside this PR to keep the current issue minimally invasive: the appropriate `snm` mapping is model-specific and needs to be checked against each model's actual snow-module control volume and internal diagnostics. I plan to treat expanded `snm` coverage as follow-up work rather than broadening this probe PR further.

Barbhuiya12 (dismissed):
Re-reviewed at `4e7ebe1`. Every point from my review is answered, and two of them
are answered better than I put them. `ht validate` (30 probes, 66 models) and the
probe's gate pass here.

## The calendar

`STAGE_DAYS` 60 -> 120 with the record starting 1999-08-31 and one spinup day, so
each cycle occupies a year at the same phase and both melt stages fall in spring.
Adding the spinup day is better than what I suggested: I had kept `pr[0] = 0.0` to
protect the opening state, and a real spinup row gives the harness a proper
reference instead of relying on the first row being dry.

Over seeds 0-199:

```
model                    role        n     wrong   melt_fraction max   peak_fraction min
reference_bucket         must_pass  200        0   0.0000  (limit 0.10)   1.0000  (limit 0.50)
sacsma_snow17            must_pass  200        0   0.0000  (limit 0.10)   0.9084  (limit 0.50)
reference_snow_bypass    must_fail  200        0   0.0000                 0.7000
reference_snowless       must_fail  200        0   0.0000                 0.0000
```

Seed 32 was the one that failed before at a cycle-2 melt fraction of 0.1164. The
melt fraction is now **0.0000 on all 400 must-pass runs** — not merely under the
limit but at the floor, on both cycles. That is the shape of a fix that removed
the mechanism rather than moved the margin: with both melt windows in spring,
Snow-17's seasonal melt factor is near `MFMAX` in each, so no draw leaves a pack
behind. The peak fractions are also symmetric between cycles now, which they were
not.

## The timestep

`pr = w.volume(w.forcing[precipitation])` — the one-line fix, and
`test_snowpack_response_is_timestep_invariant` pins it with the right experiment:
100 mm expressed as 100 mm/day for a day and as 2400 mm/day for an hour, asserting
the same `peak_fraction` of 1.0 from both.

I checked that the test can actually fail, because a regression test that cannot
is worse than none — I shipped two of those on my own probe earlier in this
repository. Reverting the one line and running it:

```
FAILED tests/test_snowpack_mass_closure.py::test_snowpack_response_is_timestep_invariant
  assert daily_result.status == hourly_result.status
  +  fail
```

It fails on the old code. Good.

## The rest

- **Tests.** 431 lines where there was one changed line. Beyond the two above, it
  pins the per-segment storage indexing I flagged as the off-by-one most likely to
  go unnoticed — `test_first_segment_uses_spinup_storage` and
  `test_segment_storage_starts_from_previous_row` — and
  `test_closure_resolves_leaks_around_the_five_percent_limit` answers the separate
  point that nothing exercised the criterion near its threshold, since both gate
  failures sat six to eight times past it.
- **The `snm` contract row** now ends "an internal flux across the snow-module
  boundary, never an additional catchment outflow beside `mrro`", which is the
  clause every other ambiguous variable in that table carries.
- **`reference_snow_bypass`** now says "The bypassed snowfall enters the soil
  directly, preserving catchment mass" instead of describing water that vanishes
  from the catchment. That is what the code does and what separates it from
  `reference_leaky`. The duplicate `"snm"` in `COLUMNS` is gone.
- **`spec.py`** is back to its four functional lines. The diff no longer collides
  with the other branches that touch it.
- **`canopy_capacity_mm: 0.0`** removes the control-volume question rather than
  arguing it. The automated review's point — that a model intercepting snow above
  the snow module reports a residual equal to that store — was real and I
  confirmed it was structurally invisible to both baselines. Zeroing the canopy in
  the case means no model can hold intercepted snow at all, so the probe measures
  the snowpack budget and nothing else. Cleaner than adding `canopy` to `states`,
  which is what I had suggested was available.

## Coverage, unchanged and still worth an issue

Nine of ten packaged models remain N/A for want of `snm`, which is the ordinary
cost of a contract variable and not this branch's to fix. The point I would still
like recorded somewhere is that eight of the ten already report both `snw` and
`sbl`, so they are one column short, and the meltwater is already an explicit
quantity in each: `raim` in Snow-17 is exactly this variable, and dHBV2 already
tracks `MELTWATER` separately inside `snw`. A follow-up issue naming those seven
would turn the probe from one answerable model into eight.

## Approving

800 runs over 200 seeds, nothing on the wrong side, and the failure that
motivated the change is gone at the floor rather than merely under the limit.
The regression test for the timestep bug fails on the old code, which is the only
thing that makes it a regression test.

chrimerss:
Merged `main` into the branch as `3f27dbc` and recounted the docs, so this is
now mergeable against `3f6d582`.

All three conflicts with #94 were union-resolvable. The only judgement call was
the site flowchart: I followed #94's approach, where the MASS pillar keeps its
250-high rect like ENERGY and MOMENTUM and its twenty-one entries are re-spaced
at 8.60, rather than growing the rect to 270 — that left the three pillars
unequal and the mass one ending two pixels above the step boxes. `snowpack` sits
between `withdrawal` and `gw exchange`, matching the README order. Also in that
commit: the README recount to thirty-one probes, twenty-one of them mass, with a
clause recording that this probe can score only `sacsma_snow17` among the
physical models; `spec.py`'s trailing newline restored; and the copy-pasted
"Existing adapter mappings" comment replaced in both new `model.yaml`s.

Verified at `3f27dbc`: `ht validate` 31 probes / 72 models, 1138 tests, the full
`ht gate`, the probe's own gate separating all four baselines, a 200-seed sweep
of both `must_pass` baselines (0 wrong; `melt_fraction` 0.0000 on all 400 runs,
`peak_fraction` min 0.9084), and the archived `sacsma_snow17` row reproducing
byte-for-byte apart from its date. All nine `N/A — does not report snm` reasons
match what `missing_for` returns. CI run 35528284803 is green on both jobs.

Five mutations of the new code — dropping `abs()` on the segment residual,
swapping `storage_at` for `storage_initial`, dropping `w.volume()`, always
skipping optional sinks, ignoring the explicit `states` list — are each caught by
the new tests, so these are real regression tests rather than tests that happen
to pass. A parametric bypass sweep puts the detection edge exactly at the 5%
rule: 1–4% passes 30/30 seeds, 6% fails 30/30, and the reported ratio tracks the
leak fraction linearly.

Two things I would record as follow-up issues rather than hold this branch for.

**`min_peak_fraction: 0.5` has a lot of headroom.** A model that reports 49% of
every snowfall as `snm` — liquid water delivered to the ground at −6 °C — and
stores the rest closes its snowpack budget to machine precision, scores a
`peak_fraction` of 0.51, and passes all three criteria on 30/30 seeds. That is
not a conservation violation, so it may be correctly out of the probe's scope.
But Snow-17's observed minimum over 200 seeds is 0.9084, so 0.7–0.75 would halve
what a *reported* liquid bypass can hide at no cost to either baseline.

**The canopy scope note is prose only.** `requires: static: [canopy_capacity_mm]`
would turn it into an eligibility rule and make a model with interception above
its snow module N/A instead of VIOLATION. But `sacsma_snow17` declares neither
`needs_static` nor `uses_static`, so it would go N/A and the probe would lose a
`must_pass` baseline. Worth settling before the first model with snow-canopy
interception starts emitting `snm`.

Plus @Barbhuiya12's coverage point, which deserves to be written down somewhere:
eight of the ten packaged models already report both `snw` and `sbl` and are one
column short of being scoreable, with the meltwater already an explicit quantity
in each.

One apology owed: pushing `3f27dbc` dismissed @Barbhuiya12's approval. The only
change since the reviewed `4e7ebe1` is the merge resolution and the doc recount
above — no probe, criterion, reference model or test was touched — so the
800-run evidence in that review still stands.

Barbhuiya12 (approved):
Re-reviewed at `3f27dbc`. No apology needed for the dismissal — a merge that
lands #94 is exactly the kind of push that should clear approvals.

I checked the claim rather than taking it: the only changes to anything this
probe owns since `4e7ebe1` are two comment lines in
`reference_snow_bypass/model.yaml` and `reference_snowless/model.yaml`. No
probe.yaml, generator, criterion, reference adapter or test was touched. `ht
validate` gives 31 probes and 72 models, and the gate separates as declared:

```
must_pass  reference_bucket        closure=pass; state_bounds=pass; snowpack_response=pass
must_pass  sacsma_snow17           closure=pass; state_bounds=pass; snowpack_response=pass
must_fail  reference_snow_bypass   expected `closure`, tripped ['closure']
must_fail  reference_snowless      expected `snowpack_response`, tripped ['snowpack_response']
```

So the 800-run evidence from the last review carries over and I am re-approving.

## `min_peak_fraction`: 0.70 is the one value that cannot be used

@chrimerss — I have the numbers for this, and they rule out the bottom of the
range you suggested. `reference_snow_bypass` keeps `1 - BYPASS_FRACTION` of every
snowfall by construction, so its `peak_fraction` is not a distribution at all:

```
model                    role        min      med      max     spread
reference_bucket         must_pass  1.0000   1.0000   1.0000   constant by construction
sacsma_snow17            must_pass  0.9346   0.9648   0.9816   0.047 wide  (0.9084 over 200 seeds)
reference_snow_bypass    must_fail  0.7000   0.7000   0.7000   constant by construction
reference_snowless       must_fail  0.0000   0.0000   0.0000   constant by construction
```

The usable window is therefore bounded below by the cheat at exactly 0.70 and
above by Snow-17 at 0.9084, and **0.70 itself is the worst point in it**:

```
reference_snow_bypass, seed 7
  cycle_1_peak_fraction = 0.7
  cycle_2_peak_fraction = 0.6999999999999998
```

Two cycles of identical construction land on opposite sides of 0.70 because the
second accumulates two ULP low. At `min_peak_fraction: 0.70` the verdict is
decided by the last bits of a double rather than by anything physical. Your 0.75
is safely inside the window; 0.70 is not a tight choice, it is an undefined one.

## Tightening it costs the bypass control its isolation, and the gate will not tell you

This is the part I would not have expected. At every candidate above 0.70:

```
min_peak_fraction 0.70   reference_snow_bypass   expected `closure`, tripped ['closure', 'snowpack_response']   GATE PASSED
min_peak_fraction 0.75   reference_snow_bypass   expected `closure`, tripped ['closure', 'snowpack_response']   GATE PASSED
min_peak_fraction 0.80   reference_snow_bypass   expected `closure`, tripped ['closure', 'snowpack_response']   GATE PASSED
```

The gate stays green throughout, because it only asks that the declared criterion
be *among* those tripped. Nothing in the suite notices that the control stopped
being a control for one thing.

That matters here more than it usually would. `reference_snow_bypass` exists to
show that `closure` catches water leaving the snowpack **while the whole-catchment
budget still closes** — it is the model that distinguishes an internal leak from
`reference_leaky`. Once it also trips `snowpack_response`, a reader cannot tell
which criterion the probe is actually demonstrating, and a later regression in
`closure` alone would still leave the gate green on this model.

**A fix that keeps both, which I tested:** raise the threshold and move the cheat
out of its way at the same time — `min_peak_fraction: 0.80` with
`BYPASS_FRACTION` 0.30 -> 0.15, giving the bypass a `peak_fraction` of 0.85,
above the new floor and below Snow-17's 0.9084.

```
min_peak_fraction 0.80 + BYPASS_FRACTION 0.15
  must_pass  reference_bucket        all pass
  must_pass  sacsma_snow17           all pass
  must_fail  reference_snow_bypass   expected `closure`, tripped ['closure']
  must_fail  reference_snowless      expected `snowpack_response`, tripped ['snowpack_response']
  GATE PASSED
```

The cost is that the leak the control declares falls from 30% to 15% of snowfall,
still three times the 5% limit, so `closure` keeps a clear margin. If that trade
is not worth it, the alternative is to keep `0.5` and write the liquid-bypass case
into the README as a disclosed scope limit — which is what the probe already does
for the cases it cannot see, and is honest.

Either way I would not take 0.70.

## The canopy note

Worth recording that the case has already closed this: `generate.py` sets
`canopy_capacity_mm: 0.0`, so no model can hold intercepted snow on this probe
whatever its interception scheme. The prose note describes a limit that the
forcing removes. An eligibility rule via `requires.static` would only start to
matter if a future revision reinstated a canopy — and you are right that it would
cost `sacsma_snow17`, which declares neither `needs_static` nor `uses_static` and
would go N/A. I would leave it as prose and revisit only if the capacity ever
becomes non-zero.

## Approving

The merge is inert with respect to everything this probe owns, the gate holds, and
the threshold question above is a refinement rather than a defect in what is here.

chrimerss (approved):
Second mass-pool approval. Reviewed at 3f27dbc: gate, 1138 tests, 200-seed sweep and archive rows verified; threshold refinement tracked as a follow-up.

PR #103, opened 2026-09-18
momentum/froude-regime: score a reported stage against the section that must carry it

### LINKED ISSUE #72: [PROBE] momentum/froude-regime
# Proposal: `momentum/froude-regime`

**Authors**
```
Yuanhang Liu, Independent Researcher, Changchun, China, ORCID 0009-0005-9633-6004
```

**Conservation law**
momentum

**Residual equation**

For an open-channel reach on a mild slope, the Froude number must remain subcritical:

```
Fr = v / sqrt(g * D) < 1
```

where
- `v` is the cross-sectionally averaged velocity (m/s),
- `g = 9.81 m/s²`,
- `D` is the hydraulic depth (m).

For a wide rectangular reach of width `w` and water depth `d`, with discharge `Q` and stage `h`:

```
Q = v * w * d
D ≈ d = h - z_b   (z_b = bed elevation)
Fr = Q / (w * d * sqrt(g * d)) = Q / (w * d^1.5 * sqrt(g))
```

The probe declares a reach as mild-sloped (`slope < slope_critical`, supplied in static attributes) and asks that every scored step satisfies `Fr <= 1` within a small tolerance.

**Discharge source**

`Q` [m³ s⁻¹] is taken from `dis` when the model reports it. If the model reports only
`mrro` [mm day⁻¹], the probe converts it using a prescribed reach/catchment area supplied as a
static attribute (`drainage_area_km2`):

```
Q [m3 s-1] = mrro [mm day-1] * 1e-3 [m/mm] * drainage_area_km2 * 1e6 [m2/km2] / 86400 [s/day]
```

This conversion is the same one used by other flux-based momentum criteria, so a model that reports
only `mrro` can be scored. `stage` comes from the diagnostic channel introduced by PR #70
(`momentum/stage-discharge-monotonic`); `d` is `stage - bed_elevation_m` with a 1 cm floor.

**Dependency on #70**

This probe reads `stage`, so it is sequenced **after PR #70 merges**. Until then no model emits
`stage` and the probe cannot be scored. Implementation will not start before #70 lands.

**Tolerance and its denominator**

The tolerance is an absolute one on a dimensionless number: `Fr <= 1.0 + 0.05`. The 5 % margin accounts for numerical noise in the velocity estimate and the approximation of the cross-section as rectangular. Steps where `d` is below a 1 cm floor are skipped because the Froude number is ill-defined for a dry or nearly dry channel. The scoring window is chosen to contain at least one day with meaningful flow.

**How would an unphysical model fail this?**

A model can report a water budget and hydrograph that look plausible while computing a velocity/depth pair that is dynamically impossible for the prescribed mild slope. The discriminating reference model is `reference_supercritical_flat`: it routes the same discharge but computes velocity from a fixed, too-small cross-section (e.g. it treats the reach as 10 cm deep regardless of stage), so `v` is large enough to push `Fr > 1` whenever the hydrograph rises. It conserves mass, and it may even pass `routing_conservation`, but it violates the momentum budget that a mild reach cannot run supercritical.

A second failure mode is a model that inverts depth and velocity: it reports a deep, slow flood as shallow and fast because its stage-discharge rating is not hydraulically consistent. `stage-discharge-monotonic` will not catch this because the rating can still be monotonic; `froude-regime` catches the physically impossible velocity.

**How is the case generated?**

Synthetic daily weather over two years with a short spinup and a single multi-day flood in the second year. Static attributes prescribe a wide, mild-sloped rectangular reach (`slope` << critical slope, `width_m`, `bed_elevation_m`, `drainage_area_km2`). The forcing is temperate rain-dominated weather with a storm sequence that fills and drains the channel once, so the model must report both low-flow and high-flow regimes.

Because the case is on the synthetic track, the reach geometry is known and fixed; the model does not need to infer it. The probe reads the model-reported `stage` diagnostic and `dis` (or `mrro` + `drainage_area_km2`) and computes `Fr` directly from the prescribed geometry.

**Expected outcomes for the required baselines**

| Model | Reports `stage`? | Reports `Q`? | Expected | Why |
| `reference_bucket` | after #70 | `mrro` only | **N/A (INCOMPLETE)** | no channel routing; `channel` store identically zero, no reach discharge geometry |
| `flex_lumped` | after #70 | `dis` + `mrro` | **PASS** | routes through channel store; Manning-consistent stage on a mild reach |
| `flex_topo` | after #70 | `dis` + `mrro` | **PASS** | same, channel-only stage from Manning |
| `sacsma_snow17` | after #70 | `dis` + `mrro` | **PASS** | routes to channel; subcritical by construction |
| `reference_supercritical_flat` (new) | yes | `dis` | **FAIL** | fixed 10 cm depth → `Fr > 1` on the rising limb |

Reference models see the full record. The must-pass gate applies only to models that report both
`stage` and a channel discharge; a model that cannot report `stage` is scored INCOMPLETE, which is honest.

**Will you build it?**

Yes — assign it to me.

kawh1111:
Closes #72.

## The physics

A stage and a discharge are two readings of one cross-section, and open-channel
flow caps how fast that section can deliver water:

```
Fr = Q / (w * d**1.5 * sqrt(g)) <= 1
```

Past a Froude number of one the water outruns the wave that carries it, and on
a mild slope that regime cannot be sustained — the reach would have to be steep
enough, or short enough, for gravity to stop setting the speed. Fr above one is
not a small error in a budget; it is a different regime.

`Q` comes from `dis`, or from `mrro` over the catchment area when a model
reports only a depth rate. `d` is the `stage` diagnostic, which the contract
defines as a depth above the reach bed. `w` is the width the **case** declares,
so every model is judged in the same channel rather than in one of its own
choosing; the slope and the roughness are the case's for the same reason, since
a reach is mild only relative to its roughness. Nothing is fitted.

## Why no existing criterion already reads it

`momentum/routing-conservation` bounds the water a reach holds.
`momentum/stage-discharge-monotonic` asks whether the gauge moves with the flow,
and whether the loop runs the right way. Neither asks whether the depth a model
reports is one the declared section could carry that flow through. No budget
asks either: a stage is not a store, so nothing is differenced over it, and a
model can conserve every millimetre of water while reporting a depth that has
nothing to do with its discharge.

That is asserted in the proposal, so the PR carries a model built to test it
rather than a claim.

## The counterexample is the argument

`reference_shallow_rating` is the exact bucket with one thing changed: its gauge
applies Manning's law to a section **five times wider** than the case declares,
which is what a rating curve reused from another reach looks like. Everything
else is the bucket: same water to the floating point, same runoff, `channel`
identically zero.

That gauge is not obviously broken, and that is the design requirement:

| Property | Value | Consequence |
| --- | --- | --- |
| `cv(stage)` | 0.56 – 0.94 across the gate seeds — **identical** to the honest gauge's | `non_degenerate` passes; no variability threshold can tell them apart |
| monotone in discharge | yes | `rating_monotonic` passes |
| single-valued rating (one state, no water in transit) | yes | `rating_loop` reads no loop, which is the honest reading for a bucket |
| water balance | exact | every `mass/*` probe passes |

Driven through every probe in the suite that can ask it anything, it passes
**all of them** and trips **one** criterion:

```
probes tripped: 1
  momentum/froude-regime: froude_subcritical=0.901
```

Separation on the probe's own gate seeds:

| | worst Fr | median Fr | supercritical share |
| --- | --- | --- | --- |
| `reference_bucket`, `flex_lumped`, `flex_topo`, `sacsma_snow17` (must pass) | 0.389 – 0.445 | 0.245 – 0.284 | 0 of 1460 steps |
| `reference_shallow_rating` (must fail) | 1.772 – 1.928 | 1.183 – 1.210 | 83.3% – 90.1% |

The honest gauges' worst step is 4× below the failing model's worst, and the
failing model is above the limit on most of the record rather than at one
marginal step.

## What is in it

- `src/hydroturing/criteria/froude.py` — `froude_subcritical`, scored as the
  share of scored steps that went supercritical rather than the worst step, so
  one numerical edge reads differently from a reach that is supercritical
  through every flood. A missing column, a case that declares no width and a
  case that declares no area all **fail** it rather than crashing it. Steps are
  scored only when they report a depth this section could hold: above the 1 cm
  floor, and below `max_depth_m` when a probe sets one.
- `AGENTS.md` — `stage` now states its datum: the water level a gauge would
  read, measured from the reach bed. Without it `stage` was readable as an
  absolute level, and an absolute level passes this criterion comfortably,
  because an offset only ever adds to the depth and the depth enters at the
  three-halves power.
- `models/reference_shallow_rating/` — the must-fail above.
- `tests/test_froude.py` — 24 counterexamples, including the wider-reach rating
  and a test that its variability is *identical* to the honest gauge's, so the
  argument above cannot rot silently. The `mrro` fallback and its 86400, the
  width read from the case, the depth floor and ceiling, both tolerance knobs,
  the missing columns, and this probe's `requires.static` (with
  `reference_flat_stage` recorded N/A rather than scored) are pinned too.
- Registered: `criteria/__init__.py`, `spec.py`'s `TRUSTED_SUBPROCESS_MODELS`,
  `tests/test_report_detail.py`'s HEADLINES.
- Docs synced: README (probe table, model table, `reference_shallow_rating`,
  Twenty-nine → Thirty, the three physical models' counts), CONTRIBUTORS,
  ROADMAP (`merged`), `models/sacsma_snow17/README.md`, `docs/writing-a-probe.md`
  (criteria and reference-model tables), and the site in three languages (probe
  row, model counts, the flowchart's fourth momentum dot, a new i18n key).
- `models/result.csv`: appended, one row per evaluated model. The three
  physical models pass; the seven submitted ones report no `stage` and are N/A
  before any container starts.

The branch carries `main` at `45784d5`, which is the commit #83 landed in, so the
counts below are the merged suite's rather than the pre-merge ones.

## Verification

- `ht validate`: **30 probes, 65 models** (19 mass, 7 energy, 4 momentum).
- `ht gate --probe momentum/froude-regime`: 4 must_pass + 1 must_fail separated
  on `froude_subcritical`.
- Full suite: **1116 passed, 8 skipped**; `tests/test_docs_in_sync.py` and
  `tests/test_report_detail.py` pass with no count correction needed.
- The ceiling, measured rather than argued: all four must-pass baselines score
  all 1460 steps with `too_deep_steps` 0, and their deepest reading is 3.99 m
  (`reference_bucket`, worst Fr 0.445 inverted through the section's
  normal-depth relation) against the 10 m ceiling — a factor of 2.5.
- The suite-wide sweep above, for the "caught by nothing else" claim.

## Limitations

In the probe's README: it reads the pair, not either reading against the truth —
a model that derives its stage from its own discharge through any normal-depth
relation stays subcritical whatever its discharge error, because `Fr` collapses
to a function of the depth; the cross-section is the wide-rectangular
approximation on the case's geometry, not the model's; steps at the depth floor
are not scored; and the datum is the contract's, which the criterion can only
enforce for a level that is *far* from it — a metre of offset is
indistinguishable from a deep reach by any value test, which is why the contract
has to say what the datum is rather than the criterion infer it.


kawh1111:
All nine findings are addressed in `df9072b`.

**`high` — the README hunk reverted #83.** Confirmed, and it was an old state being restored rather than anything the probe needs. The `modflow6` and `reference_exchange_exact` rows were gone and `reference_bucket`'s standing had regressed to "the five energy probes". I rebuilt the file from `main`'s copy at `45784d5` and re-applied only what this probe earns, so `git diff 45784d5 -- README.md` is now seven lines: the 29→30 probe count and three→four momentum, "twelve of the twenty-nine"→"thirty", "six more we want"→"five", the new probe row, the three per-model denominators this probe moves (`flex_lumped` 22→23, `flex_topo` and `sacsma_snow17` 21→22), the new reference row, and the Status line. Nothing else in the file differs from `main`.

**`medium` — the two tables in `docs/writing-a-probe.md`.** Both rows added: `froude_subcritical` to the criteria table (describing the limit and the scored-step floor the way its neighbours describe theirs) and `reference_shallow_rating` to the reference-model table.

**`medium` — no `requires.static`.** Declared as `static: [width_m]`, and the four must-pass baselines plus the must-fail reference now declare `uses_static: [width_m]` — every one of them reads that key for its normal-depth stage whenever the case supplies it, so the declaration is true of the adapters rather than a way around the check. Measured on the built branch: `ht run --model reference_flat_stage --probe momentum/froude-regime` reports `N/A (INCOMPATIBLE) — model does not declare that it consumes static width_m`, where it was previously scored. The gate still separates: four must-pass, and `must_fail reference_shallow_rating expected froude_subcritical, tripped ['froude_subcritical']`.

One deliberate deviation from your fix: I did **not** require `bed_elevation_m`. It is the datum the criterion subtracts from an absolute water level, and no adapter in the repository consumes it, so requiring it would make the four baselines and the must-fail declare a key they never read — the opposite of what this mechanism is for. The case declares it explicitly (at zero) and the criterion documents the zero default as the honest reading for a depth; the probe README now states plainly that a model reporting a level in its own datum is recorded N/A rather than failed.

**`medium` — the invented 18 m channel.** `FALLBACK_WIDTH_M` is gone. A case that declares no width is a statement about the case, so the criterion fails and names the missing key, exactly as the `area_km2` branch four lines into `_discharge` already did.

**`low` — the PASS message.** The share is reported in both branches now, and "stays subcritical on every scored step" is claimed only when `n_exceed == 0`. The lenient-limit path is asserted on the message, not just on status, so the sentence archived in `detail` cannot contradict the row again.

**`low` — the division by zero.** The guard is `n == 0 or n_scored == 0 or n_scored / n < min_scored_fraction`.

**`low` — the dropped "More".** Restored on the Status line.

**`low` — the geometry defaults.** `18.0`, `0.0015`, `0.035` — the bucket's — and the extra `manning_n` guard is dropped, so `RATING_WIDTH_FACTOR` is the only difference between the two models on every case, including the ones that declare no reach geometry.

Three tests added for the new branches: a case declaring no width, the floor turned off with nothing scored, and a lenient limit over a partly supercritical record. `tests` is 1112 passed, 8 skipped; `ht gate --probe momentum/froude-regime` passes. CI is green on `df9072b`: both `gate` and `container` passed.

chrimerss (changes requested):
Thanks @kawh1111. All nine points from the automated review of 1bc6f33 are fixed in df9072b, and the criterion itself has no correctness bug. Two of its physics assumptions are not yet written into the contract, though, so a future model that reports `stage` differently from the current adapters could be scored wrongly. Those two are the blockers.

## Verified at df9072b

- pytest: 1118 passed, 2 skipped. `ht validate`: 30 probes, 65 models. `ht gate --probe momentum/froude-regime` separates as declared, and CI (gate and container) is green.
- The `flex_lumped`, `flex_topo` and `sacsma_snow17` archive rows reproduce byte-for-byte with `--gate-seeds`.
- `reference_shallow_rating` run through the whole suite fails only `froude_subcritical` (21 of 22 scored), as the PR says.
- Over 100 random seeds, `reference_bucket` passes every time (worst Fr 0.41–0.46) and `reference_shallow_rating` fails every time (worst Fr 1.75–1.94, supercritical on 82–99% of scored steps). A Manning gauge in the declared section is subcritical by construction (Fr = 0.353·d^(1/6) here), so the margin does not depend on the seed.
- Sensitivity: a rating drawn for a section 3× too wide is caught on every gate seed; 2.5× is not (worst Fr 0.95–1.03).
- I also tried a gauge that spills onto a floodplain above the case's 2.5 m bankfull depth (100–400 m of floodplain, n 0.06–0.08). It passes with worst Fr ≤ 0.75, so the rectangular section is not a problem at these flows.

## Before merge

1. **Pin the datum of `stage`.**
   - AGENTS.md:169 defines `stage` only as "the water level a gauge in the reach would read", with no datum, and the probe does not require `bed_elevation_m`. The criterion subtracts the case's bed (zero) and treats what is left as depth.
   - I scored the bucket's own discharge against a Manning stage in the declared section, shifted to other datums, using `froude_subcritical` with the probe's parameters. For a model that declares `width_m`:
     - a gauge zero 0.3 m above the bed FAILs on every gate seed, with worst Fr ≈ 49 and 97–99% of steps supercritical;
     - a gauge zero 0.5 m above the bed FAILs as "dry or flat";
     - an absolute level (stage + 100 m) PASSes with worst Fr ≤ 0.004, and so does the 5×-wide rating with the same offset.
   - The README (line 123) says a model "reporting a water level in its own datum" is recorded N/A (INCOMPATIBLE). That is only true if the model also does not declare `width_m`.
   - Fix: either define `stage` in AGENTS.md's variable table as depth above the reach bed, in metres, or require `bed_elevation_m`. The first costs nothing today: every adapter that reports `stage` already reports a depth above the bed (a Manning depth, or `reference_flat_stage`'s constant), and `momentum/stage-discharge-monotonic` does not depend on the datum. Then correct the README sentence to match.
2. **Require `slope` and `manning_n` as well as `width_m`.**
   - A reach is mild only relative to its roughness: the critical slope is g·n²/d^(1/3). generate.py:7 makes the same point ("on a steep reach Fr > 1 is the expected regime and the criterion would be wrong to call it a violation").
   - A model that reads the declared width but uses its own roughness n = 0.012 reports a stage and discharge that are consistent with each other (Manning in the same 18 m section), and it FAILs on all three gate seeds (worst Fr 1.09–1.19) as a VIOLATION. The message says the two readings "are not two readings of the same cross-section", which is not true. An own slope of 0.01 fails the second gate seed (1.064). At n = 0.015 it still passes (0.89–0.97).
   - This is the same argument you applied to `width_m`: the verdict rests on inputs the model may never have read. Fix: `requires.static: [width_m, slope, manning_n]`, and add both keys to `uses_static` in the five manifests. Every one of those adapters already reads both keys in its normal-depth stage function, and `reference_bucket`'s new model.yaml comment already says so.
3. **Regenerate the seven submitted-model rows.**
   - `models/result.csv` lines 592–601 are dated 2026-09-17, before `requires.static` existed. At df9072b, `ht run --gate-seeds --csv` writes the detail `does not report stage; model does not declare that it consumes static width_m` for cwatm, dhbv2, google_flood_forecast, lisflood, summa, wflow_sbm and modflow6. The archive has `does not report stage`. The verdict and reason (N/A, INCOMPLETE) do not change.
   - The run returns N/A before any container starts. Please regenerate the rows after 1 and 2, since item 2 changes the detail again.

## Smaller items (fine as a follow-up)

- No test pins `requires_static` for this probe. `tests/test_radiation_probe.py` and `tests/test_routing_lag_consistency.py` both assert theirs, so dropping the `static:` line here would go unnoticed. A test asserting the declared keys, and that `reference_flat_stage` is N/A (INCOMPATIBLE) rather than scored, would close it.
- `site/index.html`: in all three languages, the `momentum/froude-regime` row keeps its place in the wanted block with a `merged` tag. AGENTS.md says it moves to the merged block. `momentum/stage-discharge-monotonic` was left the same way in #70, so both could move together.
- `models/sacsma_snow17/README.md:69`, the line under the one this PR edits, still says "the five energy probes" are N/A. On main it is six energy probes plus the groundwater-exchange probe. This was already stale before this PR.
- The PR description says 17 tests and 1109 passed; there are now 20 tests.

## Review status

No approvals yet. As the author you are excluded from reviewing this probe, so it needs two approvals from the rest of the momentum pool on the final head, plus a green acceptance gate.

Barbhuiya12 (commented):
Reviewed at `df9072b`. `ht validate` (30 probes, 65 models) and
`ht gate --probe momentum/froude-regime` pass here, separating as declared with
`reference_shallow_rating` tripping `froude_subcritical` alone.

The idea is a good one and the framing is the part I like most: scoring the pair
against each other rather than either reading against a truth, so a model that
derives stage from discharge through *any* normal-depth relation is consistent by
construction and the criterion still has something to say, because the celerity
depends on the depth and not on the flow. @chrimerss's `requires.static` items are
right and I have nothing to add to them.

One thing that is not in his list, because it only shows up if you feed the
criterion a stage the current adapters never produce.

## A gauge datum passes the negative control straight through

`AGENTS.md` defines `stage` as "the water level a gauge in the reach would read".
A gauge reads against its own datum, so an absolute level is a contract-compliant
reading of that sentence. `froude.py` handles the datum properly — it subtracts
`bed_elevation_m` — but that key is not in this probe's `STATIC`, so
`FALLBACK_BED_M = 0.0` applies and the reported stage is taken as the depth.

Taking `reference_shallow_rating`'s own output on gate seed 3 and adding a datum,
changing nothing else:

```
as run                   FAIL   98.5% of steps supercritical (1438/1460)  worst Fr 1.852  median 1.231
+ 0.5 m gauge datum      FAIL    0.1% of steps supercritical (1 of 1460)  worst Fr 1.147  median 0.099
+ 1.0 m gauge datum      PASS    0                                        worst Fr 0.798  median 0.041
+ 25  m gauge datum      PASS    0                                        worst Fr 0.021  median 0.000
```

One metre defeats the control. `Fr = Q / (w √g d^{3/2})`, so the depth enters at
the three-halves power and a datum offset is the cheapest possible way to look
subcritical. Nothing here is a bug in the criterion — the arithmetic is right at
every line — but the probe currently relies on every model reporting stage as a
depth above the bed, and that is an assumption about the contract rather than
something the contract says.

**This is about to stop being hypothetical.** #114 changes the shared adapters —
`reference_bucket`, `flex_lumped`, `flex_topo` — from `stage = depth` to
`stage = bed_elevation_m + depth`, and draws `bed_elevation_m` from
`uniform(25.0, 180.0)` metres in its own generator. Those are the same adapters
this probe's `must_pass` baselines use. They stay safe here only because this
probe's static omits the key and the adapters default it to zero, so the two PRs
happen not to collide. That is a narrow escape rather than a design, and @licm13
is asking the same question from the other side of #114: define the datum of
`stage` at contract level.

**The cheap guard is already in your own case.** `generate.py` declares
`bankfull_depth_m: 2.5` and `froude.py` never reads it. A derived depth of 26 m in
an 18 m-wide reach whose bankfull is 2.5 m is not a depth in this section, and the
criterion currently scores it as a very comfortable pass. A ceiling — some
multiple of bankfull, above which the reading is refused as not a depth rather
than scored — costs a few lines, uses a number the probe already publishes, and
turns the silent pass above into a loud N/A. It also protects the probe from
whatever the datum question settles on, since a model reporting a true gauge level
would then be told so instead of being congratulated.

There is a depth floor (`DEFAULT_MIN_DEPTH_M`, one centimetre, with the reasoning
written out) and no ceiling. The floor exists so a receding flow does not divide
by zero; the ceiling would exist for a larger reason, which is that the criterion
cannot otherwise tell a depth from an elevation.

## Smaller

`requires.static: [width_m, slope, manning_n]` is @chrimerss's item 2. If the
ceiling above goes in, `bankfull_depth_m` belongs in that list too, for exactly
the reason he gives for the other three: the verdict would otherwise rest on an
input the model never had to read.

The `vel` column in #95 is worth watching from here. Once a model can report the
section-averaged velocity, `Fr = v/√(gD)` can be formed from the model's own
velocity rather than from `Q/(w·d)`, which removes the rectangular-section
assumption from the numerator. Not for this PR — `vel` is not in `main` yet — but
this criterion is the one that would most directly benefit, and it is worth a line
in the README saying so, so the next author finds it.

## Checked and correct

- The pair-consistency framing, and the reason it still has teeth: a normal-depth
  stage is self-consistent by construction, and the Froude number is not a
  restatement of that construction because the celerity depends only on the depth.
- The tolerance of 0.05 on Fr is argued as covering the rectangular approximation
  and normal-depth rounding rather than a wrong velocity, which is the right thing
  for it to cover.
- No width fallback: refusing to invent a section rather than scoring every model
  in a channel no case declared. That is the same instinct the datum needs.
- The dry-reach exclusion, and scoring only steps that clear the floor for reasons
  other than the floor.

kawh1111:
Both reviews are addressed in `4cadf9c`.

## Before merge

**1. The datum of `stage` is now in the contract.** `AGENTS.md` defines it as the water level a gauge would read, **measured from the reach bed**. Following that through, the criterion stops subtracting a bed — `bed_key`, `FALLBACK_BED_M` and the `bed_elevation_m` diagnostic are gone, the case no longer declares one, and the README sentence you caught is corrected: a model reporting a level in its own datum is not recorded N/A, it is outside the contract. Your datum numbers are also why the ceiling below exists — with the datum merely *defined*, a model that ignores it still passes comfortably, and one metre of offset is enough to turn the must-fail into a pass.

This settles the same question the automated review raised on #114 as "the `stage` contract does not say which datum it is measured from", and that licm13 raised there as "the `stage` datum change needs to be resolved at the contract level". It settles it against `stage = bed_elevation_m + depth`, so #114's baseline change would need to come back out; `CV(stage)` in `momentum/stage-discharge-monotonic` is unaffected for a model that reports a compliant stage, which is licm13's second point. I have not touched #114.

**2. `requires.static: [width_m, slope, manning_n]`.** Your `n = 0.012` case is the argument and I reproduced its shape. All five baselines declare all three keys, and every one of them reads all three in its normal-depth stage function. A test pins the tuple and records `reference_flat_stage` as `N/A (INCOMPATIBLE)` with the message naming all three keys, so dropping the line now fails the build.

**3. The seven rows are regenerated** — appended, dated 2026-09-19, carrying the detail the harness now writes: `does not report stage; model does not declare that it consumes static width_m, slope, manning_n`. Verdict and reason unchanged (N/A, INCOMPLETE), and no container starts for any of them.

## The ceiling

It is in, as `max_depth_m`: off unless a probe sets it, because how deep is too deep is a statement about the reach, and this probe sets 10 m. It is a probe parameter rather than `bankfull_depth_m` read from static, for the same reason I left `bed_elevation_m` out last round — no adapter consumes bankfull, so requiring it would make five manifests declare a key they never read. The value is four times the case's declared bankfull depth and two and a half times the deepest reading a must-pass baseline gives, and that is measured rather than argued: all four score all 1460 steps with the ceiling refusing none (`too_deep_steps` 0), and the deepest is 3.99 m — `reference_bucket`'s worst Fr of 0.445 inverted through the section's normal-depth relation. Over a hundred seeds the worst honest Fr reaches about 0.46, still only about 4.9 m, so the headroom holds beyond the gate seeds.

A stage a hundred metres above the bed is now refused with the ceiling named rather than scored as a comfortable pass. A stage *one* metre above it still passes, and no value test can do better than that — which is the point of fixing the datum in the contract instead of inferring it, and it is written into the README's limitations as such.

On `vel` (#95): agreed, and there is now a short "what would make it sharper" section pointing the next author at it, since forming the numerator from the model's own velocity would drop the rectangular assumption from one of the two terms.

## Smaller items

- A test pins `requires_static` and `reference_flat_stage`'s N/A, as you suggested.
- `site/index.html`: `momentum/froude-regime` moves to the merged block in all three languages — and so do `momentum/stage-discharge-monotonic` and `mass/multi-decadal-drift`, which sat in the wanted region as merged rows for the same reason. Those three were the whole contamination, so the blocks are contiguous again. The last two are not this PR's; they are two lines each and reverting them is fine by me.
- `models/sacsma_snow17/README.md`: six energy probes now, plus the two other probes it is N/A on (`mass/gw-sw-exchange-consistency` and `momentum/routing-lag-consistency`), which is the eight the archive records.
- The PR description's test count is corrected to 24, and the 1109 figure to the current run.

## Verified at `4cadf9c`

- `ht validate`: 30 probes, 65 models. `ht gate --probe momentum/froude-regime`: 4 must_pass + 1 must_fail, separated on `froude_subcritical`.
- Full suite: **1116 passed, 8 skipped**; `tests/test_docs_in_sync.py` and `tests/test_report_detail.py` pass with no count correction.
- The three physical models' archive rows still reproduce: the ceiling refuses none of their steps, so their `detail` sentences are unchanged.

This is the last head unless the review asks for more.

Barbhuiya12 (dismissed):
Re-reviewed at `4cadf9c`. The datum is settled the right way round, and I would
rather have this answer than the one I asked for.

`ht validate` (30 probes, 65 models), full `ht gate`, `pytest -q` (1124 passed)
and the probe's own gate all pass here.

## The fix

I asked for a ceiling. You pinned the contract first and made the ceiling a
backstop, which is the correct order — the ceiling can only ever catch a gross
offset, and `AGENTS.md` now says what the number means:

> measured from the reach bed: the datum is the bed, so an absolute level, or a
> level above any other zero, is a different quantity and not this variable

Deleting `DEFAULT_BED_KEY` and `FALLBACK_BED_M` outright is the part I would have
got wrong. Subtracting a bed with a zero default reads as a safe accommodation and
is actually the mechanism: it makes an absolute level *look* convertible. Refusing
to convert is the honest behaviour.

Re-running my own test against the new head:

```
reference_shallow_rating, as run   FAIL   98.5% supercritical   worst Fr 1.852
+  0.5 m datum                     FAIL    0.1%                 worst Fr 1.147
+  1.0 m datum                     PASS                         worst Fr 0.798
+  5.0 m datum                     PASS                         worst Fr 0.178
+ 25.0 m datum                     FAIL   refused: 0 of 1460 steps deep enough to score
```

So the ceiling closes the elevation case and leaves the metre-scale one open —
which is exactly what the README now says, in the sentence I would have written
if I had thought about it harder:

> What no value test can catch is a *small* offset — a stage a metre above the
> bed is indistinguishable from a deep reach — which is why the datum has to be
> fixed by the contract rather than inferred from the numbers.

That is the right claim and the right scope for it. Nothing further from me here.

## The ceiling does not touch honest flow

A ceiling can only cost false N/A, so I swept seeds 0-99 plus the gate seeds for
all five baselines — 515 runs — reading `too_deep_steps`:

```
model                      role        n    wrong   worst Fr        too_deep_steps
reference_bucket           must_pass  103       0   0.412 - 0.466   0 on every seed
flex_lumped                must_pass  103       0   0.369 - 0.397   0 on every seed
flex_topo                  must_pass  103       0   0.399 - 0.428   0 on every seed
sacsma_snow17              must_pass  103       0   0.388 - 0.424   0 on every seed
reference_shallow_rating   must_fail  103       0   1.753 - 1.984   0 on every seed
```

Nothing lands on the wrong side, and the 10 m ceiling refuses not one step of
honest flow across the draw. Against the 1.05 limit the passing models keep a
factor of 2.3 and the cheat clears it by 1.7, so the separation is the reach's
rather than the seed's — which is what a Manning gauge in a declared section
should give.

@chrimerss's three items also check out: `requires.static: [width_m, slope,
manning_n]` with all five manifests declaring them in `uses_static`, and the
seven submitted-model rows regenerated at 2026-09-19 with the fuller detail
string.

## One thing this now decides for #114

The datum pin resolves an open question in someone else's branch, and not in its
favour, so it should not be discovered at merge time.

#114 (`momentum/uniform-flow-friction-consistency`) computes `depth = stage - bed`
with `bed_elevation_m` drawn from `uniform(25.0, 180.0)` m, and changes the shared
adapters — `reference_bucket`, `flex_lumped`, `flex_topo` — to report
`stage = bed_elevation_m + depth`. Under this PR's contract row those adapters
report a quantity that is not `stage`, and a contract-compliant model reporting a
0.5 m depth hands #114 `0.5 - 25 = -24.5 m`, tripping its own "stage is at or
below the bed" guard.

The two cannot both merge as written. I think this PR has it right and #114's
`bed_elevation_m` should go: @licm13 already observed there that it adds no
physics to the Manning identity, and with `stage` pinned as a depth the key drops
out of the residual entirely — which also dissolves his blocking concern about
`CV(stage)` on `stage-discharge-monotonic`, since the adapters would stop adding a
datum at all. I have said the same on #114 and am flagging it here so @chrimerss
sees both sides.

## Approving

The finding I raised is answered more thoroughly than I raised it, the fix is
verified against the case that motivated it, the ceiling costs nothing over a
wide draw, and the residual limit is stated where a reader will meet it.

kawh1111:
Thanks for the re-review — and for the 515-run sweep, which goes past what the note asked for. The seed range is the part of this I could not check myself, and a 2.3x / 1.7x separation held across it is a stronger statement than the four gate seeds made.

Noted too that you have put the #114 conflict on the record with numbers. That row is now what both branches turn on, so: I am not touching either branch, the head here stays at `4cadf9c`, and nothing further gets pushed unless a review asks for it.

The residual window you re-measured is the one the README now states — a metre-scale offset is indistinguishable from a deep reach, which is why the datum is fixed by the contract rather than inferred from the numbers. No change needed there.

chrimerss (changes requested):
At `4cadf9c`. This is not a review of the Froude physics, which I have no further
objection to — it is the datum, which @Barbhuiya12 has now raised on both branches
and which I have settled on #114. This branch is on the losing side of it, so it
needs to flip before it merges.

## The decision

**`stage` is a water-surface elevation on a fixed vertical datum, and the datum is
not the bed.** That is the USGS convention — a gauge reads a depth above a gauge
datum — and the reason for it is exactly why the two are separate quantities: a
bed scours and fills, a datum does not. A `stage` pinned to the bed is a reading
whose zero moves under the model reporting it.

So the row this branch added at `AGENTS.md:169` —

> the water level a gauge in the reach would read, **measured from the reach bed**:
> the datum is the bed, so an absolute level, or a level above any other zero, is a
> different quantity and not this variable

— has to go, and `froude_subcritical` has to go back to forming a depth rather than
reading one.

## What that means here

1. **Restore the subtraction.** `_depth()` takes `stage` as the depth today
   (`deep_enough = stage > min_depth_m`, `np.maximum(stage, min_depth_m)`). It
   needs `depth = stage - bed_elevation_m` again, and `min_depth_m` /
   `max_depth_m` need to bound **that**, not the level. Right now both bounds are
   bounds on an elevation, which is why they read strangely against a real gauge.
2. **Declare the datum in the case.** `generate.py` deliberately supplies no
   `bed_elevation_m` ("No `bed_elevation_m`. `stage` is a depth measured from the
   reach bed by contract"). Supply it, and add it to `requires.static` beside
   `width_m`, `slope` and `manning_n`. The rule I am putting in
   `docs/writing-a-probe.md` is that any criterion forming a depth, or a ratio or
   power of the level, must require the key — `froude_subcritical` takes
   `d^{3/2}`, so it is squarely in scope. A model that does not declare it
   consumes the datum is then `N/A (INCOMPATIBLE)`, which is the right answer for
   a model judged against a section it never read.
3. **`max_depth_m: 10.0` does not substitute for the datum, and the comment above
   it should stop saying it does.** The ceiling is argued as the thing that stops
   "a stage reported from some other zero" passing, because an offset only adds to
   the depth. It catches a gross offset and misses the one that matters:
   @Barbhuiya12 measured `reference_shallow_rating` — this probe's own `must_fail`
   control — flipping to PASS with a **1 m** offset, worst Fr falling 1.852 to
   0.798. Both his figures imply a worst-step depth of 1.33 m, so the offset run
   tops out near 2.3 m, nowhere near 10 m. A metre defeats the control and the
   ceiling never sees it. With the datum declared and subtracted the ceiling is
   still worth keeping, but as what it honestly is — a section sanity bound — not
   as the defence against a datum mismatch.

## Process

@Barbhuiya12's approval at `4cadf9c` was given against the convention that is not
being adopted, and he said as much in the same breath on #114 ("not my call to
make, and I have said the same on #103"). I am dismissing it rather than letting
it carry a merge; please re-request once the flip is in.

Whichever of #103 and #114 merges second will need the rebase and the suite
recount either way.

Barbhuiya12 (commented):
Withdrawing the part of my approval that endorsed the datum row, so it is not
counted on the side that lost.

I approved `4cadf9c` on the strength of the ceiling, the 515-run sweep and the
contract row together. @chrimerss has settled the row the other way on #114, and
the reasoning is better than mine: a bed scours and fills and a datum does not, so
tying `stage` to the bed gives a mobile-bed model a number whose zero moves under
it. My cost comparison priced the criterion patches and not what the contract can
express — under my reading a model could never report a real gauge level as
`stage`, and that is a larger thing to give up than two subtractions. I was
weighing the wrong side of it.

**What still stands from that review**, because none of it depends on which way
the datum went:

- The 515-run sweep: 0 wrong-side verdicts, `must_pass` worst Fr 0.369-0.466 and
  `reference_shallow_rating` 1.753-1.984 against the 1.05 limit, so the 2.3x / 1.7x
  separation is the reach's rather than the seed's.
- `too_deep_steps` was 0 on every seed for every baseline, so whatever the ceiling
  becomes under the new row, at 10 m it costs honest flow nothing here.
- @chrimerss's three earlier items check out: `requires.static`, the five manifests,
  and the seven regenerated archive rows.

**What the flip will need**, since the criterion currently has no bed handling at
all — it was deleted in this revision, at my suggestion:

`froude_subcritical` forms `d^{3/2}`, so by the rule @chrimerss is putting in
`docs/writing-a-probe.md` it is on the ratio-or-power side and needs
`bed_elevation_m` in `requires.static` and a real subtraction back in `froude.py`.
Worth taking the shared `depth_series(run)` helper he is asking #114 for rather
than writing a second copy — and taking the version that **raises** on a missing
key, not one that defaults the bed to zero, since defaulting to zero is exactly
the sentence he struck from #114's row.

The measurement that motivated the ceiling is unaffected by any of this: a one
metre offset still turns `reference_shallow_rating` from 98.5% supercritical into
a comfortable pass, and a value test still cannot see a metre. Under the new row
that stops being a backstop against a contract-compliant reading and becomes a
guard against a model that ignored the declared datum, which is a narrower job but
still worth doing.

Sorry to have pointed you at a row that then moved — the ceiling and the sweep
were worth having either way.

kawh1111:
Re-requesting, as asked. Two things changed, and one of them is structural.

**The flip.** `stage` is now the water-surface elevation on the case's fixed
vertical datum, and the depth is `stage - bed_elevation_m`. `froude.py` forms it
through the shared `depth_series(run)` instead of reading the level as a depth;
the probe requires `bed_elevation_m` beside the section, so a model that never
read the datum is N/A (INCOMPATIBLE) rather than failed; and `generate.py`
declares it as a real elevation (120 m) rather than zero, since a zero datum is
the same conflation wearing a disguise. `max_depth_m` stays, but the comment now
describes it as what it honestly is — a section sanity bound, not the defence
against a datum mismatch, which the one-metre offset @Barbhuiya12 measured shows
it never was. The `AGENTS.md` row this branch had added is gone; the variable
table now carries the settled definition.

**The base.** This is now stacked on #114
(`probe/uniform-flow-friction-consistency`), and that is deliberate rather than
incidental. Requiring `bed_elevation_m` is not confined to this probe: it makes
every `must_pass` baseline `N/A (INCOMPATIBLE)` unless the shared adapters also
report an elevation, and that work lives in #114. Rather than duplicate it here
and hand you a merge conflict in `criteria/base.py` and the four adapters, this
branch sits on it and the pull request shows only this probe's 17 files. Your
point that whichever of the two merges second needs the rebase and the suite
recount still holds — this only makes the dependency visible instead of implied.

**Verified at `e87d3cf`**, restacked on `ff59fca9` after #114 moved — `ht validate`
33 probes / 77 models; full `ht gate` PASSED; `pytest -q` 1196 passed, 8 skipped.
Both GitHub checks are green on this head: `container` and `gate`.
`models/result.csv` regenerated for this probe (10 rows: the three subprocess
baselines PASS, the seven container models N/A INCOMPLETE), and the README, site
and `sacsma_snow17` counts moved by one.

One detail worth naming: `reference_shallow_rating` now reports
`bed_elevation_m + depth`, so its own gauge is an elevation on the declared datum
like every other adapter's. That is the whole of what the datum flip asks of a
model, and the must-fail still separates on `froude_subcritical` alone.

kawh1111:
Restacked on `main` now that #114 has landed, and the base is retargeted from
`probe/uniform-flow-friction-consistency` to `main` — this branch is a single
commit on `ee4dd8c` and the pull request shows only this probe's 17 files.

The datum flip is in `main` with #114, so this branch no longer carries an
`AGENTS.md` row: `stage` is the water-surface elevation on the case's fixed
vertical datum and the depth is `stage - bed_elevation_m`. What this branch does
with it is unchanged from what you asked for — `froude.py` forms the depth
through the shared `depth_series`, `requires.static` names `bed_elevation_m` and
`generate.py` declares a real elevation (120 m) rather than zero. `max_depth_m`
stays as a section sanity bound, which is what the one-metre offset measured on
this probe showed it always was.

The only conflict was `README.md`, where `main` had gained the two snowpack
reference rows; the resolution keeps them.

Verified on `d0a8a41`: `ht validate` 33 probes / 77 models, full probe gate
PASSED, `pytest -q` **1207 passed, 8 skipped**, and the four must-pass baselines
score all 1460 steps with `too_deep_steps` 0 (worst Fr 0.389–0.445 against the
must-fail's 1.894). The PR description carries the same numbers.

Re-requesting your review.

cehw:
Thanks for working through the datum change, Yuanhang. Using the shared depth calculation now ties the Froude check to the section the case actually declares, and the normal and shallow-rating controls still separate as intended. I think this is a useful basis for the probe, with two local corrections needed in how the results are classified.

The first is the datum-mismatch classification. [`depth_series` already identifies the incompatible convention](https://github.com/kawh1111/HydroTuring/blob/d0a8a415afa3299435734352121822c85fa77754/src/hydroturing/criteria/froude.py#L261), but the catch here converts it into a failed criterion. Reporting depth as `stage` therefore gives FAIL/VIOLATION on all three gate seeds through the actual subprocess and CSV path. Letting `CriterionIncompatibleError` reach the harness, before the broader `ValueError` handler, restores N/A/INCOMPATIBLE for this control. It also preserves the new harness's treatment of genuine failures on other scored seeds. Could we retain that classification and pin the final verdict in a regression?

The second is that an invalid flow value can disappear from the assessment. With one peak `dis` value replaced by NaN, the [`q > 0` mask](https://github.com/kawh1111/HydroTuring/blob/d0a8a415afa3299435734352121822c85fa77754/src/hydroturing/criteria/froude.py#L291) drops that day and the probe still passes on each seed, scoring 1459 rather than 1460 steps. The CSV reader does not require `dis` here because the probe also supports `mrro`. Checking the selected discharge series for finite values before the mask gives FAIL/ERROR and keeps both valid input paths available. Both local corrections leave the normal and shallow controls' verdicts unchanged. These are reporting controls, rather than measured failures of a native river model.

The rebase also preserves the snow-model rows and updates the stage definition in the PR description. The remaining documentation suggestions are non-blocking: carry the PR body's acknowledged metre-scale offset limitation into the probe README, use 33 probes (21 mass, 7 energy, 5 momentum), label the variability table `CV(depth)`, and distinguish this flow-regime check from #114's steady friction balance.

Validation at `d0a8a41` on Linux: **1215 tests passed**, `ht validate` reports 33 probes and 77 models, and the full `ht gate` passed.

kawh1111:
Both local corrections are in, and so are the four documentation items.

**The datum-mismatch classification.** You are right, and the catch was mine to
remove. `depth_series` raises `CriterionIncompatibleError` for a column that is a
depth where the contract asks for a level, and the criterion no longer converts
it: the exception reaches the harness, which records the seed as N/A
(INCOMPATIBLE). Pinned twice — a `pytest.raises` on the criterion, and an
end-to-end `run_probe` in which `reference_bucket`'s own run has the declared bed
taken back out of `stage`:

    (outcome.verdict, outcome.reason) == (NOT_SCORED, INCOMPATIBLE)
    len(outcome.incompatible) == 3
    "seed …: reported stage is far below bed_elevation_m on most finite steps;
     stage must be water-surface elevation on the case's fixed vertical datum,
     not depth above the bed"

The `ValueError` handler for a case that declares no usable bed is untouched — it
is a case-side failure rather than a convention mismatch.

**The disappearing flow value.** Fixed before the mask, as you said, and the bad
value is now counted instead of skipped. The same hole exists on `stage`, which
is the other half of the same mask, so both series are refused; the check is read
on the window the criterion scores, so a spinup value cannot fail it. The
precedent I followed is `radiative_identity`, which refuses a non-finite surface
temperature with `diagnostics={"non_finite_steps": n}`, and this reports the same
key. The `mrro` path stays available: the check is on the selected series, not on
a required column.

**The four documentation items.** `CV(depth)` it is, and the numbers hold under
that label: 0.56 – 0.94 over the gate seeds read over the whole record, identical
between the must-fail and the honest gauge because a constant factor cancels from
a coefficient of variation. The metre-scale limitation is in the probe README with
the window measured rather than asserted — +0.5 m turns one gate seed into a pass,
+1 m all three, +5 m leaves all three comfortable, and +10 m is where the ceiling
takes over. The counts are corrected to thirty-three probes, twenty-one mass,
seven energy, five momentum, in both places `README.md` states them: that is a
sentence no test reads, which is how it drifted. The distinction from #114 is in
"What no budget can see" — the friction balance asks whether depth, slope and
roughness agree with each other at steady flow, this asks which regime the
resulting depth puts the flow in.

**Three numbers I corrected while checking the rest, none of which you flagged.**
The separation table's honest worst Fr was 0.389 – 0.445 and its median 0.256 –
0.281; on the gate seeds they measure 0.376 – 0.453 and 0.245 – 0.284, and the
must-fail's worst Fr is a range, 1.772 – 1.928, rather than one seed's 1.894. The
claim that all four must-pass baselines score all 1460 steps is not quite true:
`flex_topo` scores 1449 and `sacsma_snow17` 1423 on one seed each, where a handful
of steps fall under the depth floor — `too_deep_steps` is 0 on every seed of every
baseline. And the ceiling's margin was quoted from the wrong seed: the deepest
honest reading is 4.44 m on the gate seeds, not 3.99 m, and 5.27 m over a hundred
seeds, which still leaves 1.9x of room.

Validation at this head: `ht validate` 33 probes / 77 models, `pytest tests`
**1209 passed, 8 skipped**, `ht gate --probe momentum/froude-regime` passes.

This is the head I am asking you to read. Unless a review asks for something else,
it is the last one.

Barbhuiya12 (commented):
Re-reviewed at `1729da5`. `ht validate` (33 probes, 77 models) and the probe's
gate pass here. The datum flip is done the way @chrimerss ruled it on #114:
`froude.py` forms the depth through the shared `depth_series`, `requires.static`
names `bed_elevation_m`, and the generator declares a real 120 m elevation rather
than zero — which is the right call, since a zero datum is exactly where the two
conventions coincide and the mismatch hides. Thank you for describing
`max_depth_m` as what it is, a section sanity bound; the one-metre offset showed
it was never the defence against a datum mismatch, and the README now says so.

@cehw's two corrections both hold on the current head:

```
depth reported in the stage column    -> N/A (INCOMPATIBLE)            not a violation
one NaN at the peak discharge         -> FAIL, non_finite_steps = 1   was a pass on 1459 of 1460 steps
```

One thing left, and it is the second of those two points again, by a door the
NaN fix did not close.

## A step is excused as dry on its depth alone, even while water is flowing

```python
scored = deep_enough & (q > 0.0)
```

A step with discharge but a depth at or under `min_depth_m` is dropped as dry.
That is backwards for this criterion. Water moving at a sub-centimetre depth is
not a dry reach — at a given discharge it is the *fastest* flow the record can
contain, because $\mathrm{Fr} = Q / (w\sqrt{g}\, d^{3/2})$ grows without bound as
$d \to 0$. The mask drops exactly the steps that most violate what the probe
tests.

Taking `reference_bucket` on seed 3 and giving its flood peaks — the 30% of
steps with the largest discharge — a shallower depth, discharge unchanged:

```
honest                          PASS   scored 1460/1460   worst Fr 0.44   supercritical   0
peaks at 20% of their depth     FAIL   scored 1460/1460   worst Fr 4.86   supercritical 438
peaks at 5 mm                   PASS   scored 1022/1460   worst Fr 0.30   supercritical   0
```

The verdict is monotone in the wrong direction. A model that is too shallow at
the peaks fails; the same model made *more* wrong passes, because the 438 steps
it would fail on fall under the floor and leave the scored set. At 5 mm those
steps would carry Froude numbers in the thousands. The remaining 1022 scored
steps are 70% of the record, far above the 5% `min_scored_fraction`, so nothing
reports that a third of the flood went unexamined.

The comment directly above that line already names this exact escape —
"an invalid input leaving the assessment by the door an honest dry step uses" —
and closes it for a non-finite reading. A finite depth under the floor with real
discharge behind it goes through the same door.

## The fix, and it is free

Decide dryness on the flow, not the depth. `_depth` already clamps the depth to
the floor for the division, so every flowing step can be scored at the clamped
depth:

```python
scored = q > 0.0
```

`min_depth_m` then does only the job its own comment gives it — keeping a
receding flow from dividing by zero — and a step is excused only when nothing is
moving. Same run:

```
                                              now     with the fix
honest                                        PASS    PASS
peaks at 20% of their depth                   FAIL    FAIL
peaks at 5 mm, discharge unchanged            PASS    FAIL   (438 supercritical)
peaks at 5 mm and no discharge (truly dry)    PASS    PASS   (still excused)
```

A genuinely dry step is still excused. Only a step that is carrying water gets
scored.

To check it costs the honest models nothing, I ran all five baselines over seeds
0-99 on the current head and on the patched one:

```
                            current                                with the fix
model                       wrong  worst Fr       scored           wrong  worst Fr       scored
reference_bucket               0   0.412-0.466   1460-1460            0   0.412-0.466   1460-1460
flex_lumped                    0   0.369-0.397   1460-1460            0   0.369-0.397   1460-1460
flex_topo                      0   0.399-0.428   1420-1460            0   0.399-0.428   1460-1460
sacsma_snow17                  0   0.388-0.424   1351-1460            0   0.388-0.424   1400-1460
reference_shallow_rating       0   1.753-1.984   1460-1460            0   1.753-1.984   1460-1460
```

No verdict moves and no worst-Froude figure moves in the third decimal. The
informative rows are `flex_topo` and `sacsma_snow17`: the fix scores up to 40 and
49 more steps on them, so those *were* flowing steps under the floor that the
current mask had been excusing — and every one of them is subcritical. An honest
model's low-depth flowing steps are slow, as they should be; only a model that
misreports its depth produces a fast one. The steps `sacsma_snow17` still leaves
unscored carry no discharge, and stay excused.

## Where that leaves me

Not approving yet, only because of the mask. Everything else is right, and it
is a one-line change that the sweep above says moves nothing an honest model
reports. Once it is in with a regression on the 5 mm case — the one that passes
now — I will approve.

This is the same shape as #140, which is open against `4ac4c7e` for
`uniform_flow_friction`: a criterion that lets the model decline the part of the
record it would fail on, one level below where the seed floor can see it. Worth
fixing in both places the same way.

kawh1111:
Fix is in on `e8da3ef9`, and it is the mask.

**Dryness is read from the flow now.** `scored = within_section & (|q| > 0.0)`, with the
floor doing only what its own comment says — giving the division a number. The depth is
clamped to it, so a flowing step is scored at the clamped depth however shallow the model
reports it, and the only step still excused is one with nothing moving through it. Your
counterexample, on `reference_bucket` gate seed 3 with the discharge untouched throughout:

```
honest                          PASS   scored 1460/1460   worst Fr    0.435
peaks at 20% of their depth     FAIL   scored 1460/1460   worst Fr    4.864
peaks at 5 mm                   FAIL   scored 1460/1460   worst Fr 2834.7
```

The last line is the one that passed before. The steps it would have excused are exactly
the steps it goes supercritical on, and the arithmetic is the point: the floor is what the
division uses, so a peak reported at five millimetres reads as a flood discharge with a
centimetre of water under it, which is what it is.

**Two places where I kept more than the one-liner**, because the same door reopens otherwise:

- **The ceiling stays in the mask.** `scored = q > 0.0` would have taken `max_depth_m` out
  with the floor, and that bound is the only thing between a model that ignored the declared
  datum and a comfortable pass: an offset only adds to the depth, and a deeper depth is a
  smaller Froude number. Measured on the must-fail model with an offset added to its reported
  stage — **+0.5 m** turns one gate seed into a pass, **+1 m** all three (worst Fr 0.89 /
  0.96 / 0.64), **+5 m** comfortable, and only at **+10 m** does the ceiling refuse every
  step (scored 0/1460, 1460 above the ceiling). Clamping a too-deep step the way the floor
  clamps a shallow one would put that reading back inside the range the bound exists to
  exclude.
- **The magnitude, in the mask and in the quotient.** `Fr = |Q| / (w d**1.5 sqrt(g))`.
  A negative reading is water moving, so a sign test in the mask takes a step out of the
  sample for being "not positive" — your finding one step further out — and a signed
  numerator comes out negative, which is below the limit at every step: a reading that
  cannot fail rather than one that passes.

**Cost, re-measured rather than taken.** Seeds 0–99, five baselines, no verdict moves and no
worst Fr moves in the third decimal:

```
model                       worst Fr        scored
reference_bucket            0.412 - 0.466   1460 - 1460
flex_lumped                 0.369 - 0.397   1460 - 1460
flex_topo                   0.399 - 0.428   1460 - 1460
sacsma_snow17               0.388 - 0.424   1400 - 1460
reference_shallow_rating    1.753 - 1.984   1460 - 1460
```

which is your patched column to the third decimal. The informative rows are `flex_topo` and
`sacsma_snow17`: they were leaving flowing steps under the floor, and every one of those
steps is subcritical, as an honest model's low-depth flowing steps should be.

**Regressions.** `test_a_shallow_step_carrying_water_is_scored_not_excused` is the 5 mm case
above; `test_a_step_with_no_flow_is_still_excused` takes the same record and zeroes the
discharge on the dry half, which is still excused; `test_the_clamped_floor_is_what_a_shallow_flow_is_scored_at`
pins that the floor is the number the division uses; `test_the_ceiling_still_refuses_where_the_floor_no_longer_does`
pins the asymmetry; `test_a_negative_reading_is_water_moving_rather_than_an_excuse` pins the
magnitude. The docstring bullet that said dry steps are excluded is rewritten. 31
counterexamples now.

`ht validate` 33/77, `ht gate --probe momentum/froude-regime` passed, full suite **1213
passed, 8 skipped**.

**One door left open, measured rather than left to be found.** A step is excused when the
model reports *exactly* no flow through it, and the model is the one reporting. On the same
seed: honest flow on half the record and a report of zero on the other half gives scored
769/1460 and a PASS; a report of zero on 95% of it gives scored 92/1460 and still a PASS.
So `min_scored_fraction: 0.05` is the only bound on how much of a record a model may declare
dry, and it sits a fifth of the way down. The honest draw now leaves 1455 – 1460 scored on
the gate seeds and 1400 – 1460 over seeds 0–99, the shortfall being `sacsma_snow17`'s steps
with no flow at all, so the parameter has room — but it is the case's declaration rather than
the criterion's, and what a criterion may skip is the rule #118 is about to settle. I have
not bundled it into this commit: 1.0 is out of reach for that reason, and ~0.9 would close
the door with the margin above. Say which and it goes in, or say leave it and it is recorded
as it is in the probe's README.

Also `models/result.csv`: the ten rows dated 09-21 recorded this criterion's message as it
was before the change, for the same model versions. I replaced them rather than adding a
second set — two contradictory details for one version is worse than one — so the file now
carries a single row per model, dated 09-23. If you would rather the archive only ever grows,
say so and I will put them back.

Barbhuiya12 (dismissed):
Re-reviewed at `e8da3ef`. The 5 mm case that passed before now fails — scored
1460/1460, worst Fr 2834.7 — and `ht validate` (33 probes, 77 models), the probe's
gate and the five new regressions all pass here. That was my condition, so I am
approving. Two of your changes are better than what I asked for, and you asked me
two questions.

## Where my one-liner was wrong

I suggested `scored = q > 0.0`, and on both counts you kept more than it would have:

- **The ceiling.** Dropping `within_section` would have taken `max_depth_m` out of
  the mask along with the floor, and that is the one bound between a model that
  ignored the declared datum and a comfortable pass. I measured that offset on this
  very PR last week; it would have been a strange thing to reopen.
- **The magnitude.** `q > 0` excuses a reverse-flow step as if nothing were moving,
  and a signed numerator makes Fr negative, which is below the limit on every step:
  a reading that cannot fail. `|q|` in both the mask and the quotient is right.

Thank you for checking both rather than taking the suggestion as written.

## `min_scored_fraction`: take 0.9, but it narrows the door rather than closing it

I ran your open door directly: `reference_bucket`, seed 3, with the largest-
discharge steps made supercritical and then reported as carrying no flow at all, at
three floors:

```
share of steps zeroed    0.05 (shipped)        0.90                  0.95
 3%                      PASS  1417/1460       PASS  1417/1460       PASS  1417/1460
 8%                      PASS  1344/1460       PASS  1344/1460       FAIL  1344/1460
30%                      PASS  1022/1460       FAIL  1022/1460       FAIL  1022/1460
95%                      PASS    73/1460       FAIL    73/1460       FAIL    73/1460
```

So 0.9 is worth taking. It costs the honest baselines nothing — their minimum is
1400/1460, 95.9%, in your sweep and in mine — and it closes the gross cases. But the
3% row passes at every floor, and that is structural rather than a matter of tuning.
The escape is always `1 − min_scored_fraction` of the record; the honest models
already leave up to ~4% unscored, so no floor they can meet reaches below that; and
the top few percent of discharge is exactly where a supercritical reach would be
supercritical. A floor bounds the size of the door. It cannot reach the flood peaks.

Your own rule in the new docstring points at the full fix. A step with no flow
cannot be supercritical — but only if "no flow" is true, and here the model is the
one saying so. Closing it needs dryness read from the case rather than from the
model's own discharge, and the harness already has the precedent: `event_signal`
takes the probe's physical baseline's runoff as the case's definition of what the
catchment does with its weather. A step the case says is wet and the model reports
as dry would then be a step the model declined, and could cost the pass. That is
more machinery than this PR needs, and you are right that it is the question #118 is
about to settle, so my suggestion is: **0.9 here, and the residual recorded in the
README in those terms** — a model supercritical only at its largest few percent of
discharge can report no flow there and pass; the floor bounds how much, not whether.

## The archive: replacing is right here

`main` carries no `momentum/froude-regime` rows at all, so none of the ten you
replaced was ever published. The append-only convention exists so the published
record never loses history; rows that only ever lived on this branch have no history
to lose, and two contradictory details for one model version would be worse than
one. Keep them replaced. From merge on, append.

## Approving

The mask fix is in, measured and pinned; the two refinements go past what I asked
for; and the cost to the honest baselines is nil to the third decimal of the worst
Froude number. The zero-flow residual is disclosed rather than hidden, which is the
right state for it to merge in.

kawh1111:
Thank you for the re-read, and for two answers that go further than the questions.

## Both answers taken

- **The archive: keep them replaced, append from merge on.** That is the reading we were working from, and having it stated settles the convention for this probe.
- **The floor: 0.9 accepted.** Your table is the independent measurement of the door our own reply only described, and I agree with the reading of it — a floor bounds how much of the record can leave, and the top few percent of discharge is exactly where a supercritical reach would be supercritical.

## Where 0.9 goes

**Not into this head, and the reason is the vote rather than the value.** Your approval is currently the only one on this PR, a new commit dismisses it under `dismiss_stale_reviews`, and the critical path here is @chrimerss's `CHANGES_REQUESTED` on `4cadf9cf` rather than the floor. Trading a live approval for one parameter bump that your own analysis says narrows a door rather than closes it is the wrong way round.

So both of your items land together with the next commit that has to happen anyway — a rebase, or whatever the contract question in #118 settles into — and the README sentence recording the residual in your terms goes with them rather than alone, since on its own it would cost the same vote for a wording change.

Two of the numbers behind that, measured on the current head rather than carried over from before the mask fix: on all five gate seeds every one of the four `must_pass` baselines and `reference_shallow_rating` scores **1460/1460**, and over seeds 0–99 the smallest scored fraction is `sacsma_snow17` at **1400/1460 = 95.9%** — the figure your own sweep reports. So a step floor of 0.9 keeps 5.9 points of room on the worst honest seed, and 10 on the seeds the gate actually runs. Nothing else about this head changes.

## What your two points did settle

The ceiling and the `|q|` reading — thank you for saying plainly that the one-liner would have dropped the first and excused reverse flow in the second. The criterion now reads dryness from the flow, keeps the section bound as the guard against a model that ignored the declared datum, and takes the magnitude of the discharge in both the mask and the quotient, which is what the regressions added with the fix pin.

chrimerss:
Pushed two commits on top of `e8da3ef` before merging: a merge of `main` (41029d1, #139's docs, no conflicts) and `a3deccb`, which closes the one door still open.

**The zero-flow door let this probe's own must-fail pass.** The scored set was decided on `dis` alone, so zeroing `dis` on exactly the steps `reference_shallow_rating` goes supercritical on, with `mrro` still carrying the water, left 144, 154 and 244 of 1460 steps scored on the three gate seeds. That is above the 5% floor, and all three passed. A gauge that was honest except at its top 5% of flows, zeroed there, passed as well, and the agreed 0.9 floor would not have caught that one.

- `froude.py`: where `dis` reads exactly zero and the model also reports `mrro`, the step is scored on the runoff. Under the contract the two are readings of one outflow. The non-finite refusal now runs on the flow as divided by, so a NaN behind a silent `dis` is refused rather than read as stillness. The message names how many steps came from the second reading. Both constructions now fail on all 1460 steps, and the honest baselines are unchanged.
- `probe.yaml`: `min_scored_fraction` 0.05 → 0.9, as @kawh1111 and @Barbhuiya12 agreed. It now bounds only a model that zeroes every flow column, which also breaks every mass budget.
- `reference_shallow_rating`: it now keeps the bucket's rectangular-section guard and declares `cross_section_shape`, so it is scored wherever the bucket is. Before, it was N/A on #114's probe only because that key was undeclared. It fails `uniform_flow_friction` there, since a borrowed width is wrong at steady flow too. The "no other criterion reads the pair" sentences are reworded to match, and the probe README records what this probe still adds over #114: a gauge that is honest at steady flow and 40% shallow wherever the discharge jumps by more than a quarter passes `uniform_flow_friction` on its first gate seed and fails `froude_subcritical` here (21 steps, worst Fr 1.76).
- The site row moves into the merged block in all three languages. There are four new tests; the three for the rule fail without it.

pytest 1225 passed; `ht validate` 33 probes / 77 models; full `ht gate` PASSED. The ten archive rows reproduce unchanged, so `models/result.csv` is untouched.

The push dismisses @Barbhuiya12's approval under `dismiss_stale_reviews`. I am merging on the maintainer's decision, with the waiver recorded in the merge commit.

PR #118, opened 2026-09-19
[PROBE: mass] Add routing-network closure

### LINKED ISSUE #108: [PROBE] mass/routing-network-closure: Reach-by-reach mass closure in a branching network
### Authors

Jiao Wang, Department of Land, Air, and Water Resources, University of California Davis. Davis, California, USA, https://orcid.org/0000-0002-4029-5814

### Conservation law

mass

### Residual equation

At the headwater sub-catchment scale, the discharge Q is generated and enters the river network. 
Consider a Y-shaped network where upstream reaches A and B join downstream reach C through a zero-storage junction J. 
For each reach, 
Q_in  >> channel storage/routing >> Q_out, where the routing relation is model-specific (e.g., a linear reservoir or a hydraulic routing scheme).
For each reach i and time interval t: 
r_i,t = (Q_in,i,t - Q_out,i,t) Δt - (V_i,t+1 - V_i,t) [m³].  
For junction J: r_J,t = (Q_out,A,t + Q_out,B,t - Q_in,C,t) Δt [m³]. 
where Q_in and Q_out are the interval-mean inflow and outflow for each reach [m³/s], V is reach channel storage [m³], and Δt is the interval length in seconds. 
Because junction J has no storage, its residual contains only the mismatch between combined upstream outflow and downstream inflow. The case contains no lateral inflows, withdrawals, or other unreported exchanges. Reach and junction balances are evaluated separately so that local errors cannot be hidden by whole-network cancellation.

For each reach i, require: 
Σ_t |r_i,t| <= 0.05 Σ_t(Q_in,i,t Δt).
For the junction, require:
Σ_t |r_J,t| <= 0.05 Σ_t[(Q_out,A,t + Q_out,B,t) Δt]. 
The tolerance is 5% of the total incoming water volume to each reach or junction over the scored window. 
The absolute residual from each interval is summed and then divided by the total incoming water volume, so equal-and-opposite errors at different times cannot be cancelled. 
The synthetic cases are designed to provide positive, non-negligible inflow volumes. A near-zero denominator should be treated as a degenerate case rather than a pass.

1. A model could lose water in one reach and create the same amount in another. These errors may cancel out in the whole-network budget, so basin-wide closure can pass even when one or more individual reach budgets fail.
2. A second failure mode is inconsistent transfer across a junction. For example, reaches A and B may each close against their own reported fluxes while  Q_out,A + Q_out,B != Q_in,C. 
The individual budgets can therefore appear internally consistent while water is lost or created during transfer between reaches. The junction residual catches this connectivity error.
3. A third failure mode is temporal cancellation within one reach: water lost during one interval and recreated later could produce zero net full-window residual. Summing absolute interval residuals prevents this error from being hidden. 

1. Generate a daily Y-shaped three-reach network at runtime from a seed. Two headwater sub-catchments drain to reaches A and B, which converge at a zero-storage junction J and feed downstream reach C.
2. Independent synthetic meteorological forcing is applied to the two headwater sub-catchments so that the model generates distinct runoff hydrographs entering reaches A and B. Event timing, duration, and magnitude vary between the two headwaters and across random seeds.
3. Use 365 spin-up days followed by a 90-day scored window. All reaches start empty before spin-up. During spin-up, intermittent forcing events allow the model to develop catchment and channel states, which are carried continuously into the scored period. During scoring, each headwater sub-catchment receives one independent precipitation event within the first 20 days of the scored period, followed by at least 70 dry days. This allows the resulting flows and stored water to propagate through and recede within the network.
4. A proposed first implementation would randomize precipitation timing, duration, and magnitude within prescribed ranges. Keep temperature above freezing so precipitation falls as rain, and keep potential evapotranspiration within a moderate range. Groundwater and other catchment stores may contribute to the modeled headwater runoff, but no additional groundwater-channel exchange or other unreported source or sink is allowed within the routing reaches.
5. Use multiple seeds (proposed n=5) while keeping the Y-network topology fixed.
6. The implementation requires independently reported per-reach Q_in, Q_out, and V.

### Will you build it?

Yes — assign it to me

joannawang-100:
Implements the proposed mass/routing-network-closure probe from #108.

This PR currently adds the probe definition, seeded Y-network generator, and documentation. The probe requires a minimal reach-indexed routing extension so that mass can be checked separately for each reach and at the junction.

The current design uses:
1. two independently prescribed headwater inflows;
2. reach-level routing inflow, outflow, and channel storage;
3. separate reach and junction closure checks;
4. a 90-day scored window with a long dry tail.

The corresponding routing contract, criteria, reference models, and independent physical routing adapter will be added in this PR as implementation proceeds.

Closes #108.


chrimerss (changes requested):
Reviewed at `0363e00`. Thanks for taking this on. The Y network, the per-reach and junction residuals and the long dry tail are a sound design, and the README is clear about what is still missing. I'm requesting changes so this doesn't merge before those pieces exist, not because of the approach.

I ran the generator locally over seeds 0–299: it is deterministic, and the headwater peaks (5.23–33.45 m³/s) and event depths (30–65 mm) match the README.

---

## 1. The branch breaks the whole suite

`ht validate` fails with `at requires: Additional properties are not allowed ('routing' was unexpected)`, and `pytest` stops at collection, because `tests/test_docs_in_sync.py` loads every `probe.yaml`. Merged as it stands, every probe PR after it goes red. Past the schema, `load_probe` would still reject the five unregistered `routing_*` criteria, and `protocol.py` rejects any output file other than `result.csv`, `result.nc` and `run.json`, so `routing.csv` would be refused too.

**Suggested path.** Split the work:

1. A contract PR first: `requires.routing` / `emits.routing` in both schemas, `spec.py`, the `protocol.py` output allow-list and a reader for `routing.csv`, the AGENTS.md variable table, and `verify-adapter`. It changes what every adapter can report, so it deserves its own review.
2. Then this PR, rebased on it, with the five criteria, the `reference_network_*` models, the `wflow_sbm` extension, archived rows and the AGENTS.md merge checklist.

Please mark this PR as a draft until then. Note that `Closes #108` will close the proposal on merge.

## 2. `wflow_sbm` should not be in `must_pass`

Every `must_pass` entry on main is a trusted subprocess model. `wflow_sbm` runs in Docker over the full 455-row record, and `ht gate` runs every probe, so every probe PR's 15-minute gate job would have to build and run Wflow. Use the exception in `docs/writing-a-probe.md` that `mass/gw-sw-exchange-consistency` uses: gate on `reference_network_exact` alone, archive the extended `wflow_sbm` rows in `models/result.csv` as the independent physical check, and say why in the README. Wflow is the right physical model here, since it has a real river network, but none of the four gate baselines has one, so this is the case the exception exists for.

## 3. A 1% leaky router cannot calibrate the absolute allowance

The README says `reference_network_leaky_router` "constrains how far the absolute allowance may be relaxed". At 1% it barely does. I modelled it as a 1-day linear store that keeps 1% of each day's inflow in a dead store, run over the whole record, and checked reach A on 200 seeds:

| retained share | peak storage / allowance | `min_allowance_mm` at which it escapes |
| --- | --- | --- |
| 1% | 50–83× | 2.48 mm |
| 0.1% | 5.0–8.3× | 0.25 mm |

At 1% the floor could be raised fifty-fold before the gate noticed. #83 settled on a tenth of a percent for the same reason. Please use 0.1%, and write the escape point next to `min_allowance_mm` once it is calibrated against Wflow, as `momentum/routing-conservation` does.

## 4. The recession bound is stricter than the README says

With zero inflow through the 71–79-day tail, the proportional allowance is gone 30 days after the last peak, and only the 0.05 mm floor is left. A linear store with a 4-day time constant peaks at 0.50 of the allowance on reach A, over 300 seeds; with 5 days it reaches 2.4×, and with 8 days 25×. The crossing is at 4.4 days (1.02× on the worst seed; 4.3 days gives 0.87×). So the real limit is an e-folding time just under 4.4 days, not the 15 days that `max_lag_days` suggests. That is probably reasonable for 60–150 km² reaches, but the README should say so, since it also says a mass probe should leave timing to other probes.

---

## Two points in the automated review that don't hold

So you don't spend time on them:

- **Channel capacity.** No router that passes the recession bound reaches the declared capacity. Over 300 seeds, a pure lag of 1, 5, 10 or 15 days peaks at 0.49, 0.97, 0.97 and 0.97 of it at A and at 0.29–0.96 at C. Linear stores that pass the recession bound (k ≤ 4 days) peak at 0.68 or less. The last spin-up inflow ends at least 27 days before scoring, so nothing is carried over. What does need fixing: the comment at `generate.py:32` calls the bounds "deliberately loose", but they are 2.6% above the largest scored event, and no `must_fail` model exercises the upper bound. `reference_network_negative` is pinned to negative discharge. Either add a control that trips the capacity limit or drop the limit. Reach closure already catches invented water.
- **Junction.** The fault is not left open if an adapter constructs `q_in,C = q_out,A + q_out,B`. The loss then shows up in reach C's own closure instead, with the same denominator and the same 5% threshold. On seed 0, a 3, 6 and 10% loss at J gives C a residual of 0.030, 0.060 and 0.100 of its inflow. What is lost is only which criterion names it. The README at lines 77–81 should say that, rather than implying the cheat escapes.

## Smaller, agreeing with the automated review

- `storage_upper_static: channel_capacity_m3` and the per-reach areas sit inside `routing_network`, not at the top level where `state_bounds` looks. The recession floor must use `contributing_area_km2[i]`, not the top-level `area_km2` (150), or reach A's floor is 2.5 times too loose.
- `generate.py:139`: `dict(STATIC)` shares the nested `routing_network` between cases. Use `copy.deepcopy(STATIC)`.
- `generate.py:116`: the dry-tail check is a bare `assert`, which disappears under `python -O`. Raise instead.
- README line 113: the scored events start within the first 14 days but can run to day 19; the inflow-free tail is 71–79 days, not 76 or more.

joannawang-100:
@chrimerss Thanks for the detailed review. I have marked this PR as a draft. I will first prepare a separate contract PR covering the routing schemas, `spec.py`, `routing.csv` protocol support, the `AGENTS.md` variable table, and `verify-adapter`. Once that work is reviewed and merged, I will rebase this probe PR and complete the criteria, reference models, Wflow extension, archived results, and merge-checklist updates.

Barbhuiya12 (commented):
Read at `0363e00` while it waits on the routing contract. @chrimerss has covered
the design thoroughly and I agree with all four points, so there is nothing to
re-run here yet: it is a draft, `ht validate` fails on `requires.routing` as he
says, and there is no criterion to gate. One thing for when the criteria are
written, because it cost two merged probes a round this week and is easier to
design out now than to find afterwards.

**Whatever a criterion skips, decide it from the case, not from the model's own
output.** Twice in the last few days a criterion excused part of a record on a
reading the model controls, and the model could use that to decline the part it
would fail on:

- `uniform_flow_friction` skips an unsteady plateau and scores the seed on the
  rest. A friction error four times the tolerance becomes a pass with a 3% ripple
  on the plateau where it sits (#140).
- `froude_subcritical` in #103 treats a step as dry on depth alone. A model
  carrying the flood-peak discharge at a sub-centimetre depth — the fastest flow
  in the record — has those steps dropped, and a failure on 438 steps becomes a
  pass.

This probe has the same ingredients: a long dry tail, a recession bound that falls
to the 0.05 mm floor thirty days after the last peak, and closure scored per reach
and at a junction. If any of those criteria excuse a reach, a junction, or a stretch
of the tail on something the model reports — a zero inflow, a store below some
floor, a window the model declares quiet — that is the door. The safe forms are to
key the exclusion to the prescribed headwater forcing, which the model cannot move,
or to take @chrimerss's rule from `4ac4c7e`: a failure on a scored unit stands,
and declining a unit can buy an N/A but never a pass.

On calibrating `min_allowance_mm` against Wflow once it exists: I bisected the
allowance on #83 to find where `reference_leaky_router` stops being caught, per
gate seed, through the harness rather than by reimplementing the bound. Happy to
run the same here against the 0.1% router when the criteria and models are in, so
the escape point written next to the parameter is measured on this network rather
than carried over from that one.

PR #134, opened 2026-09-21
snowpack-mass-closure: raise the peak floor, and let a probe demand an isolated must_fail control

chrimerss:
Two follow-ups from #102, in separate commits — the second can be dropped without touching the first.

## 1. `min_peak_fraction` 0.5 → 0.80, `BYPASS_FRACTION` 0.30 → 0.15

At 0.5 the floor was far below anything a snow model produces, and a model that reports 49% of every snowfall as `snm` — liquid water delivered to the ground at −6 °C — and stores the rest closes its snowpack budget to machine precision, scores a peak fraction of 0.51, and passes all three criteria on 30 of 30 seeds.

Raising it alone does not work, which @Barbhuiya12 found on #102 and I reproduced. `reference_snow_bypass` retains `1 - BYPASS_FRACTION` of every snowfall exactly, so at 0.30 its peak fraction is the constant 0.70 and any floor above that makes it trip `snowpack_response` beside `closure` — and that control exists to show water leaving the snowpack *while the catchment budget still closes*, which is what separates it from `reference_leaky`. 0.70 itself is worse: on seed 7 the two cycles land at `0.7` and `0.6999999999999998`.

So both move together. The control's peak fraction becomes 0.85 — above the new floor, below Snow-17's 0.9084 — and its declared leak, 15%, is still three times the 5% closure limit it is caught on.

Over 200 seeds:

| model | role | wrong | peak fraction min | criteria tripped |
| `reference_bucket` | must_pass | 0/200 | 1.0000 | — |
| `sacsma_snow17` | must_pass | 0/200 | 0.9084 | — |
| `reference_snow_bypass` | must_fail | 0/200 | 0.8500 | `closure` on every seed |
| `reference_snowless` | must_fail | 0/200 | 0.0000 | `snowpack_response` on every seed |

Two new tests pin it, and both fail if `BYPASS_FRACTION` returns to 0.30 or the floor passes 0.85.

The probe README now also discloses what the floor does *not* close off: a model that declares a large liquid share and stores the rest conserves mass, so this probe passes it. That question belongs to `energy/snowpack-cold-content`.

## 2. `baselines.must_fail_only`

The gate accepts a `must_fail` control when its declared criterion is *among* those tripped ([cli.py:356](src/hydroturing/cli.py)). That has to stay the default — 17 of the suite's 66 controls pick up a second criterion as a side effect, `reference_degenerate` tripping `et_plausible` beside `non_degenerate` among them — but it means nothing notices when a control whose job is to separate two failure modes stops doing it. Raising the peak floor past `reference_snow_bypass` left the gate green at every value tested.

`must_fail_only` names the subset that must trip its criterion and no other; everything left out keeps today's rule. A name that is not a `must_fail` baseline is a spec error, so a typo cannot quietly restore the loose rule, and the gate line reads "expected `closure` only" when the stricter rule applies, so the output states which rule it used.

`mass/snowpack-mass-closure` is the first user. Reverting `BYPASS_FRACTION` to 0.30 now exits 1 where it used to pass.

## Validation

- `ht validate`: 31 probes, 72 models, suite 0.1.0
- `pytest -q`: 1142 passed
- full `ht gate`: passed
- `ht gate --probe mass/snowpack-mass-closure`: passed, `reference_snow_bypass` reported as ``expected `closure` only, tripped ['closure']``
- 200-seed sweep of all four baselines: table above
- the archived `sacsma_snow17` row regenerates byte-for-byte apart from its date, so `models/result.csv` is unchanged
- commit 1 verified green on its own

Both commits touch a probe's criteria and a reference model, so this wants two mass-pool approvals under the rule from #100.


chrimerss:
@copilot's finding was correct, and worth more than its "low" label: I verified it
rather than taking it on faith. Deleting the two lines that enforce the stricter
rule left **1142 tests passing** — the CLI silently back on the membership rule,
which is exactly the regression `must_fail_only` exists to catch. The snowpack
test asserts on a `run_probe` outcome and the harness tests only parsed the list,
so neither reached `cmd_gate`.

Fixed in `71797c7` with three tests that drive `cmd_gate` directly, supplying each
control's failing criteria rather than running the real models — what is under
test is the gate's rule, not what those models happen to trip today:

- a `must_fail_only` control tripping a second criterion exits 1, is marked
  BROKEN, and the failure names `'closure' only`;
- the same control, isolated, exits 0 and its line reads ``expected `closure` only``,
  so the output states which rule it applied;
- `reference_snowless`, which did not opt in, keeps the loose rule — an extra
  criterion is not a failure and its line carries no `only`.

Each fails on the mutation it is there for:

| mutation | result |
| --- | --- |
| strict branch deleted | 1 failed |
| `isolated = True` for every control | 1 failed |
| `only` label dropped from the output | 2 failed |

`pytest -q` is now 1145 passed.

